PR #14, opened 2014-07-22
result of sorted volumes is wrong

KatharinaLemke:
The attribute order of mods:part hast to be a positive integer.
As I can see the belonging database field is
tx_dlf_documents.volume_sorting - but it is of the type tinytext.
As a consequence the result of sorted volumes is wrong like this:
[...]/390/40/400/450/50/500/[...]
So I use "CAST ... AS UNSIGNED" to get a correct volume list but it
would be better to change the data type of the affected database field.


albig:
We can reproduce this issue but have not yet decided howto fix it.

IMHO the order field is even of type integer - not positiveinteger. That's what I found in the MODS-spec 3.4 and 3.5. But that's a minor question ;-)

PR #16, opened 2014-07-28
Fix facet limit

KatharinaLemke:
If you want to make for example an alphabetical index of persons, you
need all entries of the selected facet "persons". If you set
"limitFacets" in the typo3 backend, the maximum value is 100. I think it
is inconseqent to offer the possibility to set this value - but only
until 100.
If you need more than 100 entries for a facet, you have to set
facet.limit of SOLR because by default facet.limit=100.


albig:
Looks reasonable to give the flexform facet.limit to the SOLR search parameters. Looking forward to see your usecase "alphabetical index of persons" in real.

PR #20, opened 2014-11-10
Two small code fixes

stweil:
One of them is needed for newer versions of TYPO3 (especially for the current LTS version 6.2). We did not test it with old TYPO3 LTS 4.5, but I expect that it will work there, too.
Another bug was detected while we validated the generated HTML code.

### DIFF
--- a/dlf/cli/class.tx_dlf_cli.php
+++ b/dlf/cli/class.tx_dlf_cli.php
@@ -115,7 +115,7 @@ class tx_dlf_cli extends t3lib_cli {
 		);
 
 		// Run parent constructor.
-		parent::t3lib_cli();
+		parent::__construct();
 
 	}
 
--- a/dlf/plugins/search/setup.txt
+++ b/dlf/plugins/search/setup.txt
@@ -15,7 +15,7 @@ plugin.tx_dlf_search.facetsConf {
 		ACT < .NO
 		ACT.wrapItemAndSub = <li class="tx-dlf-search-act">|</li>
 		ACTIFSUB < .NO
-		ACTIFSUB.wrapItemAndSub = <li class="tx-dlf-search-act" tx-dlf-search-ifsub>|</li>
+		ACTIFSUB.wrapItemAndSub = <li class="tx-dlf-search-act tx-dlf-search-ifsub">|</li>
 	}
 	2 < .1
 	2 {


albig:
Would you please split these two commits and provide as two pull requests? Both are fixes but don't belong together.

The fix for TYPO3 6.2 is fine for 4.6 and 4.7 but breaks 4.5. Please make a version comparison like e.g. in tx_dlf_helper::intInRange().

http://api.typo3.org/typo3cms/45/
http://api.typo3.org/typo3cms/47/

Thank you!

Alexander

stweil:
Hi Alexander,

feel free to cherry-pick only those commits of the pull request which you want to apply. You can do this without waiting for a new pull request.

Did you test the TYPO3 6.2 fix with 4.5 or do you just assume that it might break that version? I'd expect that the new PHP5 code also works with 4.5 because calling the constructor like that is a PHP5 feature, not a feature introduced with some TYPO3 version.

Regards
Stefan

stweil:
The TYPO3 6.2 patch would break PHP4 because it replaces a named constructor by the new constructor which was introduced in PHP5. I don't think we support PHP4, so the patch should be fine. See http://stackoverflow.com/questions/1883913/php4-constructor-naming-convention-in-php5 for details.

albig:
In TYPO3 4.5 is no constructor in class t3lib_cli. That's of course not the problem, but the "cli_args array" will never be loaded then. I can't imagine that this makes any sense. But no, I haven't tried it yet.

stweil:
There is a constructor in class t3lib_cli. See https://svn.typo3.org/TYPO3v4/Core/trunk/t3lib/class.t3lib_cli.php or have a look at your local code and search for constructor. As long as you don't use PHP4, the new code should work with any version of TYPO3.

stweil:
Here is a stripped down version of the code which shows that the new code still works with TYPO3 4.5.

``` php
<?php

class t3lib_cli {
  /**
   * Constructor
   * Make sure child classes also call this!
   *
   * @return void
   */
  function t3lib_cli() {
    // Loads the cli_args array with command line arguments
    echo("Constructor t3lib_cli was called\n");
  }
}

class tx_dlf_cli extends t3lib_cli {
  public function __construct() {
    echo("Constructor tx_dlf_cli was called\n");
    // Run parent constructor.
    parent::t3lib_cli();
  }
}

class tx_dlf_cli2 extends t3lib_cli {
  public function __construct() {
    echo("Constructor tx_dlf_cli2 was called\n");
    // Run parent constructor.
    parent::__construct();
  }
}

$test = new tx_dlf_cli();
$test2 = new tx_dlf_cli2();

?>
```

Just run the code through a command line PHP5 interpreter. Here is my result:

```
# php5 /var/tmp/constructor.php 
Constructor tx_dlf_cli was called
Constructor t3lib_cli was called
Constructor tx_dlf_cli2 was called
Constructor t3lib_cli was called
```

albig:
Thanks for the example. I didn't remember this PHP4 behaviour (constructor named like class name). It's too long ago ;-)

TYPO3 4.5 requires PHP 5.2.x-5.5.x and doesn't support PHP4 at anymore.
Goobi-presentation has set 5.3.0 as minimum. Don't ask me why ;-)

PR #23, opened 2014-12-17
Add Typeswitch for DFG-Viewer, Update OpenLayers

albig:
This adds a type-switch plugin to use with typoscript conditions. This makes it possible for to render page content differently depending on the provided METS-file (e.g. for newspapers).

Additionally the pageview plugin has been updated from OpenLayers 2.12 to 2.13.1.


stweil:
Is dlf/lib/OpenLayers an unmodified copy of OpenLayers 2.13.1? What about using a git submodule for OpenLayers (and other external code in dlf/lib) in the future?

albig:
Yes, it's an unmodified copy of OpenLayers without examples, docs etc. Maybe, we could think about git submodule on the jump to OpenLayers 3.

PR #29, opened 2015-02-10
version without any label from BFallert

BFallert:
### COUNT_VOLUMES_NUM

We need a version without any label

Signed-off-by: BFallert bernd.fallert@bib.uni-mannheim.de


albig:
Why don't you override the language label of the collection plugin? This will erase the "title" and "volume" Labels: 

plugin.tx_dlf_collection._LOCAL_LANG.de {
  titles =
  title =
  volumes =
  volume =
}

plugin.tx_dlf_collection._LOCAL_LANG.en {
  titles =
  title =
  volumes =
  volume =
}

BFallert:
I need the output of this plugin in different places. Only in the start-page i want to have the issue without a label.

If i use your solution do i need to make a copy of the plugin, if i need both versions (with and without label)?

sebastian-meyer:
Just include the proposed Typoscript in different pages' TS templates. You can theoretically use different language labels for each page if you want to. No need to duplicate the plugin code.

You would use the above Typoscript for your homepage and the default TS on any other site.

PR #30, opened 2015-03-04
Avoid calling deprecated method t3lib_div::loadTCA for TYPO3 6.1 and new...

stweil:
...er

t3lib_div::loadTCA was deprecated in TYPO3 6.1. It is now an empty
method. Don't use it with TYPO3 6.1 or later versions because that
would fill the deprecation log of TYPO3 (if it is enabled).

See also
http://forum.typo3.org/index.php/t/199177/
https://forge.typo3.org/issues/47835

### DIFF
--- a/dlf/ext_tables.php
+++ b/dlf/ext_tables.php
@@ -221,7 +221,9 @@ $TCA['tx_dlf_libraries'] = array (
 t3lib_extMgm::addStaticFile($_EXTKEY, 'typoscript/', 'Basic Configuration');
 
 // Register plugins.
-t3lib_div::loadTCA('tt_content');
+if (version_compare(TYPO3_branch, '6.1', '<')) {
+	t3lib_div::loadTCA('tt_content');
+}
 
 // Plugin "collection".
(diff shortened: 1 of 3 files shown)


albig:
Looks good. 

Nice way to compare the TYPO3 version. We used the "hard" way until now, like 

if (t3lib_utility_VersionNumber::convertVersionNumberToInteger(TYPO3_version) >= 6000000) {}

Just PHP version_compare() is much cleaner, I think.

sebastian-meyer:
:+1:

dlf/ext_tables.php
```diff
@@ -221,7 +221,9 @@
 t3lib_extMgm::addStaticFile($_EXTKEY, 'typoscript/', 'Basic Configuration');
 
 // Register plugins.
-t3lib_div::loadTCA('tt_content');
+if (version_compare(TYPO3_branch, '6.1', '<')) {
```
helhum, line 224: minumum TYPO3 version specified for this ext is 6.2. Why not just removing these lines completely?

sebastian-meyer: You are right! Those lines are obsolete now.

Would you mind making a pull request with your proposed changes? (I think there are more obsolete version checks for PHP and TYPO3 version in the code.)

PR #31, opened 2015-03-04
Replace deprecated function t3lib_div::array_merge_recursive_overrule

stweil:
That function was deprecated in TYPO3 6.2. Add a helper function which
calls the correct function depending on the TYPO3 version.

This approach was copied from the TYPO3 extension realurl.

### DIFF
--- a/dlf/common/class.tx_dlf_helper.php
+++ b/dlf/common/class.tx_dlf_helper.php
@@ -54,6 +54,28 @@ class tx_dlf_helper {
 	protected static $locallang = array ();
 
 	/**
+	 * Implements array_merge_recursive_overrule() in a cross-version way.
+	 *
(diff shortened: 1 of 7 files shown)


sebastian-meyer:
I suggest adding the helper function in class `tx_dlf_helper` instead of `tx_dlf_document`. `tx_dlf_helper` already contains other functions which imitate old TYPO3 features (like `testInt()` or `intInRange()`).

sebastian-meyer:
:+1:

stweil:
Now tx_dlf_helper is used as suggested. A runtime test is still to be done.

stweil:
The patch is now installed on our test system http://digi-test.bib.uni-mannheim.de/ and works, so feel free to pull it.

PR #33, opened 2015-03-11
Fix the translation of content elements within the tables tx_dlf_collections, tx_dlf_libraries, tx_dlf_metadate and tx_dlf_structures

frank-ulrich-weber:
The $labels already contain the translated content element, but with the index_name of the translated content element itself and not with the $index_name of the original that we receive here. So we have to determine the index_name of the associated translated content element.
E.g. $labels['title0'] != $index_name = title.

### DIFF
--- a/dlf/common/class.tx_dlf_helper.php
+++ b/dlf/common/class.tx_dlf_helper.php
@@ -1180,6 +1180,45 @@ class tx_dlf_helper {
 
 		}
 
+		/* The $labels already contain the translated content element, but with the index_name of the translated content element itself
+		 * and not with the $index_name of the original that we receive here. So we have to determine the index_name of the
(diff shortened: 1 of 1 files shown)


sebastian-meyer:
This fixes the issue, but I think the real problem here is, that the translated item's `index_name` differs from the original's in the first place. In my opinion this shouldn't be the case.

This behavior is caused by the TCA configuration of `index_name` as unique field which disallows using the same value for the original item and its translation. We need this uniqueness when it comes to regular metadata fields because the `index_name` is used as Solr field name and therefore has to be unique for each metadata field.

I will merge this pull request, but we should try to find a way to garantee uniqueness of Solr index field names while allowing the same `index_name` values for translation purposes at the same time.

PR #35, opened 2015-03-12
OAI plugin: Improve error handling for missing configuration

stweil:
An OAI request with verb=Identity silently fails when configuration data is missing.
Change this and tell the user more about the problem.

### DIFF
--- a/dlf/plugins/oai/class.tx_dlf_oai.php
+++ b/dlf/plugins/oai/class.tx_dlf_oai.php
@@ -973,7 +973,7 @@ class tx_dlf_oai extends tx_dlf_plugin {
 
 			}
 
-			exit;
+			return $this->error('Incomplete plugin configuration');
 
 		}
 


dlf/plugins/oai/class.tx_dlf_oai.php
```diff
@@ -967,13 +967,9 @@ protected function verbIdentify() {
 
 		} else {
 
-			if (TYPO3_DLOG) {
-
-				t3lib_div::devLog('[tx_dlf_oai->verbIdentify()] Incomplete plugin configuration', $this->extKey, SYSLOG_SEVERITY_ERROR);
-
-			}
+			t3lib_div::devLog('[tx_dlf_oai->verbIdentify()] Incomplete plugin configuration', $this->extKey, SYSLOG_SEVERITY_ERROR);
```
sebastian-meyer: Why did you remove the check for `TYPO3_DLOG`? I think Goobi.Presentation should only post messages to devLog if an administrator has explicitly turned logging on by setting `TYPO3_DLOG` in the TYPO3 configuration. At least this is how I understood the purpose of this setting.

sebastian-meyer:
Apart from my comment above this seems to be a reasonable change. :+1:

dlf/plugins/oai/class.tx_dlf_oai.php
stweil: TYPO3_DLOG was introduced in 2004 to implement logging "with a minimum loss of speed if no logging is asked for". In our case here speed is certainly not an argument.

The new code will write a log message only if the devlog extension is enabled and if there is a bad configuration, so normal production installations won't get logging here even without TYPO3_DLOG.

And those people who install devlog might be interested to see error messages in their TYPO3 log, but usually they don't want debug messages from all parts of TYPO3 which they would get when TYPO3_DLOG is set.

That's why I think removing the check for TYPO3_DLOG here is good. Otherwise only the end user but not the system administrator would see that there is a problem.

In addition, I looked at other TYPO3 code to see how logging is handled. The extension t3jquery writes error logging without checks. The extension realurl has a wrapper function and uses its own configuration to enable logging. The TYPO3 core uses several strategies for logging, some check TYPO3_DLOG or similar settings, some use wrapper functions, others don't check anything.

Maybe a helper function would be a good solution for dlf, too. Then we could remove a lot of if statements and do a more advanced check only in the wrapper.

sebastian-meyer: I agree with your proposal to add a helper function as wrapper for error handling, but that should IMHO be done in a separate pull request and for the whole extension at once.

stweil:
Now an updated version of the patch only passes the error message to the end user (web interface).
Logging of the error to TYPO3's devlog is no longer enforced.

Enforced logging to devlog can be handled in a uniform way for all parts of the dlf extension in a new helper function for logging which will be added by a future pull request.

sebastian-meyer:
I just realized there is another problem with this. The OAI-PMH specification only allows certain error codes (see http://www.openarchives.org/OAI/openarchivesprotocol.html#ErrorConditions) and non of these matches a wrong or missing configuration. Therefore Goobi may not return an OAI-PMH error, but has to give a general error message without the OAI-PMH context. This currently isn't supported by the plugin's error handling.

Another hint: Even if the error code was allowed, the way `tx_dlf_oai->error($type)` handles the codes does not support `$type` values which are not part of the OAI-PMH specification. You would have to define a new error type within `locallang.xml` and refer to that.

stweil:
I'd keep it simple, return the error code badArgument (the only one allowed for verb=Identify) and set the free error text as it is. There is no need to add a new error code (nor is it allowed to do so). Human readers will read the text and ignore the error code.

When Goobi.Presentation gets an OAI-PMH request, the answer should follow the OAI-PMH specification. This is what my patch does - without this patch users get an empty page for the test case.

stweil:
As an alternative, Identify could always succeed and return some default values (for example "not configured" or "unknown") if configuration data is missing.

sebastian-meyer:
You misunderstood my comment. Your patch doesn't work the way you think it does. Please have a look at how `tx_dlf_oai->error()` works and how it handles its given argument. The argument is used as both the error code and the key for the translation table (`locallang.xml`) which holds the human readable and translatable error message.
Therefore your patch would return the error code `Incomplete plugin configuration` instead of `badArgument` or any other allowed code.

I like your idea of adding default values for `verb=Identify` to circumvent this error. That's a smart solution!

stweil:
I sent a new pull request with default values (and more), so this request here can be closed.

PR #36, opened 2015-03-16
OAI plugin: Improve error handling for missing configuration

stweil:
This replaces my previous pull request:

An OAI request with verb=Identity silently failed when configuration data
was missing.

Change this and show default values for admin e-mail and repository name.
Reduce also the severity of the logging from "error" to "notice".

The earliest datestamp does not depend on the configuration.
It is now calculated even when the configuration is incomplete.

### DIFF
--- a/dlf/plugins/oai/class.tx_dlf_oai.php
+++ b/dlf/plugins/oai/class.tx_dlf_oai.php
@@ -900,6 +900,11 @@ class tx_dlf_oai extends tx_dlf_plugin {
 		}
 
 		// Get repository name and administrative contact.
+		// Use default values for an installation with incomplete plugin configuration.
+
+		$adminEmail = 'unknown@example.org';
+		$repositoryName = 'Goobi.Presentation OAI-PMH interface (incomplete configuration)';
+
 		$result = $GLOBALS['TYPO3_DB']->exec_SELECTquery(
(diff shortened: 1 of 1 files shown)


albig:
Thank you for the new pull request. @sebastian-meyer will have a look at it after his holiday. I'm not so familiar with OAI yet.

sebastian-meyer:
That looks better! Thank you!

But maybe we should use the TYPO3 admin email as default email address (or none if that isn't set either) and something like "Goobi.Presentation OAI-PMH interface" as default repository name?

stweil:
I'm not sure about the e-mail address. The OAI interface is a public interface, and the e-mail address will be visible for anybody. Using an address at example.org won't harm because example.org was invented for that purpose. Using the TYPO3 admin e-mail might be an unwanted side effect which violates privacy.

The default repository name is uncritical, any name is possible there. What about  "Goobi.Presentation OAI-PMH interface (incomplete configuration)"? I'd like to show to users who test it manually (hopefully the people who did the installation) that there remains some tasks to be done.

sebastian-meyer:
I agree with both.

stweil:
The default repository name is now  "Goobi.Presentation OAI-PMH interface (incomplete configuration)".

sebastian-meyer:
Thanks!

PR #37, opened 2015-03-17
xsl: Add support for EPICUR metadata

stweil:
Goobi.Presentation supports the EPICUR metadata format, so why not support it in the XSL style sheet, too...


sebastian-meyer:

PR #39, opened 2015-03-25
plugins/toolbox: Fix class name used for PDF page downloads

stweil:
tx-dlf-tools-pdf-site does not look reasonable.
Maybe tx-dlf-tools-pdf-side was intended, but tx-dlf-tools-pdf-page
is even better.

Users who used the old class name will have to fix their installation,
so some remark in the release notes is necessary.


sebastian-meyer:
We should add this to the release notes, so everyone can change their stylesheets accordingly.

PR #42, opened 2015-04-02
toolbox/pdf: Handle special cases for PDF page download in double page view

stweil:
The old code handled the cases where either left and right page had a
PDF or where there was no PDF for both pages.

The new code also handles the cases where only one of both pages has a PDF.
It then shows the correct download link (either for the left side or for
the right side).

### DIFF
--- a/dlf/plugins/toolbox/tools/pdf/class.tx_dlf_toolsPdf.php
+++ b/dlf/plugins/toolbox/tools/pdf/class.tx_dlf_toolsPdf.php
@@ -114,49 +114,50 @@ class tx_dlf_toolsPdf extends tx_dlf_plugin {
 	 */
 	protected function getPageLink() {
 
-		$pageLink = array ();
+		$page1Link = '';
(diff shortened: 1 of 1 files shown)


sebastian-meyer:
Short array notation ($a = [1, 2, 3, 4]) was added with PHP 5.4. Since Goobi.Presentation currently still supports PHP 5.3 (just like TYPO3 6.2 does), please change this accordingly.

stweil:
@sebastian-meyer: In the updated patch the short array notation was replaced by the old array notation.

Additional remark: anybody who still uses PHP 5.3 should consider upgrading. Although TYPO3 formally supports 5.3, they suggest at least 5.4. As PHP 5.3 is no longer supported (since August 2014?), that suggestion is reasonable.

sebastian-meyer:
Thank you very much!

I agree with your recommendation to use PHP 5.4, but we shouldn't enforce it as long as TYPO3 doesn't.

PR #48, opened 2015-04-24
require_once() in class.ext_update.php breaks Extension Manager in 6.2.1...

albig:
...2

### DIFF
--- a/dlf/class.ext_update.php
+++ b/dlf/class.ext_update.php
@@ -22,8 +22,6 @@
 *  This copyright notice MUST APPEAR in all copies of the script!
 ***************************************************************/
 
-require_once('common/class.tx_dlf_helper.php');
-
 /**
  * [CLASS/FUNCTION INDEX of SCRIPT]
  */


sebastian-meyer:
Hm, removing the `require_once()` call completely will break compatibility to TYPO3 4.x. Thus `ext_emconf.php` should be changed accordingly.

stweil:
@albig Which kind of breakage do you get? We run TYPO3 6.2 and did not change that part of the code. require_once calls to core files must be removed for the migration from 4.5 to 6.2 (the code without them also works with 4.5), but here we don't have a core file.

albig:
dlf only supports TYPO3 from 4.5, so require_once() is not necessary anymore.

@stweil: Do you use 6.2.12 from April 16th? If yes, have a look in the Extension Manager. It didn't show up on my installatin as the EM fails with the following PHP Error:

<pre>
PHP Fatal error:  require_once(): Failed opening required 'common/class.tx_dlf_helper.php' [...] in
[...]/TYPO3.CMS/typo3/sysext/extensionmanager/Classes/Utility/UpdateScriptUtility.php(111) : eval()'d code on line 25\n', 
</pre>

This is new code in 6.2.12 which checks all update scripts, if there is something to do:
- http://wiki.typo3.org/TYPO3_CMS_6.2.12#Changes
- https://github.com/TYPO3/TYPO3.CMS/commit/c58ba3e
- https://forge.typo3.org/issues/66263

Thanks to @chrode for helping.

stweil:
We still use TYPO3 6.2.11. I get this error when I switch from 6.2.11 to 6.2.12. With your patch, the above error is gone, but now I get a new one:

```
Uncaught TYPO3 Exception
Serialization of 'Closure' is not allowed

Exception thrown in file
/var/www/html/typo3conf/ext/devlog/class.tx_devlog.php in line 134.
```

albig:
This happens due to #49

PR #67, opened 2015-07-24
respect page size if given in ALTO fulltext files

albig:
The WIDTH and HEIGHT attribute is optional of the Page element. If it
doesn't exists we take WIDTH and HEIGHT of the PrintSpace element.


stweil:
Tested-by: Stefan Weil sw@weilnetz.de

Maybe LABEL and MILESTONE should still be added, so this improvement does not get lost in patch releases.

PR #70, opened 2015-08-28
Small updates and fixes

jacmendt:


dlf/plugins/pageview/tx_dlf_pageview_fulltext_control.js
```diff
@@ -32,6 +33,12 @@ var dlfViewerFullTextControl = function(map){
 
     /**
      * @private
+     * @type {string}
+     */
+    this.lang = lang == 'de' || lang == 'en' ? lang : 'de';
```
stweil, line 38: This needs to be fixed when new languages are added. Wouldn't this be sufficient?
<code>this.lang = lang;</code>

Issue #65, opened 2015-07-21
Re-work/simplify copyright header


stweil:
The TYPO3 project re-worked the copyright/license headers, see https://forge.typo3.org/issues/59535.
I suggest that Goobi.Presentation should do something similar (for the same reasons).

See also the discussion on https://forge.typo3.org/issues/56776.

sebastian-meyer:
I switched the license text to GPL3 and changed the headers accordingly.

PR #75, opened 2015-09-28
Ol3

jacmendt:
Add an image manipulation tool to goobi-presentation. The image manipulation is based on the ol3 webgl renderer and therefore needs that the image tile provider allow [CORS](http://www.w3.org/TR/cors/). Because right now ol3 web gl renderer does not support the drawing of polygons, but their are needed for fulltext behavior, the image manipulation control uses it own map object.

I further moved the attaching of click event handler for the fulltext and image manipulation to the javascript control objects (before they were within the template / php -controller). The fulltext control could now be initializing by adding the control element (e.g. anchor element) with a id `tx-dlf-tools-fulltext`. It will then load the event handler by itself.


.gitignore
```diff
@@ -4,6 +4,7 @@
 .project
 .settings/
 .idea/
+.project
```
stweil: duplicate line, not needed

dlf/plugins/pageview/tx_dlf_pageview_fulltext_control.js
```diff
@@ -226,12 +328,12 @@ var dlfViewerFullTextControl = function(map, lang){
 dlfViewerFullTextControl.prototype.enableFulltextSelect = function(textBlockFeatures, textLineFeatures) {
 
     // add features to map
-    this.textBlockLayer.getSource().addFeatures(textBlockFeatures);
-    this.textLineLayer.getSource().addFeatures(textLineFeatures);
+//    this.textBlockLayer.getSource().addFeatures(textBlockFeatures);
+//    this.textLineLayer.getSource().addFeatures(textLineFeatures);
```
stweil: Are those lines still needed?

dlf/plugins/pageview/tx_dlf_utils.js
```diff
@@ -104,4 +168,52 @@ dlfUtils.setCookie = function(name, value) {
...
+
+    	image = {
+            'width': width,
+            'height': height,
+            'scale': imageObj.width/width,
+        }
+
+    }
+    
+    var scale = image.scale,
```
stweil, line 218: What happens if image is undefined (see code above)?

dlf/plugins/pageview/tx_dlf_utils.js
```diff
@@ -104,4 +168,52 @@ dlfUtils.setCookie = function(name, value) {
...
+    	height = imageObj.height,
```
stweil: This hides the function parameter with the same name.

dlf/plugins/toolbox/tools/imagemanipulation/class.tx_dlf_toolsImagemanipulation.php
```diff
@@ -0,0 +1,114 @@
...
+*
+*  The GNU General Public License can be found at
+*  http://www.gnu.org/copyleft/gpl.html.
+*
+*  This script is distributed in the hope that it will be useful,
+*  but WITHOUT ANY WARRANTY; without even the implied warranty of
+*  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+*  GNU General Public License for more details.
+*
+*  This copyright notice MUST APPEAR in all copies of the script!
```
stweil, line 22: See issue #65.

sebastian-meyer, line 22: We should update the license headers all at once, so using the current/old one here is OK.

dlf/plugins/pageview/tx_dlf_utils.js
jacmendt: Sorry but didn't get your point her. Could you please further explain?

dlf/plugins/pageview/tx_dlf_utils.js
```diff
@@ -104,4 +168,55 @@ dlfUtils.setCookie = function(name, value) {
...
+            'scale': imageObj.width/width,
+        }
+
+    }
+    
+    if (image === undefined)
+    	return [];
+    
+    var scale = image.scale,
+    	height = imageObj.height,
```
stweil: Here you declare a local variable called "height". There is already a function parameter called "height" which can no longer be accessed (so it is hidden by the new variable). This is confusing for readers of the code, and I expect that static code analyzers will complain.

dlf/plugins/pageview/tx_dlf_utils.js
stweil: Yes, of course. See my new comment in your updated code.

dlf/plugins/toolbox/tools/imagemanipulation/class.tx_dlf_toolsImagemanipulation.php
stweil, line 22: No objection. It does not matter whether we add some more files with this "MUST APPEAR" notice, as long as we'll remove it some day nevertheless. :-)

PR #79, opened 2015-12-14
Feature elasticsearch

sebastian-meyer:
Add support for Elasticsearch backend
Add FEDORA compatible download tool


sebastian-meyer:
This currently works only in combination with Goobi.Publication, because search results are linked to plugins from extension "dpf".

PR #90, opened 2016-01-07
Replace deprecated class t3lib_iconWorks

stweil:
It was deprecated with TYPO3 6.0.


stweil:
Would it be OK if I combine several of such commits in one pull request, or do you prefer to get them one by one?

sebastian-meyer:
You can combine them into one pull request since they all address the same issue.

Thank you for your contributions!

PR #92, opened 2016-01-07
Remove compatibility code for TYPO3 < 6.2

stweil:
It is no longer needed because only TYPO3 6.2 and newer is now supported.

### DIFF
--- a/dlf/common/class.tx_dlf_helper.php
+++ b/dlf/common/class.tx_dlf_helper.php
@@ -845,17 +845,7 @@ class tx_dlf_helper {
 	 */
 	public static function intInRange($theInt, $min, $max = 2000000000, $zeroValue = 0) {
 
-		if (t3lib_utility_VersionNumber::convertVersionNumberToInteger(TYPO3_version) >= 6000000) {
-
-			// TYPO3 > 6.0
-			return t3lib_utility_Math::forceIntegerInRange($theInt, $min, $max, $zeroValue);
-
-		} else {
-
-			// TYPO3 4.5 - 4.7
-			return \TYPO3\CMS\Core\Utility\GeneralUtility::intInRange($theInt, $min, $max, $zeroValue);
-
-		}
+		return t3lib_utility_Math::forceIntegerInRange($theInt, $min, $max, $zeroValue);
 
 	}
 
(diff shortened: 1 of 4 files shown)


dlf/common/class.tx_dlf_helper.php
```diff
@@ -845,17 +845,7 @@ public static function getURN($base, $id) {
...
-			// TYPO3 > 6.0
-			return t3lib_utility_Math::forceIntegerInRange($theInt, $min, $max, $zeroValue);
-
-		} else {
-
-			// TYPO3 4.5 - 4.7
-			return \TYPO3\CMS\Core\Utility\GeneralUtility::intInRange($theInt, $min, $max, $zeroValue);
-
-		}
+		return t3lib_utility_Math::forceIntegerInRange($theInt, $min, $max, $zeroValue);
```
sebastian-meyer, line 848: Maybe we should completely remove this helper function and replace all calls to it with a direct call to `t3lib_utility_Math::forceIntegerInRange`? The helper function was only introduced to have a downwards compatibility for TYPO3 4.5, which is no longer needed.

dlf/common/class.tx_dlf_helper.php
```diff
@@ -1123,17 +1113,7 @@ public static function saveToSession($value, $key) {
...
-			// TYPO3 > 6.0
-			return t3lib_utility_Math::canBeInterpretedAsInteger($theInt);
-
-		} else {
-
-			// TYPO3 4.5 - 4.7
-			return \TYPO3\CMS\Core\Utility\GeneralUtility::testInt($theInt);
-
-		}
+		return t3lib_utility_Math::canBeInterpretedAsInteger($theInt);
```
sebastian-meyer, line 1116: Maybe we should completely remove this helper function and replace all calls to it with a direct call to `t3lib_utility_Math::canBeInterpretedAsInteger`? The helper function was only introduced to have a downwards compatibility for TYPO3 4.5, which is no longer needed.

stweil:
t3lib_utility_Math is also one of the deprecated interfaces. The replacement is (like in the previous patches) a lengthy string \TYPO3\CMS\Core..., so the calling lines will get longer.

I can either keep the helper functions and replace t3lib_utility_Math there, or I can remove the helper functions and replace all callers with TYPO3 7 compatible code. Either way it should be a separate patch.

sebastian-meyer:
I'd prefer removing the helper functions for consistency. Doing it in a separate patch is fine! Thanks!

PR #97, opened 2016-01-10
Fix interface to \TYPO3\CMS\Core\Messaging\FlashMessageQueue

stweil:
The methods addMessage and renderFlashMessage supported static calls
which were deprecated in TYPO3 6.1 and removed in TYPO3 6.3 and later
versions.

In addition, renderFlashMessage escapes html code since TYPO3 7.4
which is not what is needed for Goobi.Presentation.

### DIFF
--- a/dlf/common/class.tx_dlf_document.php
+++ b/dlf/common/class.tx_dlf_document.php
@@ -1310,7 +1310,7 @@ final class tx_dlf_document {
 						TRUE
 					);
 
-					\TYPO3\CMS\Core\Messaging\FlashMessageQueue::addMessage($message);
+					tx_dlf_helper::addMessage($message);
 
 				}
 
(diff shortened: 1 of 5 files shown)


sebastian-meyer:
`dlf/common/class.tx_dlf_document.php` and `dlf/common/class.tx_dlf_indexing.php` are not derived from `dlf/common/class.tx_dlf_module.php`, so the backported function `addMessage()` isn't in their scope and can't be called from `$this`. Or am I missing something?

(Only the backend modules in `dlf/modules/...` extend `dlf/common/class.tx_dlf_module.php`.)

stweil:
You are correct, thanks for checking. The new code will not work for those files. Then it looks like I should move those functions to tx_dlf_helper (and make them static).

sebastian-meyer:
Yes, that's the way we handled backported API functions in the past.

Thanks in advance!

stweil:
The updated pull request which uses tx_dlf_helper is now available.

PR #107, opened 2016-03-03
Remove dead code and fix saving of map view parameters

jacmendt:


dlf/plugins/pageview/flexform.xml
```diff
@@ -68,22 +68,10 @@
 										<numIndex index="0">LLL:EXT:dlf/plugins/pageview/locallang.xml:tt_content.pi_flexform.features.overviewmap</numIndex>
 										<numIndex index="1">OverviewMap</numIndex>
 									</numIndex>
-									<numIndex index="1" type="array">
-										<numIndex index="0">LLL:EXT:dlf/plugins/pageview/locallang.xml:tt_content.pi_flexform.features.panpanel</numIndex>
-										<numIndex index="1">PanPanel</numIndex>
-									</numIndex>
 									<numIndex index="2" type="array">
```
sebastian-meyer: `index` should be changed to `1`

PR #126, opened 2016-04-28
Add color invert filter to image manipulation tool

jacmendt:
- Adds an invert color filter to the image manipulation tool. @gnuj has to add the corresponding styling for it.


gnuj:
Allready added styles to the corresponding projects. So this request could be merged and shipped to our projects.

PR #138, opened 2016-06-21
fix #137: make sure the moved and zoomed image is in viewport initially

### LINKED ISSUE #137: Main image may be positioned outside viewport.
The image position and zoom level ist stored in a session cookie. If the position values are outside the current viewport, the main image won't be seen.

This happens e.g. in case of different image dimensions in one digital object.

Example:
http://digital.slub-dresden.de/werkansicht/dlf/6439/

albig:


albig:
This fix #137.

PR #151, opened 2016-12-05
- Adaptation to MySQL >= 5.7.6 where STORED is a reserved word now

frank-ulrich-weber:


stweil (commented):
It would also be possible to rename "stored", for example in "metadata_stored" or "md_stored" (in the code, not in the database of course).

What about the "stored" in `dlf/common/class.tx_dlf_solr.php`?

sebastian-meyer:
In `dlf/common/class.tx_dlf_solr.php` "stored" is not used in a "AS" statement but as a field name. If that is a problem with MySQL >= 5.7.6, too, we need to change even more queries. In that case, renaming the database field would possibly be the best idea.

frank-ulrich-weber:
I missed that line in dlf/common/class.tx_dlf_solr.php. Thanks.
I think that there are no more occurrences and I like to have the same name for the collumn and the result set.


________________________________

Frank-Ulrich Weber
Product Manager Software Solutions


[cid:Z_Logo_RGB_180px_2b974e26-85b9-4005-92dd-9bb8df881ab3.png]<http://www.zeutschel.de> <http://www.zeutschel.de>

[cid:Facebook-34x34_ab94d89a-875f-49f2-81f3-e136c66e4bb5.png]<https://www.facebook.com/pages/Zeutschel-GmbH/193873073980288?fref=ts>


[cid:Twitter-34x34_f9819937-1c34-4eab-b2fc-944fcf2e8938.png]<https://twitter.com/zeutschelgmbh>


[cid:YouTube-34x34_8cf03759-cc15-472e-a763-e628ea59d43b.png]<http://www.youtube.com/user/zeutschelbookscanner>


[cid:google_34x34_daf218c4-f635-49e8-af7a-ed2a74c251ea.png]<https://plus.google.com/110507211572689796815/posts>


Zeutschel GmbH | Heerweg 2 | 72070 Tübingen | Deutschland
p: +49 (7071) 9706-56 | m: | f: +49 (7071) 9706-44
e: Frank-Ulrich.Weber@zeutschel.de<mailto:Frank-Ulrich.Weber@zeutschel.de> | w: http://www.zeutschel.de

[cid:zeta-banner-86x75_fuerWebsite_c5e46c08-490e-49fa-b13f-d59217ddd169.png]<http://www.zeutschel.de/links/Zeta-App>


Geschäftsführer/President: Joerg Vogler | Registergericht Stuttgart: HRB 380917

frank-ulrich-weber:
Should I change the the query to tx_dlf_metadata.stored within dlf/common/class.tx_dlf_solr.php?

sebastian-meyer:
That seems to be a reasonable solution. Maybe prepend all the field names in that query with the table name in order to avoid confusion?

stweil:
Just for the records an [info for MariaDB](https://mariadb.com/kb/en/mariadb/virtual-computed-columns/):

> The STORED keyword was not supported. It was added in MariaDB 10.2.1 to make virtual fields definitions compatible with MySQL 5.7

So the modification is needed for MariaDB as well. We use that modern fork of MySQL on all newer installations (not only for Kitodo).

stweil:
@sebastian-meyer, according to the [MySQL documentation](http://dev.mysql.com/doc/refman/5.7/en/keywords.html), `STORED` is a reserved keyword, and such keywords need quoting for use as identifiers such as table and column names.

sebastian-meyer:
Thanks for the heads-up! Then maybe we should consider renaming the column to avoid future problems with unquoted queries?

stweil:
I don't have a strong opinion on renaming columns. On the one hand that implies an incompatibility which must be handled by scripts and documentation (upgrades, downgrades, imports from Goobi?). On the other hand the current names are really very general (so future conflicts are to be expected) and could be made more specific by adding a prefix.

PR #159, opened 2017-03-23
Add quotes to reserved keyword as field name

Erikmitk:
The keyword `stored` was added to the list of reserved words as of MySQL 5.7.6 and the `CREATE`-Statement throws a syntax error from this MySQL version onward

https://dev.mysql.com/doc/refman/5.7/en/keywords.html

http://stackoverflow.com/questions/23446377/syntax-error-due-to-using-a-reserved-word-as-a-table-or-column-name-in-mysql#23446378

### DIFF
--- a/dlf/ext_tables.sql
+++ b/dlf/ext_tables.sql
@@ -88,7 +88,7 @@ CREATE TABLE tx_dlf_metadata (
     default_value text NOT NULL,
     wrap text NOT NULL,
     tokenized tinyint(4) DEFAULT '0' NOT NULL,
-    stored tinyint(4) DEFAULT '0' NOT NULL,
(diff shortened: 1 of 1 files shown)


Erikmitk:
This change only allows the extension to be installed correctly, but there are probably more consequences to this change. Any use of the reserved field name probably has to be escaped and tested as well.

Erikmitk:
After experimenting a bit more with this, we found that the change doesn’t really do anything. The escaping of the field name works perfectly when fed directly into MySQL. But when the Query gets executed when the extension is installed, I guess the escaping goes missing because it still doesn’t create the table tx_dlf_metadata.

So the field name `stored` is still problematic when you use a MySQL Version > 5.7.6, despite this code change.

PR #161, opened 2017-03-23
fix #158: replace usage of file_get_contents by GeneralUtility::getUrl

### LINKED ISSUE #158: Use TYPO3 getUrl() instead of pure PHP file_get_contents()
If using TYPO3 behind a HTTP-Proxy, Kitodo.Presentation will fail to load the METS-files. The global proxy settings of TYPO3 won't be applied as the pure PHP `file_get_contents()` method is used.

TYPO3 GeneralUtilities provides the method `getUrl()` which is available in 6.2, 7.8 and 8.x. This ensures the usage of given proxy settings and may use curl inside. The result should be the same.

Files to be adapted:
- plugins/pageview/class.tx_dlf_fulltext_eid.php
- common/class.tx_dlf_document.php
- common/class.tx_dlf_indexing.php
- hooks/class.tx_dlf_em.php
- hooks/class.tx_dlf_tcemain.php

albig:
Using the TYPO3 core function GeneralUtility::getUrl() respects setting e.g. for a HTTP proxy. This patch replace the usage in common document, indexing and inside the fulltext plugin.

In the connection handling with SOLR server there remains some file_get_contents which uses context settings.

### DIFF
--- a/dlf/plugins/pageview/class.tx_dlf_fulltext_eid.php
+++ b/dlf/plugins/pageview/class.tx_dlf_fulltext_eid.php
@@ -53,7 +53,7 @@ class tx_dlf_fulltext_eid extends \TYPO3\CMS\Frontend\Plugin\AbstractPlugin {
 
 		$url = \TYPO3\CMS\Core\Utility\GeneralUtility::_GP('url');
 
-		$fulltextData = file_get_contents($url);
+		$fulltextData = file_get_contents(\TYPO3\CMS\Core\Utility\GeneralUtility::getUrl($url);
 
 		header('Last-Modified: ' . gmdate( "D, d M Y H:i:s" ) . 'GMT');
 		header('Cache-Control: no-cache, must-revalidate');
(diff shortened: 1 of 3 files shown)


dlf/plugins/pageview/class.tx_dlf_fulltext_eid.php
```diff
@@ -53,7 +53,7 @@ public function main($content = '', $conf = array ()) {
 
 		$url = \TYPO3\CMS\Core\Utility\GeneralUtility::_GP('url');
 
-		$fulltextData = file_get_contents($url);
+		$fulltextData = file_get_contents(\TYPO3\CMS\Core\Utility\GeneralUtility::getUrl($url);
```
henning-gerhardt, line 56: This change does not looks correct if `file_get_contents` call should be replaced.

albig: That's true. Thank you for reviewing. I'll fix it.

PR #183, opened 2017-06-07
CORS detection routed via server eID script

albig:
The imagemanipulation tool requires loading the images via WebGL. This is only allowed if CORS is enabled. [CORS = Cross-Origin Resource Sharing](https://www.w3.org/wiki/CORS_Enabled) requires a header tag "Access-Control-Allow-Origin" by the image provider server.

If CORS is not enabled, the WebGL layer is simply black. To avoid this behaviour, Kitodo.Presentation tries to check if CORS is enabled.

The current implementation does something similar. It tries to load the image in javascript context. If this fails, it should fail in WebGL context too. This is not true in all conditions! In case of mixed content (Kitodo.Presentation running HTTPS, images provided via HTTP) this test always fails. But WebGL is fine with mixed content if the CORS header are sent.

This patch use another approach which was already used to fetch the fulltext files. It fetches the image header via an eID script on the Kitodo.Presentation server. This is done via HTTPS or HTTP as the whole frontend. If the header contains the "Access-Control-Allow-Origin" tag, the CORS-test is true.

As fetching the fulltext files and the image header is in fact the same task, the eID script has been renamed and unified for both. It would be easily possible to fetch even the images via eID. This is not implemented here.

ToDo: Check the value of the "Access-Control-Allow-Origin"-tag to be really perfect.

### DIFF
--- a/dlf/ext_localconf.php
+++ b/dlf/ext_localconf.php
@@ -65,7 +65,7 @@ $GLOBALS['TYPO3_CONF_VARS']['SC_OPTIONS']['GLOBAL']['cliKeys'][$_EXTKEY] = array
 // Register AJAX eID handlers.
 $GLOBALS['TYPO3_CONF_VARS']['FE']['eID_include']['tx_dlf_search_suggest'] = 'EXT:'.$_EXTKEY.'/plugins/search/class.tx_dlf_search_suggest.php';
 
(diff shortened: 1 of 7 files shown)


sebastian-meyer:
+1

That's great! Would fetching the images via eID script even resolve the problems related to mixed content?
(DDB is currently experiencing problems and is considering implementing an image proxy, but your solution is much more elegant.)

albig:
I enhanced the patch to support loading images via eID. This is working but there should be done some more enhancement concerning caching. Currently, the images are loaded twice via eID. Without this setting, the images are cached by the browser and it will be loaded only once via network.

albig:
With Cache-Control header, the browser caching is working now as expected.

PR #191, opened 2017-06-15
adds IIIF API v2 functionality

christopher-johnson:
This adds a custom mimetype (application/vnd.kitodo.iiif2) to support an IIIF API v2 info.json response and the function (dlfUtils.fetchIIIF2Data) to support parsing it into the ImageSourceObj.  The primary difference between the v1 and v2 is the removal of the  "scale_factors", "tile_width" ,"tile_height" properties.  
These are now in a tiles array that looks like this:
` "tiles" : [
     { "width" : 256, "height" : 256, "scaleFactors" : [ 1, 2, 4, 8, 16 ] }
  ]` 

The other changes are cosmetic and primarily to make my IDE not complain... 

This has been tested locally using [this example](https://github.com/ub-leipzig/kitodo/blob/master/kitodo-production/tests/iiiftest.xml).

### DIFF
--- a/dlf/plugins/pageview/tx_dlf_utils.js
+++ b/dlf/plugins/pageview/tx_dlf_utils.js
@@ -30,19 +30,19 @@ dlfUtils.CUSTOM_MIMETYPE = {
 dlfUtils.RUNNING_INDEX = 99999999;
 
 /**
- * @param {Array.<{src: *, width: *, height: *}>} imageSourceObjs
- * @param {string} mimetype
- * @paran {string} opt_origin
+ * @param imageSourceObjs
+ * @param {string} opt_origin
  * @return {Array.<ol.layer.Layer>}
  */
-dlfUtils.createOl3Layers = function(imageSourceObjs, opt_origin){
+dlfUtils.createOl3Layers = function(imageSourceObjs, opt_origin) {
 
-    var origin = opt_origin !== undefined ? opt_origin : null,
+    let origin = opt_origin !== undefined ? opt_origin : null,
         widthSum = 0,
         offsetWidth = 0,
         layers = [];
 
     imageSourceObjs.forEach(function(imageSourceObj) {
+        let tileSize;
         if (widthSum > 0) {
             // set offset width in case of multiple images
             offsetWidth = widthSum;
@@ -51,7 +51,7 @@ dlfUtils.createOl3Layers = function(imageSourceObjs, opt_origin){
         //
         // Create layer
         //
-        var extent = [ 0 + offsetWidth, 0, imageSourceObj.width + offsetWidth, imageSourceObj.height],
+        let extent = [offsetWidth, 0, imageSourceObj.width + offsetWidth, imageSourceObj.height],
             layer;
 
         if (imageSourceObj.mimetype === dlfUtils.CUSTOM_MIMETYPE.ZOOMIFY) {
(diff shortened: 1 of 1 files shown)


albig:
I tried your example locally. It's working as expected.

And there is no other way to distinguish the API versions? Shouldn't we do that depending on the infos we find in info.json which is fetched at the beginning?

christopher-johnson:
There is a lot that could be done with the external resource typing using the info.json profile property of any arbitrary IIIF Image API response.  There are libraries for this that are used in the UV ([manifold](https://github.com/viewdir/manifold) and [manifesto](https://github.com/viewdir/manifesto)).  Note that there are [several variations](https://github.com/viewdir/manifesto/blob/master/src/ServiceProfile.ts) in the service profile string itself.  A utility class for methods to type check IIIF service profiles is [here](https://github.com/viewdir/manifesto/blob/master/src/Utils.ts#L90).  

Typing external resources by means of the info.json is a good idea if you want to move away from the custom MIME type resource binding and abstract more generic handlers.

sebastian-meyer:
Even if we cannot get rid of the MIME type differentiation entirely, IMHO it would be good to at least have only one MIME type defined for all supported IIIF versions.

Cheers,
Sebastian
________________________________
Von: Christopher Johnson
Gesendet: Thu Jun 15 14:50:00 GMT+02:00 2017
An: kitodo/kitodo-presentation
Cc: Subscribed
Betreff: Re: [kitodo/kitodo-presentation] adds IIIF API v2 functionality (#191)


There is a lot that could be done with the external resource typing using the info.json profile property of any arbitrary IIIF Image API response. There are libraries for this that are used in the UV (manifold<https://github.com/viewdir/manifold> and manifesto<https://github.com/viewdir/manifesto>). Note that there are several variations<https://github.com/viewdir/manifesto/blob/master/src/ServiceProfile.ts> in the service profile string itself. A utility class for methods to type check IIIF service profiles is here<https://github.com/viewdir/manifesto/blob/master/src/Utils.ts#L90>.

Typing external resources by means of the info.json is a good idea if you want to move away from the custom MIME type resource binding and abstract more generic handlers.

christopher-johnson:
As suggested, I have modified the PR and removed the new MIME type.  Now the API spec is evaluated if the `protocol ` property (with an expected value) exists in the info.json response.  `protocol ` only exists in responses > v2.  I also added a type check that evaluates if the info.json response is IIIF supported.

Including a `service profile` for every image resource is recommended in the [IIIF Presentation API](http://iiif.io/api/presentation/2.1/#service).   If the METS file is seen as a "backwards compatible" surrogate to a manifest.json, then it might be useful to add a definitive additional service profile property/attribute. This would have no direct association to the custom MIME type that generically identifies a IIIF resource.  Of course, profiling an image would only be relevant if new client viewer libraries are added that require more granular image specs.

One other comment about the PR is the change to ES6 variable declarations.  I think that this adds code clarity as unlike `const`, `var` is ambiguous about reassign-ability.

sebastian-meyer:
Looks good now! Merging your pull request...

Regarding the service profile: AFAIK the service profile has to be provided by the host of the IIIF image API. Or is there anything we can do in Kitodo.Presentation?

christopher-johnson:
I have realized now that in profiling this change, I removed the eID=tx_dlf_geturl_eid from the uri in the isCorsEnabled function [here](https://github.com/kitodo/kitodo-presentation/commit/4e6f43ec9fcaf524f12ad4c9b7ade605801962f8#diff-e31e8547a5204d08e064b07cdca8aed6L452) and did not put it back in in the amended commit.  This was unintentional.

My observation is that with a (correctly configured) IIIF image server, the right headers should be already be set by server, and I think that there is not much to be gained in modifying them by the client.  For example, the XHR with the eID function takes ~3x longer in my test with IIPsrv.  With other servers e.g. static images, however, it may be a different situation... I wonder if this decision to invoke the eID response header modification should be separated from the isCors logic.  If it is in its own function, it can be typed and invoked specifically for your problematic responses.

albig:
Without the eID=tx_dlf_geturl_eid the isCorsTest ist broken for mixed content environments. It must come back at this point.

christopher-johnson:
I suggest a new function that builds the URIs for the XHR in `isCORSEnabled`,  The function should accept a boolean `modifyResponseHeader` a parameter.  If true, then add  `eID=tx_dlf_geturl_eid`.  Again, I do not think that the IIIF server responses should be modified.  I can write this and submit a new PR if you agree.

christopher-johnson:
I have looked at this again this morning and now realize that it is unnecessary to do a `GET` in the isCorsEnabled function.  This loads static images twice and this is the main performance issue that I noticed initially.

Issue #192, opened 2017-06-16
Language Overlay of Metadata and Structures not working as expected


albig:
Currently it is not possible to translate the labels like "Title" of the tx_dlf_metadata table or "Strukturtyp" of the tx_dlf_structures.

Usually you copy a table row and insert it again on the same page. You edit the copy and change the language to the overlay language AND you select the default label.
If this is done, the Metadata-entry is displayed twice: One time with the default label. The second time with the dummy index_name ("title0).

sebastian-meyer:
Just leaving this here for later reference:
https://docs.typo3.org/typo3cms/FrontendLocalizationGuide/singlehtml/#core-support-for-localization

PR #193, opened 2017-06-16
Fix 192: language overlay of metadata and structures

albig:
This is a patch to be able to use the TYPO3 language overlay feature of database records.

First, it excludes fields in the TCA configuration which are not necessary to change in language overlay.

Second, in the metadata plugin, the query of all metadata is limited to the current language.

Third, the translate() helper method is fixed as it did not do, what is written in the comments ;-)

I tested this patch with the following typoscript configuration:

1. an translated page of the viewer must exist
2. config.sys_language_mode = content_fallback
3. config.sys_language_overlay = 1


albig:
This fix #192 already.

PR #194, opened 2017-06-18
restores isCorsEnabled function

christopher-johnson:
this relates to #191 and re-implements the eID parameter to modify headers for all MIME types.

### DIFF
--- a/dlf/plugins/pageview/tx_dlf_utils.js
+++ b/dlf/plugins/pageview/tx_dlf_utils.js
@@ -68,7 +68,7 @@ dlfUtils.createOl3Layers = function(imageSourceObjs, opt_origin) {
 
             tileSize = imageSourceObj.tilesize !== undefined && imageSourceObj.tilesize.length > 0
                 ? imageSourceObj.tilesize[0]
-                    : 256,
+                : 256,
                 format = $.inArray('jpg', imageSourceObj.formats) || $.inArray('jpeg', imageSourceObj.formats)
                     ? 'jpg'
                     : imageSourceObj.formats.length > 0
@@ -573,7 +573,7 @@ dlfUtils.isCorsEnabled = function(imageObjs) {
                 : imageObj.url;
 
         url =
-            window.location.origin + window.location.pathname + '?url=' + encodeURIComponent(
+            window.location.origin + window.location.pathname + '?eID=tx_dlf_geturl_eid&url=' + encodeURIComponent(
                 url) + '&header=2';
 
         $.ajax({
@@ -581,14 +581,10 @@ dlfUtils.isCorsEnabled = function(imageObjs) {
             async: false
         }).done(function(data, type) {
             response = type === 'success' && data.indexOf('Access-Control-Allow-Origin') !== -1;
-        })
-            .error(function(data, type) {
+        }).error(function(data, type) {
                 response = false;
             });
-
     });
-
-
     return response;
 };
 


dlf/plugins/pageview/tx_dlf_utils.js
```diff
@@ -68,7 +68,7 @@ dlfUtils.createOl3Layers = function(imageSourceObjs, opt_origin) {
 
             tileSize = imageSourceObj.tilesize !== undefined && imageSourceObj.tilesize.length > 0
                 ? imageSourceObj.tilesize[0]
-                    : 256,
```
albig, line 71: Please put code beautifying always in a separate commit or even better in a separate pull request.

dlf/plugins/pageview/tx_dlf_utils.js
```diff
@@ -561,38 +561,48 @@ dlfUtils.isCorsEnabled = function(imageObjs) {
...
-
-        url =
-            window.location.origin + window.location.pathname + '?url=' + encodeURIComponent(
-                url) + '&header=2';
+                    ? dlfUtils.buildXhrURI(dlfViewerSource.IIP.getMetdadataURL(imageObj.url), true)
+                    : dlfUtils.buildXhrURI(imageObj.url, true);
 
         $.ajax({
             url: url,
+            type: type,
```
albig: Good idea. But not sure, if this change a lot. With "&header=2" the eID does anyway only a HEAD request by the server. At least, it should do ;-) I have to examine this later.

dlf/plugins/pageview/tx_dlf_utils.js
```diff
@@ -561,38 +561,48 @@ dlfUtils.isCorsEnabled = function(imageObjs) {
...
+    const type = 'HEAD';
     imageObjs.forEach(function(imageObj) {
         let url = imageObj.mimetype === dlfUtils.CUSTOM_MIMETYPE.ZOOMIFY
-            ? imageObj.url.replace('ImageProperties.xml', 'TileGroup0/0-0-0.jpg')
-            :
-            imageObj.mimetype === dlfUtils.CUSTOM_MIMETYPE.IIIF
-                ? dlfViewerSource.IIIF.getMetdadataURL(imageObj.url)
+            ? dlfUtils.buildXhrURI(imageObj.url.replace('ImageProperties.xml', 'TileGroup0/0-0-0.jpg'), true)
+            : imageObj.mimetype === dlfUtils.CUSTOM_MIMETYPE.IIIF
+                ? dlfUtils.buildXhrURI(dlfViewerSource.IIIF.getMetdadataURL(imageObj.url), false)
```
albig: I'm not sure that this approach helps us. You assume, the IIIF-Server always have the right headers, right? In your case it is true, but we frequently see data provider not supporting HTTPS or even CORS. Of course, we know not IIIF-provider so far.

I would prefer a solution to fetch these header e.g. only once per session / url+path. I think it's unlikely that CORS-header change for different images of the same origin.

What do you think?

christopher-johnson: I would never assume that all IIIF servers are properly configured, but if they are not, then it should not break the client.  If `isCorsEnabled` preflight fails, then the client should simply throw an exception and move on.  I would not say that you _should not_ try to correct a misconfigured server with a response header rewriting, because this depends on the scope of your domain, but doing so introduces stateful dependencies, and a potentially a lot of extra complexity (and performance issues).  

A server (in a proper RESTful implementation of CORS), _should_ provide a preflight `OPTIONS` response that includes the Access-Control-Allow-Origin header, but this is not always done (and is not done with IIPsrv).  I think that the `isCorsEnabled` function with a `HEAD` request that does the preflight is all that is required to confirm that a server supports CORS for this client before the `GET` for that resource is sent. 

I think that grouping resources by server/path for the benefit of avoiding individual preflight requests may have benefit if the client was loading many (hundreds) of images on the same page, but this is not done with this client AFAIK.  

Regarding `mixed content` and HTTPS, I am not clear on what the exact problem is.  I have tested running TYPO3 on HTTPS and getting images from HTTP and this just throws warnings in the browser, but nothing fatal.  I think that mixed content is unavoidable, unless you blacklist HTTP resources when you initially parse the manifest.

albig: The problem with `mixed_content` ist, hat running TYPO3 on HTTPS fetching images via HTTP the isCorsTest always fails as the ajax-call is rejected. But that's not the intention of this function.
Fetching the HEAD is IMHO only possible via server as we have done it before your patch.

We only need this isCorsEnabled to be sure, the image manipulation on WebGL will work. But this even works with `mixed_contents` if the CORS-headers are set.

Your current idea does not work for static images as you fetch the HEAD of the server via eID-script not of the image.

Maybe you get my point by an example:

Call it via HTTP or HTTPS - the images are provided with CORS header via HTTP:

http://test.dfg-viewer.de/show/?tx_dlf%5Bid%5D=http%3A%2F%2Fdigital.slub-dresden.de%2Ffileadmin%2Fdata%2F351357262%2Faufdesun_351357262_mets.xml&no_cache=1

That's what isCorsEnabled fetches and always returns false:

<pre>
Request URL:https://test.dfg-viewer.de/show/?eID=tx_dlf_geturl_eid&url=http%3A%2F%2Fdigital.slub-dresden.de%2Ffileadmin%2Fdata%2F351357262%2F351357262_tif%2Fjpegs%2F00000001.tif.large.jpg&header=2
Request Method:HEAD
Status Code:200 OK
Referrer Policy:no-referrer-when-downgrade
</pre>

christopher-johnson: I think I get what you are saying, and I now see that there are several problems with the intent of this function (that I guess is to preflight CORS requests for WebGL?).  I agree that the `HEAD` approach does not work to fix it either.
1. The resource url _MUST_ be called by XHR directly and not through the eID.  This is the only way for the browser to get the CORS response headers directly from the server.  
2.  `HEAD ` or `GET` are not a substitute for `OPTIONS` and cannot return values to the client for the CORS response headers (because they are explicitly filtered by XHR).  
3.  A **preflight enabled** CORS server _MUST_ also include a header like this:
`Access-Control-Allow-Methods →HEAD, GET, OPTIONS`.  Without the explicit declaration, a preflight `OPTIONS` is not possible!   

I suspect that the preferred way to check whether or not a server supports CORS in a preflight is to let the browser handle this. My feeling is that the if server needs to be preflighted, then the right CORS headers should be implemented by the server in accordance with the specifications, and this definitely includes allowing the access-control `OPTIONS` method.   

Using server side PHP to prefetch the resource and parse the headers accomplishes what you want to do I guess, but fetching the images twice does not seem worth it.  I have reverted this patch to the original state of the function, as I do not think that I can fix this.

albig: Thank you for your investigation.

Indeed our server currently supports only `Access-Control-Allow-Methods GET`. This should be no law but it seems to be a setting which is commonly used (if you can say "commonly" in case of CORS anyway).

I'm not happy with this construction with server side PHP prefetching the header. And I agree, it should be done in browser context. But it must work in `mixed_content` conditions too.

For now, I would suggest to merge this patch and investigate this further. Maybe you want to open an issue?

PR #200, opened 2017-07-14
fix #199 roll back to ECMAScript 5

### LINKED ISSUE #199: javascript utilities of pageview plugin breaks IE11 support
Unfortunately IE11 does not support ECMAScript 6 features. See [this compatibility check](http://kangax.github.io/compat-table/es6/).

Introduced with commit a8be6e185d2ad6d25527912bba92a094b89f3071, the file tx_dlf_utils.js use e.g. [const in for loop iteration scope](http://kangax.github.io/compat-table/es6/#test-const_for-in_loop_iteration_scope) which is not supported in IE11. IE11 breaks with an error and the whole pageview is broken as no images are loaded.

As IE11 is still used by 16% our visitors (counted by Piwik at SLUB Dresden), we need to roll back to ECMAScript 5 compatible syntax.

albig:
This patch was done using babel [1] with the following syntax:

babel tx_dlf_utils.js --presets=es2015

[1] https://github.com/babel/babel

### DIFF
--- a/dlf/plugins/pageview/tx_dlf_utils.js
+++ b/dlf/plugins/pageview/tx_dlf_utils.js
@@ -68,7 +68,17 @@ dlfUtils.createOl3Layers = function (imageSourceObjs, opt_origin) {
             });
         } else if (imageSourceObj.mimetype === dlfUtils.CUSTOM_MIMETYPE.IIIF) {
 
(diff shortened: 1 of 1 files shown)


sebastian-meyer:
I can get behind the idea of reverting from `const` and `let` back to `var`. But the formatting changes seem unnecessary and imho just make the code harder to read. Can we skip them or is there a good reason for having those statements in one long line?

albig:
No, we don't need those online statements. This is just the result of the babel tool. I'll rework it.

sebastian-meyer:
Some indents are still odd, but I'll merge it anyway.

PR #201, opened 2017-07-29
Fix audioplayer when used together with pageview plugin on one page

albig:


sebastian-meyer:
Closing this since the proposed changes are part of #205 as well.

Issue #178, opened 2017-05-29
LABEL of physical structMap is ignored; ORDERLABEL is used instead


albig:
The [METS application profile](http://dfg-viewer.de/fileadmin/groups/dfgviewer/METS_application_profile_2.3.pdf) Version 2.3 describes the  physical document structure as follows:

> The ORDERLABEL attribute can contain the pagination of the
> individual page according to the original.
> The LABEL attribute can include the foliation of the individual
> page or the specification of record side and track number.

In Kitodo.Presentation, the function _getPhysicalPages() ignores the LABEL and sets "label" as ORDERLABEL if present.

https://github.com/kitodo/kitodo-presentation/blob/master/dlf/common/class.tx_dlf_document.php#L1930

I would suggest, to use LABEL if present, else ORDERLABEL if present, else empty string.

sebastian-meyer:
We should set ORDERLABEL and LABEL in _getPhysicalPages() and use whatever is reasonable later.
________________________________
Von: Alexander Bigga
Gesendet: Mon May 29 12:43:49 GMT+02:00 2017
An: kitodo/kitodo-presentation
Cc: Subscribed
Betreff: [kitodo/kitodo-presentation] LABEL of physical structMap is ignored; ORDERLABEL is used instead (#178)


The METS application profile<http://dfg-viewer.de/fileadmin/groups/dfgviewer/METS_application_profile_2.3.pdf> Version 2.3 describes the physical document structure as follows:

The ORDERLABEL attribute can contain the pagination of the
individual page according to the original.
The LABEL attribute can include the foliation of the individual
page or the specification of record side and track number.

In Kitodo.Presentation, the function _getPhysicalPages() ignores the LABEL and sets "label" as ORDERLABEL if present.

I would suggest, to use LABEL if present, else ORDERLABEL if present, else empty string.

sebastian-meyer:
See #205: `tx_dlf_document->physicalStructureInfo` now contains `@LABEL` and `@ORDERLABEL`.

PR #205, opened 2017-08-01
Build physical structure out of any type of element, not only pages

sebastian-meyer:
This should fix the audioplayer support by refactoring the physical structure handling. Now alle DIVs of the physical structure are considered, not only those of type `page`. This works for `track` and should work for future types as well.

The pull request contains renaming the former `tx_dlf_document->physicalPages` and `tx_dlf_document->physicalPagesInfo` arrays to `tx_dlf_document->physicalStructure` and `tx_dlf_document->physicalStructureInfo` respectively. This affects some other plugins as well.

The pull request also contains the changes proposed in pull request #201, which is obsolete now.

### DIFF
--- a/dlf/common/class.tx_dlf_document.php
+++ b/dlf/common/class.tx_dlf_document.php
@@ -173,29 +173,29 @@ final class tx_dlf_document {
 	protected $parentId = 0;
 
 	/**
-	 * This holds the physical pages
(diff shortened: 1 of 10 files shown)


albig:
Seems to work as expected.

sebastian-meyer:
This should fix #178, too.

PR #207, opened 2017-08-16
Fix defects reported by Coverity

sebastian-meyer:
See https://scan7.coverity.com/reports.htm#v24048/p13394/fileInstanceId=8721432&defectInstanceId=2174808&mergedDefectId=1398034


albig:
I've just checked this branch with several newspapers. The behaviour is the same as with current master.

albig (approved):
May be merged in my opinion.

PR #211, opened 2017-09-06
Link authors to given valueURI

albig:
This is a proposal to link authors to given valueURI.


sebastian-meyer (approved):
Someday we should get rid of the special handling of some metadata fields in `tx_dlf_mods` (or `tx_dlf_tei` respectively). Historically this was done, because the PHP XML library didn't support XPath 2.0, which is needed for more complex queries like for `author`.

So, for now this is in line with the general metadata handling and can be merged.

PR #212, opened 2017-09-11
add image download tool

albig:
This patch adds a new download tool for the toolbox plugin.

As sub-tools have no flexform configuration, the configuration of the
allowed fileGrps for download is put to toolbox flexform. As done in the
pageview plugin, the best found fileGrp will be used.

Two links will be generated in double mode with the translated labels
"Download left page (JPG)" and "Download right page (JPG)". In single
view mode the label is "Download single page (JPG)". The mime-type is
detected and appended to the label in case of JPG and TIFF.

### DIFF
--- a/dlf/ext_localconf.php
+++ b/dlf/ext_localconf.php
@@ -52,7 +52,7 @@ $GLOBALS['TYPO3_CONF_VARS']['SC_OPTIONS']['dlf/plugins/toolbox/tools'][\TYPO3\CM
 \TYPO3\CMS\Core\Utility\ExtensionManagementUtility::addPItoST43($_EXTKEY, 'plugins/toolbox/tools/pdf/class.tx_dlf_toolsPdf.php', '_toolsPdf', '', TRUE);
 $GLOBALS['TYPO3_CONF_VARS']['SC_OPTIONS']['dlf/plugins/toolbox/tools'][\TYPO3\CMS\Core\Utility\ExtensionManagementUtility::getCN($_EXTKEY).'_toolsPdf'] = 'LLL:EXT:dlf/locallang.xml:tx_dlf_toolbox.toolsPdf';
 
-\TYPO3\CMS\Core\Utility\ExtensionManagementUtility::addPItoST43($_EXTKEY, 'plugins/toolbox/tools/pdf/class.tx_dlf_toolsImagedownload.php', '_toolsImagedownload', '', TRUE);
+\TYPO3\CMS\Core\Utility\ExtensionManagementUtility::addPItoST43($_EXTKEY, 'plugins/toolbox/tools/imagedownload/class.tx_dlf_toolsImagedownload.php', '_toolsImagedownload', '', TRUE);
 $GLOBALS['TYPO3_CONF_VARS']['SC_OPTIONS']['dlf/plugins/toolbox/tools'][\TYPO3\CMS\Core\Utility\ExtensionManagementUtility::getCN($_EXTKEY).'_toolsImagedownload'] = 'LLL:EXT:dlf/locallang.xml:tx_dlf_toolbox.toolsImagedownload';
 
 // Register hooks.
--- a/dlf/ext_localconf.php
+++ b/dlf/ext_localconf.php
@@ -52,6 +52,9 @@ $GLOBALS['TYPO3_CONF_VARS']['SC_OPTIONS']['dlf/plugins/toolbox/tools'][\TYPO3\CM
 \TYPO3\CMS\Core\Utility\ExtensionManagementUtility::addPItoST43($_EXTKEY, 'plugins/toolbox/tools/pdf/class.tx_dlf_toolsPdf.php', '_toolsPdf', '', TRUE);
 $GLOBALS['TYPO3_CONF_VARS']['SC_OPTIONS']['dlf/plugins/toolbox/tools'][\TYPO3\CMS\Core\Utility\ExtensionManagementUtility::getCN($_EXTKEY).'_toolsPdf'] = 'LLL:EXT:dlf/locallang.xml:tx_dlf_toolbox.toolsPdf';
 
+\TYPO3\CMS\Core\Utility\ExtensionManagementUtility::addPItoST43($_EXTKEY, 'plugins/toolbox/tools/pdf/class.tx_dlf_toolsImagedownload.php', '_toolsImagedownload', '', TRUE);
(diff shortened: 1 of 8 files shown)


dlf/plugins/toolbox/tools/imagedownload/class.tx_dlf_toolsImagedownload.php
```diff
@@ -0,0 +1,154 @@
...
+ * Tool 'Image Download' for the plugin 'DLF: Toolbox' of the 'dlf' extension.
+ *
+ * @author	Alexander Bigga <alexander.bigga@slub-dresden.de>
+ * @package	TYPO3
+ * @subpackage	tx_dlf
+ * @access	public
+ */
+class tx_dlf_toolsImagedownload extends tx_dlf_plugin {
+
+	public $scriptRelPath = 'plugins/toolbox/tools/pdf/class.tx_dlf_toolsImagedownload.php';
```
sebastian-meyer: There is a typo in the path. It should be "plugins/toolbox/tools/**imagedownload**/" instead of "plugins/toolbox/tools/**pdf**/".

dlf/ext_localconf.php
```diff
@@ -52,6 +52,9 @@
 \TYPO3\CMS\Core\Utility\ExtensionManagementUtility::addPItoST43($_EXTKEY, 'plugins/toolbox/tools/pdf/class.tx_dlf_toolsPdf.php', '_toolsPdf', '', TRUE);
 $GLOBALS['TYPO3_CONF_VARS']['SC_OPTIONS']['dlf/plugins/toolbox/tools'][\TYPO3\CMS\Core\Utility\ExtensionManagementUtility::getCN($_EXTKEY).'_toolsPdf'] = 'LLL:EXT:dlf/locallang.xml:tx_dlf_toolbox.toolsPdf';
 
+\TYPO3\CMS\Core\Utility\ExtensionManagementUtility::addPItoST43($_EXTKEY, 'plugins/toolbox/tools/pdf/class.tx_dlf_toolsImagedownload.php', '_toolsImagedownload', '', TRUE);
```
sebastian-meyer: There is a typo in the path. It should be "plugins/toolbox/tools/**imagedownload**/" instead of "plugins/toolbox/tools/**pdf**/".

dlf/plugins/toolbox/tools/imagedownload/class.tx_dlf_toolsImagedownload.php
```diff
@@ -0,0 +1,154 @@
...
+
+		}
+
+		return $imageLink;
+
+	}
+
+}
+
+if (defined('TYPO3_MODE') && $TYPO3_CONF_VARS[TYPO3_MODE]['XCLASS']['ext/dlf/plugins/toolbox/tools/pdf/class.tx_dlf_toolsImagedownload.php'])	{
```
sebastian-meyer: There is a typo in the path. It should be "plugins/toolbox/tools/**imagedownload**/" instead of "plugins/toolbox/tools/**pdf**/".

dlf/plugins/toolbox/tools/imagedownload/class.tx_dlf_toolsImagedownload.php
```diff
@@ -0,0 +1,154 @@
...
+	include_once($TYPO3_CONF_VARS[TYPO3_MODE]['XCLASS']['ext/dlf/plugins/toolbox/tools/pdf/class.tx_dlf_toolsImagedownload.php']);
```
sebastian-meyer: There is a typo in the path. It should be "plugins/toolbox/tools/**imagedownload**/" instead of "plugins/toolbox/tools/**pdf**/".

dlf/plugins/toolbox/tools/imagedownload/class.tx_dlf_toolsImagedownload.php
```diff
@@ -0,0 +1,154 @@
...
+		} else {
+
+			$this->template = $this->cObj->getSubpart($this->cObj->fileResource('EXT:dlf/plugins/toolbox/tools/imagedownload/template.tmpl'), '###TEMPLATE###');
+
+		}
+
+		// Get single page downloads.
+		$markerArray['###IMAGE_LEFT###'] =  $this->piVars['double'] == 1 ? $this->getImage($this->piVars['page'], $this->pi_getLL('leftPage', '')) : $this->getImage($this->piVars['page'], $this->pi_getLL('singlePage', ''));
+
+		// Get work download.
```
sebastian-meyer: This comment doesn't make sense. The following code doesn't provide a work download, but the right image download in double page view.

dlf/plugins/toolbox/tools/imagedownload/locallang.xml
```diff
@@ -0,0 +1,30 @@
...
+	<meta type="array">
+		<type>module</type>
+		<description>Language labels for tool tx_dlf_toolsImagedownload</description>
+	</meta>
+	<data type="array">
+		<languageKey index="default" type="array">
+			<label index="singlePage">Download single page</label>
+			<label index="leftPage">Download left page</label>
+			<label index="rightPage">Download right page</label>
+			<label index="work">Download whole work</label>
```
sebastian-meyer: This label isn't used anywhere in the code.

dlf/plugins/toolbox/tools/imagedownload/locallang.xml
```diff
@@ -0,0 +1,30 @@
...
+		</languageKey>
+		<languageKey index="de" type="array">
+			<label index="singlePage">Einzelseite herunterladen</label>
+			<label index="leftPage">Linke Seite herunterladen</label>
+			<label index="rightPage">Rechte Seite herunterladen</label>
+			<label index="work">Ganzes Werk herunterladen</label>
```
sebastian-meyer: This label isn't used anywhere in the code.

sebastian-meyer (changes requested):
There are a few typos and errors in the code which should be fixed.

In general I don't like the fact, that the tool's configuration options are added to the flexform of the toolbox, since that means that there is no easy way of adding new tools with configuration without changing the dlf extension. But I don't see a better way to do it, so I guess it is fine for now. ;o)

dlf/plugins/toolbox/tools/imagedownload/class.tx_dlf_toolsImagedownload.php
albig: We should drop this whole XCLASS support somewhen as nobody uses it and it does'nt work since 6.0. But that's another story ;-)

albig:
I don't like the configuration by flexform of subtools neither. We could do this by TypoScript but this has other drawbacks.

PR #213, opened 2017-09-25
Prevent memory exhaustion while indexing large collections

sebastian-meyer:


sebastian-meyer:
Now bad documents are skipped instead of completly aborting indexing.

albig (approved):
Looks good to me.

PR #223, opened 2017-10-20
Update installation instructions for Apache Solr

Erikmitk:
Also renaming of conf/context.xml to conf/solr.xml.

The name of the file is relevant to access the installation in Tomcat. It might be tricky since the Solr has it’s own solr.xml for configuration purposes, but the name of the file is relevant for Tomcat to generate the path in the URL to access the installation.

Was: context.xml -> localhost:8080/context
Now: solr.xml -> localhost:8080/solr

That wasn’t clear in the old instruction.


sebastian-meyer:
There is a small error in Readme.txt: There are two steps number 3. ;o)
Also Solr should be written with a capital S since it is a product name.

sebastian-meyer:
Thank you!

PR #228, opened 2017-11-08
Remove all XCLASS code references

### LINKED ISSUE #225: Remove all XCLASS code from Kitodo.Presentation
Every class in Kitodo.Presentation has the possibility to overwrite it by the XCLASS mechanism of TYPO3. 

The current format doesn't work in TYPO3 6.x anymore. **Obviously nobody uses this feature** so we should remove it completly.

sebastian-meyer:
This should fix #225

### DIFF
--- a/dlf/class.ext_update.php
+++ b/dlf/class.ext_update.php
@@ -191,7 +191,3 @@ class ext_update {
 	}
 
 }
-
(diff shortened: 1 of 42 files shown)


albig (approved):
Looks good to me.

PR #229, opened 2017-11-16
Adds the option to jump directly to the pageview after the search.

frank-ulrich-weber:
In case of exactly one search result the search plugin can be configured
to open the pageview instead of the listview.

### DIFF
--- a/dlf/plugins/search/class.tx_dlf_search.php
+++ b/dlf/plugins/search/class.tx_dlf_search.php
@@ -654,16 +654,31 @@ class tx_dlf_search extends tx_dlf_plugin {
 			// Clean output buffer.
 			\TYPO3\CMS\Core\Utility\GeneralUtility::cleanOutputBuffers();
 
-			// Keep some plugin variables.
-			$linkConf['parameter'] = $this->conf['targetPid'];
+			// Jump directly to the page view, if there is only one result and it is configured
+			if($results->count() == 1 && !empty($this->conf['showSingleResult'])) {
 
-			if (!empty($this->piVars['order'])) {
+				$linkConf['parameter'] = $this->conf['targetPidPageView'];
 
 				$linkConf['additionalParams'] = \TYPO3\CMS\Core\Utility\GeneralUtility::implodeArrayForUrl($this->prefixId,
-						array (
-							'order' => $this->piVars['order'],
-							'asc' => (!empty($this->piVars['asc']) ? '1' : '0')
-						), '', TRUE, FALSE);
+					array (
+						'id' => $results->current()['uid'],
+						'page' => 1
+					), '', TRUE, FALSE);
+
+			} else {
+
+				// Keep some plugin variables.
+				$linkConf['parameter'] = $this->conf['targetPid'];
+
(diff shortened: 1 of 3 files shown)


dlf/plugins/search/class.tx_dlf_search.php
```diff
@@ -654,16 +654,31 @@ public function main($content, $conf) {
...
+				$linkConf['parameter'] = $this->conf['targetPidPageView'];
 
 				$linkConf['additionalParams'] = \TYPO3\CMS\Core\Utility\GeneralUtility::implodeArrayForUrl($this->prefixId,
-						array (
-							'order' => $this->piVars['order'],
-							'asc' => (!empty($this->piVars['asc']) ? '1' : '0')
-						), '', TRUE, FALSE);
+					array (
+						'id' => $results->current()['uid'],
+						'page' => 1
```
sebastian-meyer: If the only search result was a fulltext hit, the target should be a specific page, not just page 1.

I suggest checking if the search result references a page number and then using this - otherwise page 1.

sebastian-meyer (changes requested):
Apart from one little suggestion this looks good to me!

dlf/plugins/search/flexform.xml
```diff
@@ -220,6 +220,17 @@
...
 						</TCEforms>
 					</suggest>
+					<showSingleResult>
+						<TCEforms>
+							<onChange>reload</onChange>
+							<exclude>1</exclude>
+							<label>LLL:EXT:dlf/plugins/search/locallang.xml:tt_content.pi_flexform.showSingleResult</label>
+							<config>
+								<type>check</type>
+								<default>1</default>
```
sebastian-meyer: I'd suggest turning this off as default. That is to prevent someone updating the extension from accidently activating a new feature he/she doesn't want. Also, the way the TYPO3 flexform configuration works, the configuration would have to be rewritten to actually change the settings although it would be falsely displayed as activated.

dlf/plugins/search/flexform.xml
```diff
@@ -240,6 +251,21 @@
 							</config>
 						</TCEforms>
 					</targetPid>
+					<targetPidPageView>
+						<TCEforms>
+							<displayCond>FIELD:showSingleResult:=:0</displayCond>
```
sebastian-meyer: This option should be shown when `showSingleResult` is set to `true`, right? Then its `displayCond` should be `FIELD:showSingleResult:=:1`.

sebastian-meyer (changes requested):
There is another little thing to consider.

dlf/plugins/search/flexform.xml
frank-ulrich-weber: I see... then I change the logic "Show single result in listview" to "Show single result in pageview". Then the new feature can be disabled by defult.

dlf/plugins/search/flexform.xml
frank-ulrich-weber: No, it's the other way round. The change to "Show single result in pageview" fixes this, too.

PR #233, opened 2017-11-22
Multiple performance optimizations

frank-ulrich-weber:
The optimizations focus on tx_dlf_search, tx_dlf_list and tx_dlf_listview

### DIFF
--- a/dlf/common/class.tx_dlf_solr.php
+++ b/dlf/common/class.tx_dlf_solr.php
@@ -343,36 +343,38 @@ class tx_dlf_solr {
 	 */
 	public function search($query = '') {
 
-		// Perform search.
-		$results = $this->service->search((string) $query, 0, $this->limit, $this->params);
-
-		$this->numberOfHits = count($results->response->docs);
-
 		$toplevel = array ();
 
 		$checks = array ();
 
+		// Restrict the fields to the required ones
+		$this->params['fl'] = "uid,id,toplevel";
+
 		// Get metadata configuration.
-		if ($this->numberOfHits > 0) {
+		$result = $GLOBALS['TYPO3_DB']->exec_SELECTquery(
+			'tx_dlf_metadata.index_name AS index_name',
+			'tx_dlf_metadata',
+			'tx_dlf_metadata.is_sortable=1 AND tx_dlf_metadata.pid='.intval($this->cPid).tx_dlf_helper::whereClause('tx_dlf_metadata'),
+			'',
+			'',
+			''
+		);
 
-			$result = $GLOBALS['TYPO3_DB']->exec_SELECTquery(
-				'tx_dlf_metadata.index_name AS index_name',
-				'tx_dlf_metadata',
-				'tx_dlf_metadata.is_sortable=1 AND tx_dlf_metadata.pid='.intval($this->cPid).tx_dlf_helper::whereClause('tx_dlf_metadata'),
-				'',
-				'',
-				''
-			);
+		$sorting = array ();
 
-			$sorting = array ();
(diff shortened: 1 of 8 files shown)


dlf/common/class.tx_dlf_solr.php
```diff
@@ -343,37 +343,39 @@ public static function solrGetCoreNumber($start = 0) {
...
 
-			while ($resArray = $GLOBALS['TYPO3_DB']->sql_fetch_assoc($result)) {
+		while ($resArray = $GLOBALS['TYPO3_DB']->sql_fetch_assoc($result)) {
 
-				$sorting[$resArray['index_name']] = $resArray['index_name'].'_sorting';
+			$sorting[$resArray['index_name']] = $resArray['index_name'].'_sorting';
 
-			}
+			// Add *_sorting fields to the required fields list
+			empty($this->params['fl']) ? $this->params['fl'] = $sorting[$resArray['index_name']].',' : $this->params['fl'] .= ','.$sorting[$resArray['index_name']];
```
sebastian-meyer, line 370: There should be no trailing `,` in the first case (empty `$this->params['fl']`).

dlf/plugins/listview/class.tx_dlf_listview.php
```diff
@@ -268,6 +284,27 @@ protected function getEntry($number, $template) {
 
 	}
 
+	/**
+	 * Returns the fieldwrap of a metadatum
+	 *
+	 * @access	private
+	 *
+	 * @return	array		The parsed fildwrap
```
sebastian-meyer, line 292: Please document the function's parameters according to PhpDoc.

sebastian-meyer: And there is a typo in "fi**e**ldwrap" :)

dlf/plugins/search/class.tx_dlf_search.php
```diff
@@ -537,11 +562,6 @@ public function main($content, $conf) {
 
 				}
 
-				// Add highlighting for fulltext.
-				$params['hl'] = 'true';
-
-				$params['hl.fl'] = 'fulltext';
-
```
sebastian-meyer, line 544: Does highlighting still work without using the `fastVectorHighlighter` when removing those lines?

dlf/plugins/search/class.tx_dlf_search.php
```diff
@@ -762,7 +777,7 @@ public function makeFacetsMenuArray($content, $conf) {
 		$search['params']['facet.limit'] = $this->conf['limitFacets'];
 
 		// Perform search.
-		$results = $solr->service->search($search['query'], 0, $this->conf['limit'], $search['params']);
+		$results = $solr->service->search($search['query'], 0, 0, $search['params']);
```
sebastian-meyer, line 781: You are removing the result limitation for Solr which could potentially lead to big result sets. Is this intentional?

sebastian-meyer (changes requested):
There are a few issues which you should fix before merging.
You also need to do a rebase, because otherwise your changes would conflict with the already merged feature #229

dlf/plugins/search/class.tx_dlf_search.php
frank-ulrich-weber: Yes, because those lines have been moved to tx_dlf_list->getRecord(). Highlights are only fetched for the record and not for the whole search result.

dlf/plugins/search/class.tx_dlf_search.php
frank-ulrich-weber: I didn't remove the limit. The limit is set to 0 because we don't need any result beside the facets.

frank-ulrich-weber:
Adapted and rebased

PR #235, opened 2017-12-01
Repeat search if the result is an empty string.

frank-ulrich-weber:
Otherwise this ends up in a TYPO3 error page.

### DIFF
--- a/dlf/common/class.tx_dlf_solr.php
+++ b/dlf/common/class.tx_dlf_solr.php
@@ -346,6 +346,14 @@ class tx_dlf_solr {
 		// Perform search.
 		$results = $this->service->search((string) $query, 0, $this->limit, $this->params);
 
+		// In rare unclear cases $result can be an empty string instead of an Apache_Solr_Response.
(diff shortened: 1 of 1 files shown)


sebastian-meyer:
This seems to be a really dirty workaround for a problem, which is most likely related to `Apache Solr` or `SolrPhpClient`.

Could you please further debug this issue and then create a proper fix?

frank-ulrich-weber:
Yes, it is a dirty workaroud :-) I traced the problem and it is not the SolrPhpClient. I used directly the solr-url to do a search and I repeated the search automatically a thousands times. The error occurs after 800, 600,... searches on a machine or not on another machine. Rebuilding the index did not solve the problem. The configuration and version of solr is the same but the machine configuration (os, java,...) differs. I will not debug solr nor setup a test environment for this rare bug. I think the fix does nothing bad if everything works fine and bypasses (ok dirty:-) the problem in rare cases. If you still don't like to merge this pull-request please close it.

PR #237, opened 2017-12-05
Multiple performance optimizations

frank-ulrich-weber:


sebastian-meyer:
Since we have to make mandatory changes to the Solr schema anyway (see #244), I suggest adding the `FastVectorHighlighter` to the default schema instead of making it an option.

frank-ulrich-weber:
New pull request #259 available for new directory structure and changed indention

Issue #239, opened 2017-12-12
Javascript Error in pageview plugin (basket feature)


albig:
In current master, there is a javascript error in tx_dlf_pageview.js used by pageview plugin:

`Uncaught TypeError: Cannot read property 'condition' of undefined (tx_dlf_pageview.js:424)`

![kitodo_presentation_pageview_error](https://user-images.githubusercontent.com/3354500/33897539-00889e2c-df66-11e7-839e-c1887af79c37.png)

christopher-timm:
This Javascript error occurs if the OpenLayer build has not all required classes.
It wasn't possible for me to build the ol3-dlf.js with the ol3-dlf.json. But I included the required classes in the ol3-dlf.json. Unfortunately without testing.
On the other hand with the full ol3.js this error doesn't occur.

albig:
Ok. So we need to rebuild the openlayers build before the next release.

albig:
@frank-ulrich-weber If you fix this issue: please follow the instruction in Readme.md: https://github.com/kitodo/kitodo-presentation/blob/master/dlf/lib/OpenLayers/README.md
We need to use the SLUB fork and should investigate later on, if we could upgrade OL3.

sebastian-meyer:
Fixed with #242

PR #242, opened 2017-12-14
Fix basket

frank-ulrich-weber:
Fixes some small issues within the basket-feature and adds a suitable custom build of openlayers based on https://github.com/slub/ol3/tree/kitodo-presentation


albig (approved):
Commit 23f005f fixes issue #239 for me.

PR #244, opened 2017-12-14
Refactor and extend OAI Plugin

Erikmitk:
This pull request contains refactoring and extension of `dlf/plugins/oai/class.tx_dlf_oai.php`.

The methods `verbListIdentifiers()`, `verbListRecords()` and `resume()` basically contained the same logic and duplicated a bunch of code. I extracted those parts into `fetchDocumentUIDs()`, `generateOutputForDocumentList(…)` and `generateResumptionTokenForDocumentListSet(…)`.

`verbListIdentifiers()` and `verbListRecords()` follow the same basic structure.
1. check piVars
2. `fetchDocumentUIDs()`
3. Create a resultSet and generate output for it via `generateOutputForDocumentList()`
	1. if there are documents left a resumption token will be generated and added to the output.

`resume()` acts accordingly but gets its documentSet via resumptionToken from the database.

The extracted method `fetchDocumentUIDs`is now able to get documents via Solr query. To be able to query all output formats the Solr schema had to be extended by record_id, purl, urn and location. Those are required fields in the various output formats.  The collection field is now required as well.

The default case for exposed OAI sets is to query via the index_name of the collection (ie. collection:[index_name]). You are also able to specify a Solr query for a given collection in the TYPO3 backend. This enables to define a set by arbitrary conditions (e.g. add two collections together or restrict a set by specific parameters). The Solr query has precedence. For the Solr query itself I added a raw search method in `dlf/common/class.tx_dlf_solr.php` to fetch only what I need from Solr.

Currently, the 'old' database way of generating OAI is still supported and set as the default. You can change this behaviour by checking the Solr usage setting in the OAI plugin. To use Solr queries you have to re-index your collections due to the Solr schema changes.

Additionally, the handling of resumption tokens has been improved. We now only save the remaining document UIDs per resumption token in the database and not the whole set. I added a helper method in `dlf/common/class.tx_dlf_list.php` to remove a range of documents from a `tx_dlf_list`. That should improve database performance. The UIDs that have to be stored and fetched get less with every further call.

I started by taking the changes @albig made in https://github.com/albig/kitodo-presentation/tree/collections-oai-sor which contained some tangential changes.


dlf/common/class.tx_dlf_indexing.php
```diff
@@ -591,6 +591,16 @@ protected static function processLogical(tx_dlf_document &$doc, array $logicalUn
...
+			$solrDoc->setField('record_id', $metadata['record_id'][0]);
+
+			$solrDoc->setField('purl', $metadata['purl'][0]);
+
+			$solrDoc->setField('location',$doc->location);
+
+			$solrDoc->setField('urn', $metadata['urn']);
+
+			$solrDoc->setField('collection', $metadata['collection']);
+
```
sebastian-meyer, line 603: This doesn't work with the old index schema, because those fields aren't defined.

dlf/Configuration/TCA/tx_dlf_solrcores.php
```diff
@@ -42,7 +42,7 @@
 		'index_name' => array (
 			'label' => 'LLL:EXT:dlf/locallang.xml:tx_dlf_solrcores.index_name',
 			'config' => array (
-				'type' => 'none',
+				'type' => 'input',
```
sebastian-meyer, line 45: It is important that the `index_name` is read-only (at least for collections added automatically by the indexer), that's why it was of type `none`. The `index_name` holds the exact phrase used in the document's METS files, so in order to correctly  identify the collections a document is associated with this must not be changed.

dlf/common/class.tx_dlf_solr.php
```diff
@@ -136,7 +136,7 @@ public static function escapeQuery($query) {
 	public static function escapeQueryKeepField($query, $pid) {
 
 		// Is there a field query?
-		if (preg_match('/^[[:alnum:]]+_[tu][su]i:\(.*\)$/', $query)) {
+		if (preg_match('/^[[:alnum:]]+_[tu][su]i:(.*)$/', $query)) {
```
sebastian-meyer: AFAIK the Solr query syntax is `[field name]:([search term])`, so the parentheses are part of the query syntax, not a regex subexpression. Therefore they should be treated as literal characters and must be masked. Or am I wrong?

dlf/lib/ApacheSolr/conf/schema.xml
```diff
@@ -1,7 +1,7 @@
 <?xml version="1.0" encoding="UTF-8" ?>
 <!--
 Licensed to the Apache Software Foundation (ASF) under one or more
-contributor license agreements.  See the NOTICE file distributed with
+contributor license agreements.	 See the NOTICE file distributed with
```
sebastian-meyer: Just a small annoyance: this tab should stay a whitespace or be removed! :)

sebastian-meyer (changes requested):
I suggest making the schema changes mandatory and thus the definition of OAI sets by Solr query the new standard (removing the old DB based handling). That's because I think keeping both options would require a lot more work (and doesn't work with this pull request).

Here's why:
- The indexer would need to get fixed, because now it tries to index fields which don't exist in the old schema (`record_id`, `purl`, `location`, `urn`, `collection`). This will lead to errors.
- Defining a new collection by building a Solr query works fine for OAI sets, but doesn't work when trying to display that collection via `tx_dlf_collection` plugin. That's inconsistent and unintuitive.

I therefore suggest:
- to add a default query to each existing collection: `collection:("$index_name")` (where `$index_name` is the actual `index_name` of the collection)
- to adjust the indexer accordingly when automatically adding new collections (new collections should have the same default query as above)
- to make `tx_dlf_oai` and `tx_dlf_collection` completely rely on the index data (as you already did for `tx_dlf_oai`, just remove the DB fallback)
- to make the new schema mandatory.

The only caveat would be, that having a Solr server would no longer be optional. But I think that's fine, because every known user has one anyway.

dlf/common/class.tx_dlf_solr.php
Erikmitk: I think the question is if the parentheses are optional or not. The syntax is first and foremost [field name]:[search term] but you can optionally group the search term if you want to search for more than one word or phrase in the same field e.g [field name]:([term1] [term2]).

We should change the regex to make the parentheses optional i.e. `\(?.*\)?`. What do you think, @sebastian-meyer?

That might be the case why it didn’t work for @albig. The query should work without the grouping but didn’t because they were expected.

dlf/Configuration/TCA/tx_dlf_solrcores.php
Erikmitk: I think this `index_name` is something different. It’s in the file `tx_dlf_solrcores` and is the name of the Solr core as it is referenced inside the Solr e.g. `dlfCore1`. That should be independent of the collection/document relation.

Should I be able to rename it or not?

Erikmitk:
Due to changes in the master branch I was not able to rebase my code anymore. So I had to create a new pull request instead. See: https://github.com/kitodo/kitodo-presentation/pull/261

PR #261, opened 2018-01-04
Refactor and extend OAI Plugin

Erikmitk:
This is a duplicate of https://github.com/kitodo/kitodo-presentation/pull/244. Due to directory changes and changes in the files in the master branch I was not able to rebase my code. So I had to create this pull request instead.

Content-wise it is identical and I have already made some requested changes from https://github.com/kitodo/kitodo-presentation/pull/244.

For me two open questions remain:

1. Should the regular Expression for field queries enforce parenthesis or make them optional? They are not mandatory for a query so I’d recommend to make them optional here as well.
See: https://github.com/kitodo/kitodo-presentation/pull/244#discussion_r158266669) 
Commit: https://github.com/kitodo/kitodo-presentation/commit/de3000f8aeb4f1a7c709623246f9d8bcb71f15e0

2. Should the internal name of a Solr core be fixed or editable.
See: https://github.com/kitodo/kitodo-presentation/pull/244#discussion_r158270137
Commit: https://github.com/kitodo/kitodo-presentation/commit/31eb134f719b30c55a327719607b7e28a7fc2501

That should be all the changes for the OAI-Plugin. I will open a new PR were I add the Solr queries in the collection plugin.

For completeness the description of https://github.com/kitodo/kitodo-presentation/pull/244:

> This pull request contains refactoring and extension of dlf/plugins/oai/class.tx_dlf_oai.php.
> 
> The methods verbListIdentifiers(), verbListRecords() and resume() basically contained the same logic and duplicated a bunch of code. I extracted those parts into fetchDocumentUIDs(), generateOutputForDocumentList(…) and generateResumptionTokenForDocumentListSet(…).
> 
> verbListIdentifiers() and verbListRecords() follow the same basic structure.
> 
> check piVars
> fetchDocumentUIDs()
> Create a resultSet and generate output for it via generateOutputForDocumentList()
> if there are documents left a resumption token will be generated and added to the output.
> resume() acts accordingly but gets its documentSet via resumptionToken from the database.
> 
> The extracted method fetchDocumentUIDsis now able to get documents via Solr query. To be able to query all output formats the Solr schema had to be extended by record_id, purl, urn and location. Those are required fields in the various output formats. The collection field is now required as well.
> 
> The default case for exposed OAI sets is to query via the index_name of the collection (ie. collection:[index_name]). You are also able to specify a Solr query for a given collection in the TYPO3 backend. This enables to define a set by arbitrary conditions (e.g. add two collections together or restrict a set by specific parameters). The Solr query has precedence. For the Solr query itself I added a raw search method in dlf/common/class.tx_dlf_solr.php to fetch only what I need from Solr.
> 
> Currently, the 'old' database way of generating OAI is still supported and set as the default. You can change this behaviour by checking the Solr usage setting in the OAI plugin. To use Solr queries you have to re-index your collections due to the Solr schema changes.
> 
> Additionally, the handling of resumption tokens has been improved. We now only save the remaining document UIDs per resumption token in the database and not the whole set. I added a helper method in dlf/common/class.tx_dlf_list.php to remove a range of documents from a tx_dlf_list. That should improve database performance. The UIDs that have to be stored and fetched get less with every further call.
> 
> I started by taking the changes @albig made in https://github.com/albig/kitodo-presentation/tree/collections-oai-sor which contained some tangential changes.

### DIFF
--- a/Configuration/TCA/tx_dlf_collections.php
+++ b/Configuration/TCA/tx_dlf_collections.php
@@ -120,6 +120,16 @@ return array (
                 'eval' => 'required,uniqueInPid',
             ),
         ),
+        'index_search' => array (
+            'exclude' => 1,
+            'label' => 'LLL:EXT:dlf/locallang.xml:tx_dlf_collections.index_search',
+            'config' => array (
+            'type' => 'text',
+                'cols' => 30,
+                'rows' => 5,
+                'eval' => '',
+            ),
+        ),
         'oai_name' => array (
             'exclude' => 1,
             'label' => 'LLL:EXT:dlf/locallang.xml:tx_dlf_collections.oai_name',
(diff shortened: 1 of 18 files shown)


Erikmitk:
1. I removed the functionality of renaming a Solr Cor with https://github.com/kitodo/kitodo-presentation/pull/261/commits/24cd44b8efd3d392f51475603a53d6c90430e24f.
2. The regular expression for field queries now treats the parenthesis as optional (https://github.com/kitodo/kitodo-presentation/pull/261/commits/de3000f8aeb4f1a7c709623246f9d8bcb71f15e0).

I added the patches made by Scrutinizer so the PR should be good to go now.

PR #268, opened 2018-02-15
Add Solr queries to collection plugin

Erikmitk:
This pull request adds Solr queries to the collection plugin. Since #261 allows to define virtual collections via a Solr query those collections should be able to be shown in the collection plugin.

To be able to query the Solr an additional field for selecting a Solr core has been added to the plugin configuration.

I cleaned up some code and replaced several database queries for the corresponding Solr queries to get the necessary data.

**CollectionList**
1. fetch information about selected collections from database
2. for each collection
-- query titles and volumes info from Solr
-- generate output

**SingleCollection**
1. fetch collection information from DB
2. fetch corresponding document UIDs from Solr
3. fetch info from those documents from DB
4. generate output

Step 3 could be eliminated. But we would have to make additional changes. This particular query contains the database field metadata_sorting. That field includes all information for sorting the list view if I got that right. I haven’t figured out how the sorting actually works so that might be wrong.

image

We could probably fetch the values for sorting the documents from the index, but with the options in the selector we would have to add `author` and `year` to the Solr schema. At the moment they are only dynamic fields.

### DIFF
--- a/plugins/collection/class.tx_dlf_collection.php
+++ b/plugins/collection/class.tx_dlf_collection.php
@@ -343,8 +343,8 @@ class tx_dlf_collection extends tx_dlf_plugin {
             if (empty($listMetadata)) {
 
                 $listMetadata = array (
-                    'label' => htmlspecialchars($collectionData['collLabel']),
-                    'description' => $this->pi_RTEcssText($collectionData['collDesc']),
+                    'label' => !empty($l10nOverlay['label'])? htmlspecialchars($l10nOverlay['label']) : htmlspecialchars($collectionData['collLabel']),
+                    'description' => !empty($l10nOverlay['description']) ? $this->pi_RTEcssText($l10nOverlay['description']) : $this->pi_RTEcssText($collectionData['collDesc']),
                     'thumbnail' => htmlspecialchars($collectionData['collThumb']),
                     'options' => array (
                         'source' => 'collection',
--- a/plugins/collection/class.tx_dlf_collection.php
+++ b/plugins/collection/class.tx_dlf_collection.php
@@ -95,53 +95,33 @@ class tx_dlf_collection extends tx_dlf_plugin {
      * @return	string		The list of collections ready to output
      */
     protected function showCollectionList() {
-
-        $additionalWhere = '';
-
+        $selectedCollections = 'tx_dlf_collections.uid != 0';
         $orderBy = 'tx_dlf_collections.label';
 
         // Handle collections set by configuration.
         if ($this->conf['collections']) {
 
             if (count(explode(',', $this->conf['collections'])) == 1 && empty($this->conf['dont_show_single'])) {
-
                 $this->showSingleCollection(intval(trim($this->conf['collections'], ' ,')));
-
             }
 
-            $additionalWhere .= ' AND tx_dlf_collections.uid IN ('.$GLOBALS['TYPO3_DB']->cleanIntList($this->conf['collections']).')';
-
+            $selectedCollections = 'tx_dlf_collections.uid IN ('.$GLOBALS['TYPO3_DB']->cleanIntList($this->conf['collections']).')';
             $orderBy = 'FIELD(tx_dlf_collections.uid, '.$GLOBALS['TYPO3_DB']->cleanIntList($this->conf['collections']).')';
-
         }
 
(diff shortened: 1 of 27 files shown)


plugins/collection/class.tx_dlf_collection.php
```diff
@@ -95,53 +95,33 @@ public function main($content, $conf) {
...
-
-        } elseif ($this->conf['show_userdefined'] > 0) {
-
-            if (!empty($GLOBALS['TSFE']->fe_user->user['uid'])) {
-
-                $additionalWhere .= ' AND tx_dlf_collections.fe_cruser_id='.intval($GLOBALS['TSFE']->fe_user->user['uid']);
-
-            } else {
-
-                $additionalWhere .= ' AND NOT tx_dlf_collections.fe_cruser_id=0';
```
sebastian-meyer: Your simplification misses this case ($this->conf['show_userdefined'] > 0 && empty($GLOBALS['TSFE']->fe_user->user['uid']). This case is important, because it is used to show all userdefined collections.

plugins/collection/flexform.xml
```diff
@@ -57,6 +57,21 @@
 							</config>
 						</TCEforms>
 					</collections>
+					<solrcore>
+						<TCEforms>
+							<displayCond>FIELD:pages:REQ:true</displayCond>
+							<exclude>1</exclude>
+							<label>LLL:EXT:dlf/plugins/oai/locallang.xml:tt_content.pi_flexform.solrcore</label>
```
sebastian-meyer: This is the flexform for the collection plugin, but you are referencing the language file of the oai plugin. Please try to avoid cross-plugin dependencies!

plugins/collection/flexform.xml
```diff
@@ -57,6 +57,21 @@
...
+							<config>
+								<type>select</type>
+								<items type="array"></items>
+								<itemsProcFunc>tx_dlf_tceforms->itemsProcFunc_solrList</itemsProcFunc>
+								<size>1</size>
+								<maxitems>1</maxitems>
+								<minitems>0</minitems>
```
sebastian-meyer: I am not sure, but shouldn't this be a mandatory field (i.e. minitems=1)?

plugins/collection/class.tx_dlf_collection.php
```diff
@@ -95,53 +95,33 @@ public function main($content, $conf) {
      * @return	string		The list of collections ready to output
      */
     protected function showCollectionList() {
-
```
sebastian-meyer: Please preserve those empty lines throughout the file (there are many occasions where you removed them).

plugins/collection/class.tx_dlf_collection.php
```diff
@@ -154,135 +134,130 @@ protected function showCollectionList() {
...
-                    $img = '<img src="'.\TYPO3\CMS\Core\Utility\ExtensionManagementUtility::siteRelPath($this->extKey).'res/icons/txdlffeeds.png" alt="'.$this->pi_getLL('feedAlt', '', TRUE).'" title="'.$this->pi_getLL('feedTitle', '', TRUE).'" />';
+            foreach ($partOfNothing as $doc) {
+                $titles[] = $doc->uid;
+            }
 
-                    $markerArray[$_key]['###FEED###'] = $this->pi_linkTP($img, array ($this->prefixId => array ('collection' => $resArray['uid'])), FALSE, $this->conf['targetFeed']);
+            $volumes = $titles;
 
-                } else {
+            // All titles except those that are part of something
```
sebastian-meyer: You confused titles and volumes here. Titles are all documents which do not have an ancestor (i.e. partof == 0), but volumes are all documents which either have an ancestor (i.e. partof != 0) or no descendant (i.e. no other documents exist with descendant.partof == ancestor.uid).
The way you calculate titles and volumes always results in less volumes than titles, but in reality that can never be the case. The number of volumes must be at least as big as the number of titles, almost always it is even bigger.

sebastian-meyer (changes requested):
There are some inconsistencies and errors, please check my inline comments. And please have a look at manually created collections to avoid running into the same bug as with the OAI plugin.

Erikmitk:
I am gonna open up a new request for this, because there were a lot of changes in the mean time and I had to rebase my branch. This PR would be quite confusing.

PR #274, opened 2018-04-23
Fix #273 multilanguage support of collection records

### LINKED ISSUE #273: Localization of collection records only partly implemented
Records of the tx_dlf_collections table are offered as translatable records but this feature is not fully implemented.

- TCA is incomplete, no l10n_mode setting is used
- the recordOverlay isn't fetched in the collection plugin
- the indexer does'nt respect the sys_language_uid of collection records

albig:
This is a solution for #273. Is uses the same approach as the metadata plugin.

Used settings:

* config.sys_language_mode = content_fallback
* config.sys_language_overlay = 1 (or hideNonTranslated)
* add translation to page (collection page AND listview page)
* translate the Kitodo.Presentation plugins
* translate (==copy) a collection record --> set new lanuage, set origin, set label and description

### DIFF
--- a/plugins/collection/class.tx_dlf_collection.php
+++ b/plugins/collection/class.tx_dlf_collection.php
@@ -164,7 +164,7 @@ class tx_dlf_collection extends tx_dlf_plugin {
                 'tx_dlf_documents',
                 'tx_dlf_relations',
                 'tx_dlf_collections',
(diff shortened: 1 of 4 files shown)


plugins/collection/class.tx_dlf_collection.php
```diff
@@ -174,13 +174,25 @@ protected function showCollectionList() {
 
             while ($resArrayVolumes = $GLOBALS['TYPO3_DB']->sql_fetch_assoc($resultVolumes)) {
 
+                if (is_array($resArrayVolumes) && $resArrayVolumes['sys_language_uid'] != $GLOBALS['TSFE']->sys_language_content && $GLOBALS['TSFE']->sys_language_contentOL) {
```
sebastian-meyer: `$resArrayVolumes` has only keys `uid` and `volumes`, but not `sys_language_uid` (see SQL query above). At this point we are only checking the number of volumes per collection, so the language should be irrelevant anyway.

I think, the SQL query above as well as this `while` construct don't need to check for language overlays. The SQL query should only recognize records with associated documents anyways (which is not the case for translation overlay records).

sebastian-meyer (changes requested):
Please see my inline comment.

sebastian-meyer (approved):

PR #281, opened 2018-05-09
Fix language overlay for collections

sebastian-meyer:


albig (approved):
Concerning label and description, this patch still does it's job.

PR #282, opened 2018-05-09
Fix using manually created collections as OAI sets

### LINKED ISSUE #271: OAI sets don't work for manually created collections
Pull request #261 introduced an index based handling of OAI sets. This depends on either having a defined Solr query for each OAI set or searching the index for the `index name` of the collection. Since manually created collections are not referenced in the METS files, they are not indexed and therefore can't be found by searching for the collection's `index name`.
In this case a query for the UIDs of the collection's documents has to be performed instead.

sebastian-meyer:
This should fix #271.

I did not implement automatic reindexing of documents after adding/removing them from collections, because that would be triggered by updating a collection record and would result in reindexing all documents of the collection everytime the collection gets updated...
So after manually creating or updating a collection, one has to remember to reindex that collection via the backend module if one wants to use it as an OAI set.

### DIFF
--- a/common/class.tx_dlf_document.php
+++ b/common/class.tx_dlf_document.php
@@ -910,7 +910,7 @@ final class tx_dlf_document {
                     'tx_dlf_documents',
                     'tx_dlf_relations',
                     'tx_dlf_collections',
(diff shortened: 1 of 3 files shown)


albig (approved):
Works for me as expected.

PR #284, opened 2018-05-10
fix #283: compare value with simple array

### LINKED ISSUE #283: Subparts in Listview are not linked properly and have "no title"
Commit 60f8c21398116a80453a8f929867f93834c0160f replaces the helper function search_array_recursive by PHP search_array. The behaviour of the PHP function is a little different as it does not compare multidimensional arrays.

As result, the condition in Line 226 is always FALSE. "Subparts" are not properly enhanced. In the listview, there is only written "{no title}" with an empty link.

albig:
The comparision is done at this point with the following values/arrays:

$resArray['uid'] = 3123;

$record['subparts'][0]['u'] = 4123;
$record['subparts'][1]['u'] = 2123;
$record['subparts'][2]['u'] = 3123;

array_search(3123, $record['subparts']) --> no result

Solutions:

array_search(3123, array_column($record['subparts'], 'u') --> 2

or

array_search(array('u' => 3123), $record['subparts']) --> 2

### DIFF
--- a/common/class.tx_dlf_list.php
+++ b/common/class.tx_dlf_list.php
@@ -223,7 +223,7 @@ class tx_dlf_list implements ArrayAccess, Countable, Iterator, \TYPO3\CMS\Core\S
 
                         $record['metadata'] = $metadata;
 
-                    } elseif (($key = array_search($resArray['uid'], array_column($record['subparts'], 'u'), TRUE)) !== FALSE) {
+                    } elseif (($key = array_search(array('u' => $resArray['uid']), $record['subparts'], TRUE)) !== FALSE) {
 
                         $record['subparts'][$key] = array (
                             'uid' => $resArray['uid'],
--- a/common/class.tx_dlf_list.php
+++ b/common/class.tx_dlf_list.php
@@ -223,7 +223,7 @@ class tx_dlf_list implements ArrayAccess, Countable, Iterator, \TYPO3\CMS\Core\S
 
                         $record['metadata'] = $metadata;
 
-                    } elseif (($key = array_search($resArray['uid'], $record['subparts'], TRUE)) !== FALSE) {
+                    } elseif (($key = array_search($resArray['uid'], array_column($record['subparts'], 'u'), TRUE)) !== FALSE) {
 
                         $record['subparts'][$key] = array (
                             'uid' => $resArray['uid'],


sebastian-meyer:
I am not sure if variant 2 still works if $record['subparts'] contains arrays with more than one key.
I.e. something like:
$record = array (
    'subparts' => array (
         0 => array ('u' => 4123, 'h' => 'foobar'),
         1 => array ('u' => 2123, 'h' => 'foo'),
         2 => array ('u' => 3123, 'h' => 'bar')
    )
);

albig:
Your're right. Variant 2 does not work with your example. But it works currently in the "Digitale Sammlungen" where variant 1 only partly works :-(

albig:
Actually variant 2 is working, because in this special case, it search only in the remaining subparts.
Variant 1 fails because the subpart entries loose the 'u' key on iteration.

Example:
1st Iteration:
<pre>
$resArray['uid'] = 4123;
$record = array (
    'subparts' => array (
          0 => array ('u' => 4123),
          1 => array ('u' => 2123),
          2 => array ('u' => 3123)
   )
);
--> $key == 0
</pre>

2nd iteration
<pre>

$resArray['uid'] = 2123;
$record = array (
    'subparts' => array (
          0 => array ('uid' => 4123, 'h' => 'foo'), // no 'u' key anymore!
          1 => array ('u' => 2123),
          2 => array ('u' => 3123)
   )
);
--> $key == 0 --> that's oviously wrong.
</pre>

This is, because array_column($record['subparts'], 'u'); makes a new array with index starting with 0 with subparts array, containing the key 'u'.

Variant 1 is only working, if we keep the key 'u' together with 'uid' in the subparts array.

Variant 2 is only working if the subparts array initial only has key 'u'.

sebastian-meyer:
Variant 2 doesn't work when building the listview from a Solr result, because then there will be more keys in the array than just 'u' (see https://github.com/kitodo/kitodo-presentation/blob/master/common/class.tx_dlf_solr.php#L417)

Maybe we should revert to the `array_search_recursive` helper function?

sebastian-meyer:
I just realised that `array_search` is only invoked when dealing with a collection result. Solr results are already handled differently.
Sorry for the confusion! I am going to merge this now.

sebastian-meyer (approved):

PR #285, opened 2018-05-14
Add possibility to use different fulltext formats

### LINKED ISSUE #197: Missing format validation of fulltext parsing
Kitodo.Presentation supports fulltexts in ALTO-XML format. TEI will be supported soon.

On parsing the given file, there is no check the format is supported.

The result are several Javascript errors like this:

`Uncaught TypeError: Cannot read property 'getTextblocks' of undefined`

This should be improved. Maybe together with the TEI-Feature. Introducing this feature, we have to decide, which parser should be used depending on the file format.

sebastian-meyer:
This partially fixes #197 

Fulltext is now handled the same way as metadata loading from `dmdSec` is done. Thus it is possible to use different fulltext formats by simply providing a class implementing the `tx_dlf_fulltext` interface and registering this class via `tx_dlf_formats`.

### DIFF
--- a/common/class.tx_dlf_alto.php
+++ b/common/class.tx_dlf_alto.php
@@ -24,11 +24,11 @@ class tx_dlf_alto implements tx_dlf_fulltext {
      *
      * @access	public
      *
-     * @param	SimpleXMLElement		$xml: The XML to extract the metadata from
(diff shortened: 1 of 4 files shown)


sebastian-meyer:
Scrutinizer checks failed because of invalid composer dependencies. This is fixed in `master` and can be ignored for now.

PR #290, opened 2018-06-18
Add Solr queries to collection plugin

Erikmitk:
This is basically #268 with the requested changes.

> This pull request adds Solr queries to the collection plugin. Since #261 allows to define virtual collections via a Solr query those collections should be able to be shown in the collection plugin.
> 
> To be able to query the Solr an additional field for selecting a Solr core has been added to the plugin configuration.
> 
> I cleaned up some code and replaced several database queries for the corresponding Solr queries to get the necessary data.
> 
> **CollectionList**
> 1. fetch information about selected collections from database
> 2. for each collection
> -- query titles and volumes info from Solr
> -- generate output
> 
> **SingleCollection**
> 1. fetch collection information from DB
> 2. fetch corresponding document UIDs from Solr
> 3. fetch info from those documents from DB
> 4. generate output
> 
> Step 3 could be eliminated. But we would have to make additional changes. This particular query contains the database field metadata_sorting. That field includes all information for sorting the list view if I got that right. I haven’t figured out how the sorting actually works so that might be wrong.
> 
> image
> 
> We could probably fetch the values for sorting the documents from the index, but with the options in the selector we would have to add `author` and `year` to the Solr schema. At the moment they are only dynamic fields.

I made the following changes:

- Added missing case for determining userDefinedCollections
- Fixed process of determining volumes
- Referenced the correct language file (and added a new string)
- Made Solr core selection mandatory for the plugin
- Re-added empty lines for requested formatting

### DIFF
--- a/common/class.tx_dlf_document.php
+++ b/common/class.tx_dlf_document.php
@@ -1048,7 +1048,7 @@ final class tx_dlf_document {
 
                 if (TYPO3_DLOG) {
 
-                    \TYPO3\CMS\Core\Utility\GeneralUtility::devLog('[tx_dlf_document->getRawText('.$id.')] Invalid structure node @ID "'.$id.'"'.self::$extKey, SYSLOG_SEVERITY_WARNING);
+                    \TYPO3\CMS\Core\Utility\GeneralUtility::devLog('[tx_dlf_document->getRawText('.$id.')] Invalid structure node @ID "'.$id.'"', self::$extKey, SYSLOG_SEVERITY_WARNING);
 
                 }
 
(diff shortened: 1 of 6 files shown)


sebastian-meyer (approved):

albig (approved):
The code is working for our usecase.

There remains one problem, we have to investigate: On saving of and tx_dlf_document there seems to be no update hook to update the solr index.

If you make a manuell collection you have to possiblities to add documents:
1. edit the document and add the manual collection
2. use the dynamic feature of the collection record and make so solr query e.g. by uid:12345

In the last case, the amount of title and volume is shown in collection plugin instantly as the solr query takes just the entered query.
In the first case, the solr query is make by the collection name and the result of title/volumes in 0 and the collection is empty.

In both cases, the solr index gets only updated by reindexing the collection. This is not really handy and will lead to confusion.

sebastian-meyer:
I rebased the branch on current master and opened a new pull request. All further work is done in #319.

Issue #295, opened 2018-07-16
Replace SolrPhpClient by Solarium as Solr client


frank-ulrich-weber:
SolrPhpClient is no longer being developed. To stay up to date and to have the possibility to use newer versions of Solr (and PHP) SolrPhpClient should be replaced by [Solarium](https://solarium.readthedocs.io/en/latest/).

(This development is part of an order of the Württembergische Landesbibliothek in cooperation with the Staats- und Universitätsbibliothek Hamburg and the Gottfried Wilhelm Leibniz Bibliothek - Niedersächsische Landesbibliothek.)

frank-ulrich-weber:
The development will initially be based on:
- Solarium 4.0.0 (PHP >= 7)
- Solr 7.4.0 (Java >= 8)

Issue #296, opened 2018-07-16
Refactoring of the listview plugin


frank-ulrich-weber:
The plugin builds up the complete hit list itself without using functions of the search engine. This was required to show the hierarchy of a document. Processing the entire hit list on the other side leads to speed and memory issues, especially in the case of fulltext search. By using the new Solr Client Solarium #295 (in combination with a new Solr version) it should be now possible to use the search engine to build up the hierarchical hit list to improve the speed and memory consumption.

(This development is part of an order of the Württembergische Landesbibliothek in cooperation with the Staats- und Universitätsbibliothek Hamburg and the Gottfried Wilhelm Leibniz Bibliothek - Niedersächsische Landesbibliothek.)

PR #298, opened 2018-08-01
Replaces the old SolrPhpClient with Solarium 4.0

frank-ulrich-weber:
#295


frank-ulrich-weber:
I think the branch solarium_integration is ready for review now. The implementation of Issue #296 that is based on this branch is already available here: https://github.com/frank-ulrich-weber/goobi-presentation/tree/reworkListView. Thanks in advance!

common/class.tx_dlf_list.php
```diff
@@ -244,35 +244,47 @@ protected function getRecord($element) {
...
-
                     // If it is a fulltext search, enable highlighting and fetch the results
                     if ($this->metadata['fulltextSearch']) {
 
-                        $params = array ();
+                        $params['component'] = array (
+                            'highlighting' => array (
+                                'query' => tx_dlf_solr::escapeQuery($this->metadata['searchString']),
+                                'field' => 'fulltext',
+                                'usefastvectrohighlighter' => true
```
sebastian-meyer: This should be `usefastvect*or*highlighter` instead of `usefastvect*ro*highlighter`, right?

composer.json
```diff
@@ -20,7 +20,9 @@
   "type": "typo3-cms-extension",
   "version": "2.2.0",
   "require": {
-    "typo3/cms": ">=6.2.0 <7.9.99"
+    "php" : "^7.0.8",
+    "typo3/cms": "^7.0",
+    "solarium/solarium" : "^4.0"
```
sebastian-meyer, line 24: Is `^` the right operator? I would expect it to be `>= 4.0`.

hooks/class.tx_dlf_tcemain.php
```diff
@@ -140,7 +140,7 @@ public function processDatamap_postProcessFieldArray($status, $table, $id, &$fie
 
                     // Build request for adding new Solr core.
                     // @see http://wiki.apache.org/solr/CoreAdmin
-                    $url = 'http://'.$host.':'.$port.'/'.$path.'admin/cores?wt=xml&action=CREATE&name=dlfCore'.$coreNumber.'&instanceDir=.&dataDir=dlfCore'.$coreNumber;
+                    $url = 'http://'.$host.':'.$port.'/'.$path.'admin/cores?wt=xml&action=CREATE&name=dlfCore'.$coreNumber.'&instanceDir=dlfCore'.$coreNumber.'&dataDir=data&configSet=dlf';
```
sebastian-meyer: Is changing the `instanceDir` and `dataDir` backwards compatible? I.e. is it sufficient to re-index all documents or do you need to re-create all cores as well?
Additionally, we should use `tx_dlf_solr::getSolrConnectionInfo()` here to get the base URL (since `getSolrConnectionInfo` is more generic and no longer core-specific now).

common/class.tx_dlf_solr.php
```diff
@@ -365,23 +370,31 @@ public function search($query = '') {
...
+        // Set query.
+        $this->params['query'] = $query;
+
         // Perform search.
-        $results = $this->service->search((string) $query, 0, $this->limit, $this->params);
+        $selectQuery = $this->service->createSelect($this->params);
+        $results = $this->service->select($selectQuery);
 
-        $this->numberOfHits = count($results->response->docs);
+        $this->numberOfHits = min($results->getNumFound(), $this->limit);
```
sebastian-meyer: `$this->numberOfHits` should be the number of documents found by the query, not only the number of documents retrieved. If the service did find more documents than it is allowed to retrieve (because of a set limit), we should find another way to communicate this to the users. It would be misleading to tell the users, their query found exactly `limit` documents...

common/class.tx_dlf_solr.php
```diff
@@ -270,18 +263,15 @@ public static function getSolrConnectionInfo($core = '') {
...
-
-            $solrInfo['host'] = $conf['solrUser'].':'.$conf['solrPass'].'@'.$solrInfo['host'];
-
-        }
+        // Set username and password.
+        $solrInfo['username'] = $conf['solrUser'];
+        $solrInfo['password'] = $conf['solrPass'];
 
         // Set port if not set.
         $solrInfo['port'] = \TYPO3\CMS\Core\Utility\MathUtility::forceIntegerInRange($conf['solrPort'], 1, 65535, 8180);
```
sebastian-meyer: Newer Solr versions have changed their default port to `8983` (see for example http://lucene.apache.org/solr/guide/7_5/taking-solr-to-production.html). Maybe we should change it here accordingly?

common/class.tx_dlf_solr.php
```diff
@@ -414,7 +422,7 @@ public function search($query = '') {
 
                 $toplevel[$doc->uid]['p'][$doc->id] = array (
                     'u' => $doc->id,
-                    'h' => (!empty($results->highlighting->{$doc->id}->fulltext) ? $results->highlighting->{$doc->id}->fulltext[0] : '')
+                    'h' => ''
```
sebastian-meyer, line 431: Could you please explain why you changed this?

common/class.tx_dlf_solr.php
```diff
@@ -720,30 +738,42 @@ public function __set($var, $value) {
...
-
-        }
-
-        $solrInfo = self::getSolrConnectionInfo($core);
+        $solrInfo = self::getSolrConnectionInfo();
+
+        $config = array (
+            'endpoint' => array (
+                'dlf' => array (
+                    'schema' => 'http',
```
sebastian-meyer: Maybe we should make the schema configurable in `ext_conf_template.txt`, too? We would need to make minor changes to `$this->getSolrConnectionInfo()` and `$this->getSolrUrl()` as well, but would be able to address Solr via secure connections.

common/class.tx_dlf_solr.php
```diff
@@ -720,30 +738,42 @@ public function __set($var, $value) {
...
-            $this->service->setCollapseSingleValueArrays = FALSE;
+        try {
+
+            $this->service->ping($ping);
 
             // Set core name.
             $this->core = $core;
 
             // Instantiation successful!
             $this->ready = TRUE;
```
sebastian-meyer, line 778: `$this->ready` may be set to `TRUE` only if the ping was successful! Does `$this->service->ping()` always throw an exception if the ping fails?

sebastian-meyer (changes requested):
Please re-check the supplied Solr config files:
- You replaced the old solrconfig.xml with a newer version. Did you check the settings in the old file, if there was anything we should customize in the new file as well?
- You added a lot of language specific stopwords, stemming files, etc. Some of them seem to contain only example entries. I suggest removing all files which are not used anyway to keep the extension as clean as possible.
- You removed the `web.xml` hack, which provided the security constraints for the tomcat users. What is the new way to configure HTTP user authentication for Solr? (Or if there is no new way, we should remove the configuration options from the extension.)
- I am not sure if we should include all the Solarium and Symfony files in the extension. Wouldn't it be better to just rely on deploying them via composer?

I only did a formal review, but no actual testing, yet. Please address the questions and issues first.

common/class.tx_dlf_list.php
frank-ulrich-weber: - solrconfig.xml:
I thought it would be a good idea to use the new _default version of solrconfig.xml and adjust it. Yes, I did check the old file and anything should be customized as well.
- OK, I just used the stopwords files of the _default configset to have a "complete" one. It is not the problem to remove the unused files if you like?
- Of course, there is sill an authentication option for Solr! I removed it because this configuration is a part of Solr (Jetty) and not of this extension. I would place a hint to the matching pages of the Solr/Jetty documentation?
- I'm not sure about that point. If anyone does't use the composer the extension does't work. Of cource we can write a documentation on that, but maybe it looks daunting at first sight. Independently of that deployment via composer should already work.

frank-ulrich-weber:
- solrconfig.xml:
I thought it would be a good idea to use the new _default version of solrconfig.xml and adjust it. Yes, I did check the old file and anything should be customized as well.
- OK, I just used the stopwords files of the _default configset to have a "complete" one. It is not the problem to remove the unused files if you like?
- Of course, there is sill an authentication option for Solr! I removed it because this configuration is a part of Solr (Jetty) and not of this extension. I would place a hint to the matching pages of the Solr/Jetty documentation?
- I'm not sure about that point. If anyone does't use the composer the extension does't work. Of cource we can write a documentation on that, but maybe it looks daunting at first sight. Independently of that deployment via composer should already work.

composer.json
frank-ulrich-weber: ^4.0 is the same like >=4.0 <5.0 because I cannot promise the compatibility with solarium 5.0 right now.

hooks/class.tx_dlf_tcemain.php
frank-ulrich-weber: For new installations it should work out of the box. If you upgrade from older Solr installations it depends on Solr what you have to do. Re-creating cores by the extension should not be the goal. Re-index all documents have to be done anyway.

common/class.tx_dlf_list.php
frank-ulrich-weber: Thanks, of course it is a typo!

common/class.tx_dlf_solr.php
frank-ulrich-weber: OK, but count($results->response->docs) counts the number of documents retrieved and not the number of document found so far?

common/class.tx_dlf_solr.php
frank-ulrich-weber: I change it.

common/class.tx_dlf_solr.php
frank-ulrich-weber: OK, I check that.

common/class.tx_dlf_solr.php
frank-ulrich-weber: Yes, if i understood the documentation correctly. 
https://solarium.readthedocs.io/en/stable/queries/ping-query/

common/class.tx_dlf_solr.php
frank-ulrich-weber: Got me! That has nothing to do with the replacement of the solr framework. That was a relic from the time when the full text was still retrieved here and not in tx_dlf_list. $results->highlighting does't exist and thus empty() is always true.

sebastian-meyer: Thanks for the explanation!

common/class.tx_dlf_solr.php
sebastian-meyer:
```suggestion
        $this->numberOfHits = $results->getNumFound();
```
Then how about something like this?

composer.json
sebastian-meyer: Ah, didn't know that. Thanks!

albig: The dependancy to TYPO3 should be at minimum 7.6. All preceding versions von 7.x are not maintained at all and have an unknown number of missing features.

sebastian-meyer:
> I thought it would be a good idea to use the new _default version of solrconfig.xml and adjust it. Yes, I did check the old file and anything should be customized as well.

Yes, that's fine! Thank you!

> I thought it would be a good idea to use the new _default version of solrconfig.xml and adjust it. Yes, I did check the old file and anything should be customized as well.

I think, it would be sufficient to have the files for English and German as an example. Most of the other language files contain only example data anyway.

> Of course, there is sill an authentication option for Solr! I removed it because this configuration is a part of Solr (Jetty) and not of this extension. I would place a hint to the matching pages of the Solr/Jetty documentation?

Yes, please! This has to be documented in `Documentation/Configuration/Index.rst`

> I'm not sure about that point. If anyone does't use the composer the extension does't work. Of cource we can write a documentation on that, but maybe it looks daunting at first sight. Independently of that deployment via composer should already work.

If another extensions use Solarium as well, there will be problems with namespacing and class loading. TYPO3 8 uses composer for deployment and dependency management anyways. That's why we shouldn't include Solarium.

frank-ulrich-weber:
I overworked the code under consideration of the comments.

hooks/class.tx_dlf_tcemain.php
```diff
@@ -126,7 +126,7 @@ public function processDatamap_postProcessFieldArray($status, $table, $id, &$fie
                     }
 
                     // Set port if not set.
-                    $port = (intval($conf['solrPort']) > 0 ? intval($conf['solrPort']) : 8180);
+                    $port = (intval($conf['solrPort']) > 0 ? intval($conf['solrPort']) : 8983);
```
sebastian-meyer: We should use `tx_dlf_solr::getSolrConnectionInfo()` here instead of building the base URL from configuration again.

hooks/class.tx_dlf_em.php
```diff
@@ -59,13 +59,14 @@ public function checkSolrConnection(&$params, &$pObj) {
         }
 
         // Set port if not set.
-        $port = (!empty($this->conf['solrPort']) ? \TYPO3\CMS\Core\Utility\MathUtility::forceIntegerInRange($this->conf['solrPort'], 0, 65535, 8180) : 8180);
+        $port = (!empty($this->conf['solrPort']) ? \TYPO3\CMS\Core\Utility\MathUtility::forceIntegerInRange($this->conf['solrPort'], 0, 65535, 8983) : 8180);
```
sebastian-meyer: We should use `tx_dlf_solr::getSolrConnectionInfo()` here instead of building the base URL from configuration again.
(I am not sure if this works in the context of the extension manager, though.)

composer.json
```diff
@@ -69,6 +71,13 @@
       "plugins/toolbox/tools/imagemanipulation/class.tx_dlf_toolsImagemanipulation.php",
       "plugins/toolbox/tools/imagedownload/class.tx_dlf_toolsImagedownload.php",
       "plugins/validator/class.tx_dlf_validator.php"
+    ],
+    "psr-4" : {
+      "Solarium\\" : "lib/solarium/src",
```
sebastian-meyer: If we don't ship Solarium with the extension, but inject it as a dependency via composer, how does this work then? Shouldn't this be something like `vendor/solarium/...`? Or, alternatively, shouldn't there be an option like `"installer-paths" : { "lib/solarium/src" : ["solarium/solarium"] }`?

sebastian-meyer: Pinging @albig: What would be the prefered way to do it?

sebastian-meyer (changes requested):
Code-wise this looks way better now. There are just a few more comments below.

What's still missing is some form of documentation for upgrading. Or even better: An update script which checks for existing Solr cores in the database and automatically adds them to the "new" Solr instance, so that I can re-index all documents right away.

common/class.tx_dlf_solr.php
sebastian-meyer: Thanks a lot!

common/class.tx_dlf_solr.php
sebastian-meyer: Thank you!

composer.json
frank-ulrich-weber: I had simply overlooked this code within the remove commit. If we rely on composer, everything should work fine without the psr-4 entries here.

hooks/class.tx_dlf_tcemain.php
frank-ulrich-weber: Should we build the base URL within the tx_dlf_solr::getSolrConnectionInfo() method or within a seperate method similar to tx_dlf_solr::getSolrUrl()?

sebastian-meyer: I just want to avoid duplicate code, so instead of reading the extension's configuration and setting the default values in `tx_dlf_tcemain->processDatamap_postProcessFieldArray()` we should use the existing method `tx_dlf_solr::getSolrConnectionInfo()`.
Building the URL from the connection info here is fine. No need to change `getSolrConnectionInfo()`!

composer.json
frank-ulrich-weber: Fixed

hooks/class.tx_dlf_tcemain.php
frank-ulrich-weber: Done

frank-ulrich-weber:
I added an update method to class.ext_update.php.

sebastian-meyer (approved):

sebastian-meyer:
I have no more complaints. We will perform a final test before merging.

class.ext_update.php
```diff
@@ -280,4 +290,135 @@ protected function updateMetadataConfig() {
...
+                $context = stream_context_create(array (
+                    'http' => array (
+                        'method' => 'GET',
+                        'user_agent' => ($conf['useragent'] ? $conf['useragent'] : ini_get('user_agent'))
+                    )
+                ));
+
+                // Build request for adding new Solr core.
+                // @see http://wiki.apache.org/solr/CoreAdmin
+                $url = $solrInfo['scheme'].'://'.$host.':'.$solrInfo['port'].'/'.$solrInfo['path'].'/admin/cores?wt=xml&action=CREATE&name=dlfCore'.$resArray['index_name'].'&instanceDir=dlfCore'.$resArray['index_name'].'&dataDir=data&configSet=dlf';
```
albig: The resulting cores are named "dlfCoredlfCore0". It must be "dlfCore0" instead.

ext_emconf.php
```diff
@@ -16,21 +16,16 @@
...
-    'lockType' => '',
     'version' => '2.2.0',
     'constraints' => array (
         'depends' => array (
-            'php' => '5.3.7-',
-            'typo3' => '6.2.0-7.9.99',
+            'php' => '7.0.8-',
+            'typo3' => '7.6.0-',
         ),
         'conflicts' => array (
```
albig, line 30: If we run in composer mode, the composer.json is checked and the class loading is managed by the autoconf settings.

In non-composer mode, we need this information in ext_emconf.php. Supposed the solarium is installed below "vendor/", the following setting is working for me:

<pre>
'autoload' => array(
   'classmap' => array(
      'vendor',
      'common/class.tx_dlf_solr.php',
      'cli/class.tx_dlf_cli.php',
      'common/class.tx_dlf_alto.php',
      'common/class.tx_dlf_document.php',
      'common/class.tx_dlf_format.php',
      'common/class.tx_dlf_fulltext.php',
      'common/class.tx_dlf_helper.php',
      'common/class.tx_dlf_indexing.php',
      'common/class.tx_dlf_list.php',
      'common/class.tx_dlf_mods.php',
      'common/class.tx_dlf_module.php',
      'common/class.tx_dlf_plugin.php',
      'common/class.tx_dlf_solr.php',
      'common/class.tx_dlf_teihdr.php',
      'hooks/class.tx_dlf_doctype.php',
      'hooks/class.tx_dlf_em.php',
      'hooks/class.tx_dlf_hacks.php',
      'hooks/class.tx_dlf_tceforms.php',
      'hooks/class.tx_dlf_tcemain.php',
      'modules/indexing/index.php',
      'modules/newclient/index.php',
      'plugins/audioplayer/class.tx_dlf_audioplayer.php',
      'plugins/basket/class.tx_dlf_basket.php',
      'plugins/collection/class.tx_dlf_collection.php',
      'plugins/feeds/class.tx_dlf_feeds.php',
      'plugins/listview/class.tx_dlf_listview.php',
      'plugins/metadata/class.tx_dlf_metadata.php',
      'plugins/navigation/class.tx_dlf_navigation.php',
      'plugins/newspaper/class.tx_dlf_newspaper.php',
      'plugins/oai/class.tx_dlf_oai.php',
      'plugins/pagegrid/class.tx_dlf_pagegrid.php',
      'plugins/pageview/class.tx_dlf_pageview.php',
      'plugins/search/class.tx_dlf_search.php',
      'plugins/search/class.tx_dlf_search_suggest.php',
      'plugins/statistics/class.tx_dlf_statistics.php',
      'plugins/toc/class.tx_dlf_toc.php',
      'plugins/toolbox/class.tx_dlf_toolbox.php',
      'plugins/toolbox/tools/pdf/class.tx_dlf_toolsPdf.php',
      'plugins/toolbox/tools/fulltext/class.tx_dlf_toolsFulltext.php',
      'plugins/toolbox/tools/imagemanipulation/class.tx_dlf_toolsImagemanipulation.php',
      'plugins/toolbox/tools/imagedownload/class.tx_dlf_toolsImagedownload.php',
      'plugins/validator/class.tx_dlf_validator.php'
    )
),
</pre>

composer.json
```diff
@@ -20,7 +20,9 @@
   "type": "typo3-cms-extension",
   "version": "2.2.0",
   "require": {
-    "typo3/cms": ">=6.2.0 <7.9.99"
+    "php" : "^7.0.8",
+    "typo3/cms": "^7.6",
```
albig, line 23: In non-composer mode, I've just done a "composer update" inside the extension directory. The result is the download of a full typo3/cms in typo3conf/ext/dlf/vendor/ what is not nice at all. I removed the dependancy to typo3/cms which only loads solarium.

Actually, I have no good-practice, how to act with composer requirements in non-composer setups. :-(

class.ext_update.php
frank-ulrich-weber: Fixed

ext_emconf.php
frank-ulrich-weber: Ok, than I add a vendor folder and some infos about manual installation to the docs?

composer.json
frank-ulrich-weber: I'm not sure what I should do here now? Remove the typo3/cms requirement? Just leave it as it is?

albig: I think, there is no perfect solution to support both - composer mode and non-composer mode.

I would suggest to keep the requirement for typo3/cms, as it is the usual way, extension do. In composer mode, this is reasonable. I would drop requirement of PHP as this is declared by TYPO3 (7.6 requires PHP 7.0).

Finally I think this is a good choice:
<pre>
"require": {
    "typo3/cms": "^7.6",
    "solarium/solarium" : "^4.0"
  },
</pre>

In this case, we should add a note to the manual. In case of non-composer mode, the admin has to go to typo3conf/ext/dlf and run one "composer update". This will create the vendor directory and load solarium and unfortunately a bunch of things, we don't need here.

In ext_emconf.php we should not include the complete vendor-directory, as it contains TYPO3 itself. This could be confusing. The directory for the autoload should be "vendor/solarium".

ext_emconf.php
albig: As noted below, not the whole vendor directory should be added, but 'vendor/solarium'.
Thank you!

composer.json
sebastian-meyer: Do we need to add `vendor/solarium/` to the autoload section of composer.json as well?

And should we use `exclude-from-classmap` for `typo3/cms` just to be sure that there are no conflicts? (see https://getcomposer.org/doc/04-schema.md#exclude-files-from-classmaps)

sebastian-meyer: Please see my changes below: I added `vendor/` to the autoload classmap, but also excluded `vendor/typo3/` from autoloading.

frank-ulrich-weber: As far as I know TYPO3 7.6 requires at least PHP 5.5? Independently of that we can remove the PHP requirement, add vendor/solarium and vendor/symfony/event-dispatcher to autoload section of composer.json. Additionally I add a note to the manual for either run "composer update" once within the dlf directory or a manual download of the two dependencies without using composer. OK?

frank-ulrich-weber: Does it make sense to add empty directories vondor/solarium and vendor/symfony/event-dispatcher for an easier installation without composer?

composer.json
```diff
@@ -20,14 +20,16 @@
...
+    "solarium/solarium" : "^4.0"
   },
   "replace": {
     "dlf": "self.version",
     "typo3-ter/dlf": "self.version"
   },
   "autoload": {
     "classmap": [
+      "vendor/",
```
albig: This actually breaks the composer mode :-(
The vendor directory in typo3conf/ext/dlf/vendor will be only present in non-composer mode. In composer mode, the required libs will be installed in the project wide vendor-directory.

sebastian-meyer: OK, since it seems to be impossible to be compatible to composer mode and non-composer mode at the same time, I suggest we support composer mode as default. TYPO3 uses composer mode as preferred way of installation as well.

So, for short: Adjust composer.json and tx_emconf.php as needed for composer mode and add some documentation for people who are running in non-composer mode.

sebastian-meyer: According to [TYPO3 Core API documentation](https://docs.typo3.org/typo3cms/CoreApiReference/8.7/ApiOverview/Autoloading/) since TYPO3 7 `composer.json` is only evaluated in composer mode. In non-composer mode TYPO3 scans the whole extension directory for class files and builds its own classmap for autoloading. So we should be safe when we provide a `composer.json` for composer mode (i.e. without `vendor` directory).

sebastian-meyer: The extension's documentation should mention, that in non-composer mode you have to do a `composer update` first (to load all dependencies to `vendor/`) and (re-)install the extension via extension manager afterwards. That way all class files will be present to be scanned by the TYPO3 autoloader.

frank-ulrich-weber: I'm a little bit confused now... after a "composer update" we have a lot of additional things within the extension/vendor directory including the already mentioned typo3_src. Isn't that a problem? My suggestion is to create a valid composer.json without vendor/ and "exclude-from-classmap": ["vendor/typo3/"]". The non-composer mode for me is just don't use the command "composer" in any case. The documentation then mentions that you have to download solarium and event-dispatcher manually and place it within a subdirctory (e.G. vendor) of the extension and re-install it.

albig: +1 for the valid composer.json.
-1 for manual hacks ;-)

For non-composer mode, we could ship the sources of solarium and what is necessary in vendor/ OR we explain pleople to execute the `composer update` command in `typo3conf/ext/dlf/`. In the last case, also `vendor/typo3` will be downloaded but this doesn't hurt as long as it is excluded from autoloading. To exclude this, we can explicit mention the classes, paths in `ext_emconf.php` which is only used in non-composer mode.

frank-ulrich-weber: I just realised that there is a different autoloading behavior of T3 7.6 https://docs.typo3.org/typo3cms/CoreApiReference/7.6/ApiOverview/Autoloading/Index.html. If the extension have a composer.json it uses the autoload and psr-4 entries otherwise it scans the extension directory for classes. That was the reasen why I added a psr-4 autoloading configuration to composer.json.

common/class.tx_dlf_solr.php
```diff
@@ -720,30 +741,42 @@ public function __set($var, $value) {
...
+            'endpoint' => array (
+                'dlf' => array (
+                    'scheme' => $solrInfo['scheme'],
+                    'host' => $solrInfo['host'],
+                    'port' => $solrInfo['port'],
+                    'path' => '/'.$solrInfo['path'].'/',
+                    'core' => $core,
+                    'username' => $solrInfo['username'],
+                    'password' => $solrInfo['password']
+                )
```
albig, line 760: We need here another setting "timeout". The default "timeout" is set to 5 seconds by Solarium (see `Core/Client/Endpoint.php`). Most METS files referencing FULLTEXTS cannot be parsed in my setup within 5 seconds. 
Error message is: "Solr HTTP error: HTTP request failed, Operation timed out after 5001 milliseconds with 0 bytes "received
I played around and found 10s ok. But this should be configurable, I think.
It's quite easy, to pass the timeout parameter to the Solarium:

<pre>
$config = array (
            'endpoint' => array (
                'dlf' => array (
                    'scheme' => $solrInfo['scheme'],
                    'host' => $solrInfo['host'],
                    'port' => $solrInfo['port'],
                    'path' => '/'.$solrInfo['path'].'/',
                    'core' => $core,
                    'username' => $solrInfo['username'],
                    'password' => $solrInfo['password'],
                     'timeout' => 10
                )
            )
        );

</pre>

My example: http://digital.slub-dresden.de/data/goobi/490223001-1916092402/DresneNa_490223001-1916092402_02-a_mets.xml

albig (changes requested):

sebastian-meyer:
OK, I changed the `composer.json` to work in composer mode.
I didn't add anything to `ext_emconf.php` since the autoload declaration isn't supported prior to TYPO3 8 which we don't support, yet. (See [here](https://docs.typo3.org/typo3cms/CoreApiReference/7.6/ExtensionArchitecture/DeclarationFile/) and [here](https://docs.typo3.org/typo3cms/CoreApiReference/8.7/ExtensionArchitecture/DeclarationFile/).)

I guess, we need to add to the documentation, that you have to put all dependencies into the extension's directory (either by `composer update` or by manually copying them) before installing the extension in the extension manager, if you don't run in composer mode. However, we should strongly recommend using composer mode!

Sure, that's a PITA if you don't run in composer mode, but I don't know of a way to support both at the same time...

frank-ulrich-weber:
Ok, my last idea to that:
Create an empty folder vendor within the extension and keep autoload of vendor and exclude typo3 within composer.json. Add the same autoload information to ext_emconf.php.

non-composer mode T3 7.6:
composer update (or manual download solarium and event-dispatcher within vendor directory)
T3 uses composer.json for class loading and doesn't care about class loading within ext_emconf.php.

non-composer mode T3 8.7
composer update (or manual download solarium and event-dispatcher within vendor directory)
T3 uses ext_emconf.php for class loading and doesn't care about class loading within composer.json.

composer mode T3 7.6 and T3 8.7
composer.json is used for class loading of the extension classes and doesn't care about ext_emconf.php. The vendor folder within the extension is just an empty folder without any classes.

sebastian-meyer:
I would leave `ext_emconf.php` as it is. Its autoload information is not used by TYPO3 7 and we don't support TYPO3 8, yet. (I am working on a TYPO3 8 compliance branch, but there we will switch to PSR-4 classloading, so I have to change that anyway.)

albig:
Without changing the ext_emconf.php the Solarium classes are not found. The described scan of the "complete extension directory" seems to ignore the vendor directory. Actually, I find no Solarium classes in `typo3temp/autoload/autoload_classmap.php` but the tx_dlf_* classes are present.

The documentation about ext_emconf.php is inclomplete. The feature exists since TYPO3 7.5. See commit https://github.com/TYPO3/TYPO3.CMS/commit/9a786007b584f46e8a2e1b47a27d053a44a6022d and the [TYPO3 changelog Feature-68700](https://docs.typo3.org/typo3cms/extensions/core/latest/Changelog/7.5/Feature-68700-AutoloadDefinitionCanBeProvidedInExt_emconfphp.html).

With the modified ext_emconf.php I have tx_dlf_* classes and Solarium classes in `typo3temp/autoload/autoload_classmap.php`.

sebastian-meyer:
So we would be fine if we prepare `composer.json` for composer mode and `ext_emconf.php` for non-composer mode?

sebastian-meyer:
Fixes #295

PR #312, opened 2018-11-19
Implementation of enhancement #296 "Refactoring of the listview plugin"

frank-ulrich-weber:


common/class.tx_dlf_list.php
```diff
@@ -267,12 +268,30 @@ protected function getRecord($element) {
 
                     // Set additional query parameters.
                     $params['start'] = 0;
-                    $params['rows'] = $this->solr->limit;
 
-                    // Set query.
-                    $params['query'] = 'uid:'.tx_dlf_solr::escapeQuery($record['uid']);
+                    // All results are needed, so set the limit unreachable high.
+                    $params['rows'] = 1000000000;
```
sebastian-meyer: What's the point in having a hard limit if we don't enforce it here?
The default limit should be high enough to avoid running into problems, but it should still be reasonably low to prevent stalling the system. So instead of removing the limit here we should increase the default value if it's necessary.

common/class.tx_dlf_list.php
```diff
@@ -267,12 +268,30 @@ protected function getRecord($element) {
...
 
-                    // Perform search.
+                    // Take over existing filter queries.
+                    $params['filterquery'] = isset($this->metadata['options']['params']['filterquery']) ? $this->metadata['options']['params']['filterquery'] : array ();
+
+                    // Extend filter query to get all documents with the same uid.
+                    foreach ($params['filterquery'] as $key=>$value) {
+
+                        $params['filterquery'][$key] = array ('query' => $value['query'].' OR toplevel:true');
```
sebastian-meyer: Shouldn't this be something like
```suggestion
                        $params['filterquery'][$key]['query'] = $value['query'].' OR toplevel:true';
```
to avoid deleting other array keys besides 'query'?

frank-ulrich-weber:
Is there a chance for a timely review of this pull request? Thanks!

sebastian-meyer:
I already started a review two weeks ago and I'm almost done with it. As you can see I also made some changes.
The last concern I am currently investigating is that you moved a lot of search functionality from tx_dlf_solr to tx_dlf_listview. I am not yet sure, if that is a good idea.

frank-ulrich-weber:
Thank you for your fast reply. I will wait for the result of your review.

sebastian-meyer (changes requested):

common/class.tx_dlf_list.php
frank-ulrich-weber: Ok, then I think it would be a good idea to configure the limit within the Solr configuration of the extension. What do you think?

common/class.tx_dlf_list.php
frank-ulrich-weber: I think we can savely change this. At the moment we don't use any other array keys besides query, but it increases the flexibility for future changes. I think we should change a second occurrence, too.  [tx_dlf_solr](https://github.com/frank-ulrich-weber/goobi-presentation/blob/reworkListView/common/class.tx_dlf_solr.php#L377)

sebastian-meyer (approved):

PR #320, opened 2018-12-12
Fix typo

lutzhelm:


sebastian-meyer:
Nice catch!

At least this was consistently wrong. :smirk:

PR #324, opened 2019-01-03
 Adds the possibility to change the sorting (count/index) of the facets 

frank-ulrich-weber:


plugins/search/locallang.xml
```diff
@@ -68,6 +71,9 @@
 			<label index="tt_content.pi_flexform.facets">Diese Facetten anzeigen</label>
 			<label index="tt_content.pi_flexform.limitFacets">Maximale Facettenwerte</label>
 			<label index="tt_content.pi_flexform.resetFacets">Facetten zurücksetzen erlauben</label>
+			<label index="tt_content.pi_flexform.sortingFacets">Facetten Sortierung</label>
+			<label index="tt_content.pi_flexform.sortingFacets.count">Count</label>
+			<label index="tt_content.pi_flexform.sortingFacets.index">Index</label>
```
sebastian-meyer: `Count` and `Index` should be replaced by German translations. I don't even know, what sorting by `index` means. :)

sebastian-meyer (changes requested):

plugins/search/locallang.xml
frank-ulrich-weber: Good point :)

sebastian-meyer (approved):

Issue #304, opened 2018-09-17
Page View Plugin doesn't work out-of-the-box


sebastian-meyer:
Since the third-party extension `t3query` is no longer compatible with newer versions of TYPO3, we removed the dependency, but failed to find a proper alternative to ensure loading of the jQuery library prior to initializing OpenLayers in the Page View Plugin.
Recent versions of TYPO3 come with their own jQuery library, so we should rely on that. We need to implement it in a way that the Page View Plugin ensures loading of the jQuery library.

albig:
That's not true. If you include the static template - as you should do, jQuery is included via CDN or from dlf. Current Javascript in Kitodo.Presentation is only working with jQuery 2.x.
The TYPO3 jQuery library is IMHO intended for backend modules only.

sebastian-meyer:
Thanks for the quick reply! The static template was added, but the Page View Plugin didn't show anything. The Chrome developer console showed Javascript errors hinting at a missing jQuery library.
I'll have another look at it!

sebastian-meyer:
I can't reproduce this issue. Even with a blank minimal installation of TYPO3 8.7 the plugin works out-of-the-box.

Issue #163, opened 2017-03-25
Add composer.json and class loading information for TYPO3 7


albig:
In TYPO3 7 the class loading changed completely. The file `ext_autoload.php` is ignored but a section "autoload" in `ext_emconf.php` could be used to provide class locations and help the autoloader.

Even better is to provide a `composer.json` file with autoload information. This makes it possible to use Kitodo.Presentation in composer mode. Running TYPO3 in composer mode makes it easier for deployment and class loading should be even faster.

Read the blogpost of Helmut Hummel: [Changes in class loading in TYPO3 7LTS](http://insight.helhum.io/post/130812561790/changes-in-class-loading-in-typo3-7lts)
and  [COMPOSER AND TYPO3 7LTS](https://usetypo3.com/typo3-and-composer.html).

This issue is related to #156.

sebastian-meyer:
I added a `composer.json` with class autoloading information. But I haven't removed `ext_autoload.php` or any information in `ext_emconf.php` to still keep compatibility with TYPO3 6.2.
Before releasing v3.0.0 and requiring at least TYPO3 7.6 we should fix that, too.

sebastian-meyer:
This should be fixed in compatibility branch [kitodo-presentation/typo3-8+](/kitodo/kitodo-presentation/tree/typo3-8%2B)

sebastian-meyer:
Fixed with #329

PR #329, opened 2019-03-16
Basic compatibility with TYPO3 8 and Composer

sebastian-meyer:


sebastian-meyer:
Fixes #304 
Fixes #163

Issue #311, opened 2018-11-16
Image download tool in Toolbox plugin offers IIIF image base URI in default configuration


lutzhelm:
The default `fileGrpsImageDownload` in plugins/toolbox/flexform.xml is the same as the default values for `fileGrps`: `MIN,DEFAULT,MAX`. If a METS/MODS file contains IIIF images in those `mets:fileGrp`s, the image download plugin will simply link the given IIIF image base URI - which results in an error or in a redirect to the image info.json.

It's easy to generate an actual image link to `full/full/0/native.jpg` or `full/full/0/defaut.jpg`, but for the necessary distinction between IIIF image API 1 and 2, an additional server side request for the info.json would be needed - besides the already required client side request.

I suggest to resolve this issue for IIIF image with the necessary changes to view IIIF manifests as those changes will introduce functionality to access IIIF resources more easily - and I have to solve the issue for the display of IIIF manifests anyway.

I guess the download issue also occurs for IIP and Zoomify images.

sebastian-meyer:
Instead of requesting the `info.json` from the server, we could just check the given `MIMETYPE` in the METS file to figure out if IIIF is used (see https://github.com/kitodo/kitodo-presentation/blob/master/common/class.tx_dlf_document.php#L396).

lutzhelm:
We need the `info.json` because the full image URL is different for different IIIF versions, and the custom `MIMETYPE` does not tell us anything about it. For version 1.x,  it's `full/full/0/native.jpg`, for version 2.x, it's `full/full/0/default.jpg`, and unfortunately, for the upcoming version 3 it will change again: `full/max/0/default.jpg`.

lutzhelm:
Concerning IIP and Zoomify:

- Full images for IIP can be retrieved by using the `CVT=jpeg` command without any region / size restrictions: http://iipimage.sourceforge.net/documentation/protocol/
- AFAIK full images for Zoomify are not available through any documented requests / parameters. It would be possible to retrieve the tile images at the best resolution on the client side, draw them on a hidden canvas and return the image via data URL, but that is a hack with results of questionable quality.

sebastian-meyer:
Does #336 completely fix this issue?

lutzhelm:
As far as the original issue and IIIF are concerned, yes, and it also contains a fix for IIP.

I don't know if there is a solution for Zoomify beyond disabling the image download for Zoomify all together.

PR #336, opened 2019-04-08
CMR 3.2 - Implement IIIF Presentation API

lutzhelm:
This pull requests adds support for the IIIF Presentation API. Primarily, version 2 is supported, but manifests of the IIIF Metadata API 1 and the IIIF Presentation API 3 can also be displayed.

Major changes:
* `\Kitodo\Dlf\Common\Document` is now abstract, with `\Kitodo\Dlf\Common\MetsDocument` and `\Kitodo\Dlf\Common\IiifManifest` as subclasses.
* `tx_dlf_document.document_format` has been added to distinguish between METS and IIIF documents in the database
* IIIF Metadata can be configured by providing JSONPath expressions. This can be don in the same way as for XML formats, but there is no IIIF implementation of the `\Kitodo\Dlf\CommonMetadataInterface` as IIIF manifests in Kitodo.Presentation currently only support IIIF metadata.
* To reflect the broader usage of the corresponding database table `tx_dlf_metadataformat` , `tx_dlf_metadataformat.xpath` and `tx_dlf_metadataformat.xpath_sorting` have been renamed to `metadataquery` and `metadataquery_sorting`.
* IIIF manifest evaluation and version specific code is hidden behind Interfaces in https://github.com/ubleipzig/php-iiif-prezi-reader which has been added as a composer dependency.
* An annotation tool has been added to the Toolbox plugin to display IIIF annotations.

The OAI-PMH, AudioPlayer and Newspaper plugin are not supported via IIIF PresentationAPI yet.

Existing functionality should not be impaired - except if the updates in `class.ext_update.php` do not work properly, which is why I'd welcome an especially observant eye for a review of that class.


lutzhelm:
This PR also fixes #311 for IIIF and IIP.

Classes/Common/Document.php
```diff
@@ -20,16 +23,7 @@
  * @subpackage dlf
  * @access public
  */
```
sebastian-meyer, line 26: Please add `@abstract` keyword to phpDoc.

Classes/Common/IiifManifest.php
```diff
@@ -0,0 +1,884 @@
...
+                }
+            }
+            if (isset($fileUseFulltext)) {
+                $iiifAlto = $this->iiif->getSeeAlsoUrlsForFormat("application/alto+xml");
+                if (empty($iiifAlto)) {
+                    $iiifAlto = $this->iiif->getSeeAlsoUrlsForProfile("http://www.loc.gov/standards/alto/", true);
+                }
+                if (!empty($iiifAlto)) {
+                    // FIXME use multiple possible alto files?
+                    $this->mimeTypes[$alto[0]] = "application/alto+xml";
```
sebastian-meyer: There is no array `$alto` at this point. Is this meant to be `$iiifAlto`?

Classes/Common/DocumentTypeCheck.php
```diff
@@ -85,6 +85,7 @@ public function getDocType() {
          */
         switch ($toc[0]['type']) {
             case 'newspaper':
+                // TODO Delegate to METS document method. Newspapers currently not supported for IIIF in Kitodo.Presentation.
```
sebastian-meyer: Is there anything like a document type in IIIF? So, does the type checker work with IIIF manifests and just doesn't support newspapers, yet?
Since the type check is used in Typoscript, it is of critical importance that it doesn't return unsupported values. Otherwise it would call plugins like the calendar for documents which are not supported.

If IIIF manifests don't provide a document type at all, this should be checked around line 67 and no type (e.g. an empty string) should be returned. Otherwise at least unsupported types like newspapers have to be catched and handled accordingly.

Classes/Common/Document.php
```diff
@@ -778,64 +545,68 @@ public function getPhysicalPage($logicalPage) {
...
+     * XML fulltext representation (ALTO).
      *
-     * @return string The physical structure node's raw text
+     * @param string $id: The @ID attribute of the physical structure node (METS) or the @id property
+     * of the Manifest / Range (IIIF)
+     *
+     * @return string The physical structure node's / IIIF resource's raw text from XML
      */
-    public function getRawText($id) {
+    protected function getRawTextFromXml($id) {
```
sebastian-meyer, line 594: What's the difference between `getRawText()` and `getRawTextFromXml()`? Is there any other format we want to retrieve text from?

Classes/Hooks/ConfigurationForm.php
```diff
@@ -166,6 +169,36 @@ public function checkMetadataFormats(&$params, &$pObj) {
...
             ];
         }
+        // Add IIIF Metadata API 1 context
+        if (!$nsDefined['IIIF1']) {
+            $data['tx_dlf_formats'][uniqid('NEW')] = array (
+                'pid' => 0,
+                'type' => 'IIIF1',
+                'root' => 'none',
+                'namespace' => 'http://www.shared-canvas.org/ns/context.json',
+                'class' => 'none'
```
sebastian-meyer: I'd suggest leaving `class` empty just in case anybody defines a class named `none` which would cause problems.

Classes/Hooks/ConfigurationForm.php
```diff
@@ -166,6 +169,36 @@ public function checkMetadataFormats(&$params, &$pObj) {
...
+            );
+        }
+        // Add IIIF Presentation 2 context
+        if (!$nsDefined['IIIF2']) {
+            $data['tx_dlf_formats'][uniqid('NEW')] = array (
+                'pid' => 0,
+                'type' => 'IIIF2',
+                'root' => 'none',
+                'namespace' => 'http://iiif.io/api/presentation/2/context.json',
+                'class' => 'none'
```
sebastian-meyer: I'd suggest leaving class empty just in case anybody defines a class named none which would cause problems.

Classes/Hooks/ConfigurationForm.php
```diff
@@ -166,6 +169,36 @@ public function checkMetadataFormats(&$params, &$pObj) {
...
+            );
+        }
+        // Add IIIF Presentation 3 context
+        if (!$nsDefined['IIIF3']) {
+            $data['tx_dlf_formats'][uniqid('NEW')] = array (
+                'pid' => 0,
+                'type' => 'IIIF3',
+                'root' => 'none',
+                'namespace' => 'http://iiif.io/api/presentation/3/context.json',
+                'class' => 'none'
```
sebastian-meyer: I'd suggest leaving class empty just in case anybody defines a class named none which would cause problems.

Classes/Hooks/ConfigurationForm.php
```diff
@@ -166,6 +169,36 @@ public function checkMetadataFormats(&$params, &$pObj) {
                 'class' => 'Kitodo\\\\Dlf\\\\Format\\\\Alto'
             ];
         }
+        // Add IIIF Metadata API 1 context
+        if (!$nsDefined['IIIF1']) {
+            $data['tx_dlf_formats'][uniqid('NEW')] = array (
+                'pid' => 0,
+                'type' => 'IIIF1',
+                'root' => 'none',
```
sebastian-meyer: This won't work, since the TCA configuration evaluates this field as `unique`. You have to either change the TCA configuration as well (which I won't recommend, since having multiple root element names registered would lead to other problems) or make up unique "root" entries here. I'd suggest simply repeating `type` here.

Classes/Hooks/ConfigurationForm.php
```diff
@@ -166,6 +169,36 @@ public function checkMetadataFormats(&$params, &$pObj) {
...
+                'namespace' => 'http://www.shared-canvas.org/ns/context.json',
+                'class' => 'none'
+            );
+        }
+        // Add IIIF Presentation 2 context
+        if (!$nsDefined['IIIF2']) {
+            $data['tx_dlf_formats'][uniqid('NEW')] = array (
+                'pid' => 0,
+                'type' => 'IIIF2',
+                'root' => 'none',
```
sebastian-meyer: This won't work, since the TCA configuration evaluates this field as `unique`. You have to either change the TCA configuration as well (which I won't recommend, since having multiple root element names registered would lead to other problems) or make up unique "root" entries here. I'd suggest simply repeating `type` here.

Classes/Hooks/ConfigurationForm.php
```diff
@@ -166,6 +169,36 @@ public function checkMetadataFormats(&$params, &$pObj) {
...
+                'namespace' => 'http://iiif.io/api/presentation/2/context.json',
+                'class' => 'none'
+            );
+        }
+        // Add IIIF Presentation 3 context
+        if (!$nsDefined['IIIF3']) {
+            $data['tx_dlf_formats'][uniqid('NEW')] = array (
+                'pid' => 0,
+                'type' => 'IIIF3',
+                'root' => 'none',
```
sebastian-meyer: This won't work, since the TCA configuration evaluates this field as `unique`. You have to either change the TCA configuration as well (which I won't recommend, since having multiple root element names registered would lead to other problems) or make up unique "root" entries here. I'd suggest simply repeating `type` here.

Classes/Common/Indexer.php
```diff
@@ -445,29 +448,44 @@ protected static function processPhysical(Document &$doc, $page, array $physical
...
+                        return 1;
+                    }
+                } else {
                     return 1;
                 }
-            } else {
-                return 1;
+            }
+            if (!empty($physicalUnit['annotationContainers'])) {
+                $annotationContainerIds = $physicalUnit['annotationContainers'];
```
sebastian-meyer: You already declared `$annotationContainerIds` with `$physicalUnit['annotationContainers']` at line 452. :)

Classes/Common/Indexer.php
```diff
@@ -445,29 +448,44 @@ protected static function processPhysical(Document &$doc, $page, array $physical
...
+                foreach ($annotationContainerIds as $annotationContainerId) {
+                    if (!empty($extConf['useragent'])) {
+                        @ini_set('user_agent', $extConf['useragent']);
+                    }
+                    $annotationContainer = IiifHelper::loadIiifResource(\TYPO3\CMS\Core\Utility\GeneralUtility::getUrl($annotationContainerId));
```
sebastian-meyer, line 483: Here you get the IIIF resource for the annotation container, but you don't use it anywhere. Either there is code missing or we don't need this here.

Classes/Plugin/PageView.php
```diff
@@ -219,6 +232,59 @@ protected function getFulltext($page) {
...
+                /* @var $canvas \Ubl\Iiif\Presentation\Common\Model\Resources\CanvasInterface */
+                if ($canvas!=null && !empty($canvas->getPossibleTextAnnotationContainers(Motivation::PAINTING))) {
+                    $annotationContainers = array();
+                    /*
+                     *  TODO Analyzing the annotations on the server side requires loading the annotation lists / pages
+                     *  just to determine wether they contain text annotations for painting. This will take time and lead to a bad user experience.
+                     *  It would be better to link every annotation and analyze the data on the client side.
+                     *
+                     *  On the other hand, server connections are potentially better than client connections. Downloading annotation lists
+                     */
```
sebastian-meyer, line 261: Are we talking about OCR fulltext or about "real" annotations here? I assume the latter.
Maybe we could introduce another `Document` class property `hasAnnotations` (or similar) and move this check there? So we would have to check for annotations only once.

class.ext_update.php
```diff
@@ -121,7 +132,7 @@ protected function oldFormatClasses() {
         $result = $GLOBALS['TYPO3_DB']->exec_SELECTquery(
             'tx_dlf_formats.uid AS uid,tx_dlf_formats.type AS type',
             'tx_dlf_formats',
-            'tx_dlf_formats.class NOT LIKE "%\\\\\\\\%"' // We are looking for a single backslash...
+            'tx_dlf_formats.class IS NOT NULL AND tx_dlf_formats.class != "" AND tx_dlf_formats.class NOT LIKE "%\\\\\\\\%"' // We are looking for a single backslash...
```
sebastian-meyer, line 130: Another reason to not set `class = none` for IIIF in `ConfigurationForm.php`: It would always trigger this update.

sebastian-meyer (changes requested):
Besides my in-line comments there are a few general issues:

- There still are some METS-specific properties and methods in the abstract `Document` class which should be moved to the new `MetsDocument` implementation.
- Some properties and methods across most changed files don't have full phpDoc documentation (at least a short description, `@access`, `@param` and `@return`).
- There are a few `// TODO` comments throughout your code. Please make sure, that there are no critical issues missing. Every part of the extension should either support IIIF alongside METS or at least handle IIIF manifests graciously and without errors.
- I'd suggest not renaming `tx_dlf_metadataformat.xpath` and `tx_dlf_metadataformat.xpath_sorting` even though the field names aren't accurate anymore. Otherwise upgrading the extension could lead to problems (e. g. if you perform a database schema update in the Install Tool the `class.ext_update.php` script will fail to `ALTER TABLE ...` afterwards). We should mitigate this by keeping the field names.

Classes/Common/Document.php
lutzhelm: `getRawText()` is currently only used for indexing full text and can be called for both METS and IIIF. Full text can be given as ALTO for METS or as ALTO (via `seeAlso`) or text annotations for IIIF. If fulltext is linked as ALTO in a IIIF manifest, `getRawTextFromXml()` is used; also text annotations with motivation `painting` are parsed (if the extension is configured to treat them as full text). So `MetsDocument->getRawText()` is nothing more than a wrapper around `Document->getRawTextFromXml()`, while `IiifManifest->getRawText()` has to check both the ALTO and the annotation option.
I'll update the `getRawTextFromXml()` documentation to make that more clear.

Classes/Plugin/PageView.php
lutzhelm: Annotations could be anything, OCR fulltext as well as descriptions, for example. It's common to provide OCR fulltext as annotations with the `motivation` [`painting`](http://iiif.io/api/presentation/2#painting), but it can't be assumed that textual annotations with `motivation` `painting` are always fulltext. That's why I introduced the `AnnotationTool`. It currently displays only plain text annotations with the `motivation` `painting`, but I'd like to extent this to all motivations, with either a configuration option for the Toolbox plugin or the extension, or with a motivation selector within the AnnotationTool, with `painting` as default, if this is desired.
A check for the existence of annotations shouldn't be done for the whole `Document` because annotations are linked via the single `Canvas`. I would suggest to introduce a check in `Ubl\Iiif\Presentation\Common\Model\Resources\Canvas` with a one time lazy loading of the `AnnotationList` / `AnnotationPage`, but leave this code as it is for this PR.

Classes/Common/Document.php
lutzhelm: Solved via 37b6ad0093d6df2bdfbe9177ce9e94b810ba7d20

Classes/Common/IiifManifest.php
lutzhelm: Correct! Solved via 8e32bdf04b8bca5b2fe01d3c10feeef68c029f09.

Classes/Common/DocumentTypeCheck.php
lutzhelm: It's a little complicated. IIIF itself does not promote any property to semantically identify document types, but this metadata can be configured as JSONPath. But currently, only IIIF manifests are supported, not IIIF collections. Newspapers, however, would be provided as collections. For example, the National Library of Wales uses this approach: https://dev.llgc.org.uk/wiki/index.php?title=IIIF_Newspapers
As long as there is no common practice, I'd suggest to handle `newspaper` as unsupported type. This is done via 1f6051fe894f71bfdae5de0a12ab5d14efa2546a.

Classes/Common/Document.php
lutzhelm: Done via 319be3785257e8f409489ea60541aacaf0a3d876.

Classes/Hooks/ConfigurationForm.php
lutzhelm: Solved via 706aeebd38d24bb234d7c36e342d46440e098fe8.

Classes/Hooks/ConfigurationForm.php
lutzhelm: Solved via 706aeebd38d24bb234d7c36e342d46440e098fe8.

Classes/Hooks/ConfigurationForm.php
lutzhelm: Solved via 706aeebd38d24bb234d7c36e342d46440e098fe8.

Classes/Hooks/ConfigurationForm.php
lutzhelm: Solved via 706aeebd38d24bb234d7c36e342d46440e098fe8.

class.ext_update.php
lutzhelm: True. This change however is still relevant as `class` may now be empty.

Classes/Common/Indexer.php
lutzhelm: It's only used a line below to check if the resource is actually an `AnnotationList` / `AnnotationPage`.

Classes/Hooks/ConfigurationForm.php
lutzhelm: Solved via 706aeebd38d24bb234d7c36e342d46440e098fe8.

Classes/Hooks/ConfigurationForm.php
lutzhelm: Solved via 706aeebd38d24bb234d7c36e342d46440e098fe8.

Classes/Common/Indexer.php
lutzhelm: True! Solved via f65161a90a94ced6e5773f01a68cc139a601bc9c.

lutzhelm:
* METS properties and functions have been moved in ae7a76353bf47d6939eca4fb07fccb7162c038ff
* Documentation has been improved in b005b5af4067276a4f945bf63401ef92fbc9f8c8 and 5a1ced21938842ff86c6e1cdc95dd70a06536a14
* I checked and removed some of the TODOs. For the remaining ones I'd rather propose new pull requests. Kitodo.Presentation should actually handle IIIF manifests and errors within them graciously already.
* I've undone the changes for renaming `xpath` and `xpath_sorting`, see f6278323a3e099cac283bf654cabcc25a2935c98.

sebastian-meyer (approved):

sebastian-meyer:
Thank you very much! This looks good!

PR #341, opened 2019-04-15
Improve IIIF tile source - port code from OpenLayers PR

### LINKED ISSUE #340: Port IIIF image tile source code from OpenLayers
I just made a pull request to OpenLayers (openlayers/openlayers#9430) that would introduce a built-in IIIF tile source into OpenLayers. Whether or not this PR is accepted, I'd suggest to port the code into Kitodo.Presentation's IIIF tile source.

This would significantly improve the tile source:
* support for compliance level 0, i.e. pregenerated tiles / full image scales
* support for non square tiles
* image requests in canonical form - which also works better with image servers that do not implement size by percentage correctly
* already rudimentary support for upcoming IIIF Image API version 3

lutzhelm:
Port PR openlayers/openlayers#9430 to Kitodo.Presentation.

This adds the following improvements to the IIIF Image API tile source:
* Uses [canonical image request](https://iiif.io/api/image/2.1/#canonical-uri-syntax). Some image servers (e.g. Loris) have / had problems with size calculation by percentage.
* Adds basic Image API [version 3](https://iiif.io/api/image/3.0/) support.
* Adds [level 0](https://iiif.io/api/image/2.1/compliance/#level-0-compliance) support - the source does now support image services that only provide pregenerated tiles or images.
* Supports region calculation by absolute coordinates and by percentages, size calculation by width, by height, by width and height and by percentage if the image's `info.json` indicates support.
* Supports non-square tiles.

Closes #340.

### DIFF
--- a/Resources/Public/Javascript/PageView/OL3Sources.js
+++ b/Resources/Public/Javascript/PageView/OL3Sources.js
@@ -47,13 +47,15 @@ dlfViewerSource.findKey = function(obj, f, opt_this) {
  * @param {string} url
  */
 dlfViewerSource.tileLoadFunction = function(tileSize, tile, url) {
-    var img = tile.getImage();
+    var img = tile.getImage(),
+        tileWidth = Array.isArray(tileSize) ? tileSize[0] : tileSize,
+        tileHeight = Array.isArray(tileSize) ? tileSize[1] : tileSize;
     $(img).load(function() {
         if (img.naturalWidth > 0 &&
-          (img.naturalWidth != tileSize || img.naturalHeight != tileSize)) {
+          (img.naturalWidth != tileWidth || img.naturalHeight != tileHeight)) {
             var canvas = document.createElement('canvas');
-            canvas.width = tileSize;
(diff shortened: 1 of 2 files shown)


sebastian-meyer:
The pull request openlayers/openlayers#9430 had some further fixes before it was merged 3 days ago.
Does this pull request need those updates as well?

lutzhelm:
I did not port all of my additions to OpenLayers, only

- most the tile source code (which hasn't changed in OpenLayers after the pull request aside from changes to the documentation)
- a function for identifying the IIIF Image API compliance level
- a function for parsing Image API version 3 `info.json`s into OL tile source options
- the set of qualities and formats by API version and compliance level has been updated with the size and region calculation features 

Most of the additional commits concern documentation, examples, minifying problems and information response parsing. Those `info.json` parsing issues do not concern the ported code because they have been solved in a different way in Kitodo.Presentation.

As for version 3, I think it will be necessary to revise the tile source and the image information response parsing once the API version is final.

PR #347, opened 2019-06-14
add filter configuration for collections

christopher-timm:
This PR adds the possibility to configure the collections shown as facet.
If no configuration is selected all collections are displayed like before.

### DIFF
--- a/Classes/Plugin/Search.php
+++ b/Classes/Plugin/Search.php
@@ -187,7 +187,7 @@ class Search extends \Kitodo\Dlf\Common\AbstractPlugin {
             return '';
         }
         // Quit without doing anything if no facets are selected.
-        if (empty($this->conf['facets'])) {
+        if (empty($this->conf['facets']) && empty($this->conf['facetCollections'])) {
             return '';
         }
         // Get facets from plugin configuration.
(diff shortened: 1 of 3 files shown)


Configuration/Flexforms/Search.xml
```diff
@@ -172,6 +172,22 @@
                             </config>
                         </TCEforms>
                     </facets>
+                    <facetCollections>
+                        <TCEforms>
+                            <displayCond>FIELD:pages:REQ:true</displayCond>
+                            <exclude>1</exclude>
+                            <label>Folgende Kollektionen als Facetten anzeigen</label>
```
sebastian-meyer: Please use the localization feature for the label.

Classes/Plugin/Search.php
```diff
@@ -551,6 +551,26 @@ public function makeFacetsMenuArray($content, $conf) {
         $selectQuery = $solr->service->createSelect($search['params']);
         $results = $solr->service->select($selectQuery);
         $facet = $results->getFacetSet();
+
+        $facetCollectionArray = array();
+        foreach (\TYPO3\CMS\Core\Utility\GeneralUtility::trimExplode(',', $this->conf['facetCollections'], TRUE) as $facetCollection) {
```
sebastian-meyer: Instead of performing a single database query for every selected collection, I suggest doing one query instead (paraphrasing):
`SELECT index_name FROM tx_dlf_collections WHERE uid IN ($this-conf['facetCollections'])`

Since `$this->conf['facetCollections']` is a comma-separated list of UIDs, you can use it right away as a selector in the database query. Just do some sanitizing before to make sure that the variable indeed has the assumed format (i.e. contains only integers and commas).

sebastian-meyer (changes requested):

christopher-timm:
Thanks @sebastian-meyer. 
The requested changes are implemented.
As sanitization everything else than numbers and comma are removed.

Classes/Plugin/Search.php
```diff
@@ -551,6 +551,27 @@ public function makeFacetsMenuArray($content, $conf) {
         $selectQuery = $solr->service->createSelect($search['params']);
         $results = $solr->service->select($selectQuery);
         $facet = $results->getFacetSet();
+
+        $facetCollectionArray = array();
+
+        // replace everything expect numbers and comma
+        $facetCollections = preg_replace('/[^0-9|,]/', '', $this->conf['facetCollections']);
```
sebastian-meyer: This regular expression would allow `|` to be present in the string as well. Please remove the `|` to only allow numbers and comma.
```suggestion
        $facetCollections = preg_replace('/[^0-9,]/', '', $this->conf['facetCollections']);
```

sebastian-meyer (changes requested):

sebastian-meyer (approved):
Thank you!

PR #352, opened 2019-06-26
adds span's so that pagebrowser can be formatted like a button bar

BFallert:
With the added spans its possible to format the PageBrowser like GitHub's PageBrowser 
image

### DIFF
--- a/Classes/Plugin/PageGrid.php
+++ b/Classes/Plugin/PageGrid.php
@@ -91,7 +91,7 @@ class PageGrid extends \Kitodo\Dlf\Common\AbstractPlugin {
         if ($this->piVars['pointer'] > 0) {
             $output = $this->pi_linkTP_keepPIvars($this->pi_getLL('prevPage', '&lt;', TRUE), ['pointer' => $this->piVars['pointer'] - 1, 'page' => (($this->piVars['pointer'] - 1) * $this->conf['limit']) + 1], TRUE).$separator;
         } else {
-            $output = '<span class="first-page not-active">'.$this->pi_getLL('prevPage', '&lt;', TRUE).'</span>'.$separator;
+            $output = '<span class="prev-page not-active">'.$this->pi_getLL('prevPage', '&lt;', TRUE).'</span>'.$separator;
         }
         $i = 0;
         // Add links to pages.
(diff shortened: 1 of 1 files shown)


Classes/Plugin/PageGrid.php
```diff
@@ -91,7 +91,7 @@ protected function getPageBrowser() {
         if ($this->piVars['pointer'] > 0) {
             $output = $this->pi_linkTP_keepPIvars($this->pi_getLL('prevPage', '&lt;', TRUE), ['pointer' => $this->piVars['pointer'] - 1, 'page' => (($this->piVars['pointer'] - 1) * $this->conf['limit']) + 1], TRUE).$separator;
         } else {
-            $output = $this->pi_getLL('prevPage', '&lt;', TRUE).$separator;
+            $output = '<span class="first-page not-active">'.$this->pi_getLL('prevPage', '&lt;', TRUE).'</span>'.$separator;
```
sebastian-meyer: Shouldn't this be something like `prev-page` instead of `first-page`?

Classes/Plugin/PageGrid.php
```diff
@@ -100,7 +100,7 @@ protected function getPageBrowser() {
                 if ($this->piVars['pointer'] != $i) {
                     $output .= $this->pi_linkTP_keepPIvars(sprintf($this->pi_getLL('page', '%d', TRUE), $i + 1), ['pointer' => $i, 'page' => ($i * $this->conf['limit']) + 1], TRUE).$separator;
                 } else {
-                    $output .= sprintf($this->pi_getLL('page', '%d', TRUE), $i + 1).$separator;
+                    $output .= '<span class="act-page">'.sprintf($this->pi_getLL('page', '%d', TRUE), $i + 1).'</span>'.$separator;
```
sebastian-meyer: I suggest using `active` here instead of `act-page` since that's the opposite of `not-active`.

sebastian-meyer (changes requested):
Just some minor wording suggestions.

Classes/Plugin/PageGrid.php
```diff
@@ -100,11 +100,11 @@ protected function getPageBrowser() {
...
                 }
                 $skip = TRUE;
             } elseif ($skip == TRUE) {
-                $output .= $this->pi_getLL('skip', '...', TRUE).$separator;
+                $output .= '<span class="skipped">'.$this->pi_getLL('skip', '...', TRUE).'</span>'.$separator;
```
sebastian-meyer, line 107: Added a `<span>` for skipped pages as well.

Classes/Plugin/PageGrid.php
BFallert: you are right, my mistake! 
```suggestion
            $output = '<span class="prev-page not-active">'.$this->pi_getLL('prevPage', '&lt;', TRUE).'</span>'.$separator;
```

Classes/Plugin/PageGrid.php
BFallert: ok, you have more experience
```suggestion
                    $output .= '<span class="active">'.sprintf($this->pi_getLL('page', '%d', TRUE), $i + 1).'</span>'.$separator;
```

BFallert:
I agree with all the adjustments

sebastian-meyer (approved):
Thank you very much!
(Sometimes I am a fusspot.)

PR #354, opened 2019-07-16
Optimize database handling

### LINKED ISSUE #353: Fix several slow queries
There are several slow queries in Kitodo.Presentation which take over 1 second.

One example is at [line 419 of OaiPmh.php](https://github.com/kitodo/kitodo-presentation/blob/fb3f2a4c49f2b35240cc9c48d16990190a638afd/Classes/Plugin/OaiPmh.php#L419).

sebastian-meyer:
- Changed field types from `text` to `varchar` where possible in order to make better use of indices
- Added more column indices and a multiple-column index to support frequent and/or complex queries
- Optimized some SQL queries for speed


sebastian-meyer:
This should fix #353

ByteParty (approved):

albig (approved):

PR #358, opened 2019-07-23
Backport database optimization from #354

sebastian-meyer:


albig (approved):

ext_tables.sql
```diff
@@ -7,38 +7,40 @@ CREATE TABLE tx_dlf_documents (
...
     collections int(11) DEFAULT '0' NOT NULL,
     owner int(11) DEFAULT '0' NOT NULL,
     solrcore int(11) DEFAULT '0' NOT NULL,
-    status tinyint(4) unsigned DEFAULT '0' NOT NULL,
+    status smallint(6) unsigned DEFAULT '0' NOT NULL,
 
     PRIMARY KEY (uid),
-    KEY parent (pid),
-    INDEX partof (partof)
+    KEY (pid),
```
albig: Actually this doesn't work at all with TYPO3 and Doctrine. The key needs an index name otherwise installation via extension manager is not possible.

See error message in TYPO3 8.7 (7.6 shoud be similar):

In SchemaException.php line 75:
                                                                         
  An index with name '' was already defined on table 'tx_dlf_documents'.

albig (changes requested):

ext_tables.sql
sebastian-meyer: I added index names according to #360 and enhanced the `language` index by adding `l18n_parent` as you did in #360 
All indexes should now be the same as in `master` which allows a smooth upgrade.

albig:
The following errors occure on executing the database compares on the first run:

> BLOB/TEXT column 'record_id' used in key specification without a key length
> BLOB/TEXT column 'location' used in key specification without a key length
> BLOB/TEXT column 'oai_name' used in key specification without a key length
> 

Executing a second time, no error message occures and the SQL fields are changed as expected.

PR #362, opened 2019-07-26
Tca migration and fixes

albig:
Most hints of the TCA migration wizard are solved in the following commit.

The path to the locallang_general.xlf file has changed in TYPO3 8.7. This has been fixed in all TCA configurations, too.

### DIFF
--- a/Configuration/TCA/tx_dlf_libraries.php
+++ b/Configuration/TCA/tx_dlf_libraries.php
@@ -35,15 +35,15 @@ return [
     'columns' => [
         'sys_language_uid' => [
             'exclude' => 1,
-            'label' => 'LLL:EXT:lang/locallang_general.xml:LGL.language',
+            'label' => 'LLL:EXT:lang/Resources/Private/Language/locallang_general.xlf:LGL.language',
             'config' => [
                 'type' => 'select',
                 'renderType' => 'selectSingle',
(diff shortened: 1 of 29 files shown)


Configuration/TCA/tx_dlf_libraries.php
```diff
@@ -121,7 +121,6 @@
                 'allowed' => $GLOBALS['TYPO3_CONF_VARS']['GFX']['imagefile_ext'],
                 'max_size' => 256,
                 'uploadfolder' => 'uploads/tx_dlf',
-                'show_thumbs' => 1,
```
sebastian-meyer, line 124: Does TYPO3 show thumbnails for images by default or does removing this option mean that thumbnails are no longer shown? Is there a new way to display thumbnails then?

albig: On [Deprecation: #79440 - TCA Changes](https://docs.typo3.org/c/typo3/cms-core/master/en-us/Changelog/8.6/Deprecation-79440-TcaChanges.html)  is written:

> show_thumbs has been dropped, the functionality has been transferred to fieldWizard as recordsOverview and fileThumbnails respectively and are enabled by default. They can be disabled by setting disabled=true 

So I understand, showing thumbnails is active as default.

sebastian-meyer (approved):

Issue #363, opened 2019-07-29
Plugin Template selection fails for tools of toolbox.


albig:
Each tool of the toolbox has it's own template. The getTemplate() method of the abstractPlugin class checks for a configuration `templateFile` which is always the toolbox itself but never the desired tool.

Proposal (pull request follows): Use one template file ("toolbox.tmpl") with several markers for each tool. Select the right marker inside the tool by calling `getTemplate('###MARKER###').`

sebastian-meyer:
The idea was to be able to provide additional tools in separate extension simply by registering them in the extension's `ext_localconf.php`:
https://github.com/kitodo/kitodo-presentation/blob/fcd78b6b9bb4ac3b0fd6cbf0c916e5433b8774e3/ext_localconf.php#L48
But this requires to have separate templates for every tool as well.

sebastian-meyer:
Instead of merging all templates together we should fix the template resolving in `getTemplate()`. This must have worked in the past, because otherwise no tool would have generated any output...

albig:
getTemplate was never used inside the tools. Because it doesn't work ;-)

Am 29.07.19 um 16:43 schrieb Sebastian Meyer:
>
> Instead of merging all templates together we should fix the template
> resolving in |getTemplate()|. This must have worked in the past,
> because otherwise no tool would have generated any output...
>

sebastian-meyer:
Hm, I don't see why though.

Here we instanciate a new `cObj` for every tool:
https://github.com/kitodo/kitodo-presentation/blob/fcd78b6b9bb4ac3b0fd6cbf0c916e5433b8774e3/Classes/Plugin/Toolbox.php#L54-L55

For those `cObj` we then call `getTemplate()`:
https://github.com/kitodo/kitodo-presentation/blob/fcd78b6b9bb4ac3b0fd6cbf0c916e5433b8774e3/Classes/Plugin/Tools/ImageManipulationTool.php#L42

Which looks at `$this->conf` for the configured template file:
https://github.com/kitodo/kitodo-presentation/blob/fcd78b6b9bb4ac3b0fd6cbf0c916e5433b8774e3/Classes/Common/AbstractPlugin.php#L57

`$this->conf` is inherited from the parent `cObj` of the toolbox itself:
https://github.com/kitodo/kitodo-presentation/blob/fcd78b6b9bb4ac3b0fd6cbf0c916e5433b8774e3/Classes/Plugin/Toolbox.php#L44-L47

[...]

PR #364, opened 2019-07-29
fix 363: move all tool templates inside the toolbox template file

albig:


sebastian-meyer:
See my comments in #363

albig:
With pull-request #365 this PR is obsolete.

PR #366, opened 2019-08-02
Replace old CLI by Symfony Command Controller

albig:
The CommandLineController is [announced as deprecated](https://docs.typo3.org/c/typo3/cms-core/master/en-us/Changelog/8.6/Deprecation-79265-CommandLineControllerAndCleanerCommand.html). As replacement is suggested to use native Symfonie commands or the Extbase Command Controller.

This patch introduces two Symfonie commands `kitodo:index` and `kitodo:reindex`.

`kitodo:index` does the same as the former "dlf index". The parameter of the options have changed slightly. Example: 
`./vendor/bin/typo3 kitodo:index -d http://example.com/path/mets.xml -p 123 -s dlfCore1`

It is possible to use the short parameter names ("-d") or the long ones ("--doc"). You may enter the solr core ("--solr") as UID or as core name. In both cases the existence of the solr core is checked on the given page ("--pid").

`kitodo:reindex` was used to index a single collection. This is still possible. But you can enter a list of collections as well. Documents will only be reindexed once even if they belong to several collections. Alternatively you can reindex all documents on the given page. Example:

    # reindex collection with uid 1 on page 123 with solr core 'dlfCore1'
    ./vendor/bin/typo3 kitodo:reindex -c 1 -p 123 -s dlfCore1

    # reindex collection with uid 1 and 4 on page 123 with solr core 'dlfCore1'
    ./vendor/bin/typo3 kitodo:reindex -c 1,4 -p 123 -s dlfCore1

    # reindex all documents on page 123 with solr core 'dlfCore1'
    ./vendor/bin/typo3 kitodo:reindex -a -p 123 -s dlfCore1

The new commands were added to the extension documentation. The SQL-queries inside the commands where replaced by Doctrine QueryBuilder.


sebastian-meyer:
In order to remove the old command line interface completely we have to remove the checks in `Classes/Hooks/ConfigurationForm.php` (and probably some other places) as well.

albig:
You are right. The check of an executable cliDispather is obsolete if we drop the old CLI. I added a patch to this pull request.
Maybe there are other places but I am not aware of it.

PR #367, opened 2019-08-14
Display month and day for table of content

beatrycze-volk:
kitodo


sebastian-meyer:
Does this need to be fixed in `master` as well?

beatrycze-volk:
Yes, I have open now PR for it.

PR #377, opened 2019-09-06
Fix two SQL queries

albig:
With #376 two bugs where introduced: 

- The SQL-statement to fetch the solr core is broken 
- the result of collections is always empty in the collection plugin.

This PR fixes these two errors.

### DIFF
--- a/Classes/Plugin/Collection.php
+++ b/Classes/Plugin/Collection.php
@@ -135,7 +135,7 @@ class Collection extends \Kitodo\Dlf\Common\AbstractPlugin {
         $count = $queryBuilder->count('uid')->execute()->fetchColumn(0);
         $content = '';
         if ($count == 1 && empty($this->conf['dont_show_single'])) {
-            $resArray = $allResults[0];
+            $resArray = $result->fetch();
             $this->showSingleCollection(intval($resArray['uid']));
         }
         $solr = Solr::getInstance($this->conf['solrcore']);
--- a/Classes/Plugin/Collection.php
+++ b/Classes/Plugin/Collection.php
@@ -104,7 +104,7 @@ class Collection extends \Kitodo\Dlf\Common\AbstractPlugin {
         }
 
         // Get collections.
-        $result = $queryBuilder
+        $queryBuilder
             ->select(
                 'tx_dlf_collections.index_name AS index_name',
                 'tx_dlf_collections.index_search as index_query',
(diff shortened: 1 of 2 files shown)


Classes/Plugin/Collection.php
```diff
@@ -129,12 +129,10 @@ protected function showCollectionList() {
...
-
-        $allResults = $result->fetchAll();
+            ->orderBy($orderBy);
 
-        $count = count($allResults);
+        $result = $queryBuilder->execute();
+        $count = $queryBuilder->count('uid')->execute()->fetchColumn(0);
         $content = '';
         if ($count == 1 && empty($this->conf['dont_show_single'])) {
             $resArray = $allResults[0];
```
sebastian-meyer: This won't work, since there is no `$allResults` anymore.

Classes/Plugin/Collection.php
```diff
@@ -129,12 +129,10 @@ protected function showCollectionList() {
...
             )
-            ->orderBy($orderBy)
-            ->execute();
-
-        $allResults = $result->fetchAll();
+            ->orderBy($orderBy);
 
-        $count = count($allResults);
+        $result = $queryBuilder->execute();
+        $count = $queryBuilder->count('uid')->execute()->fetchColumn(0);
```
sebastian-meyer, line 135: Does this really count the number of results for the query above? It seems to count the total number of collections in the database.

sebastian-meyer (changes requested):

Classes/Plugin/Collection.php
albig: Yes it does work. It is the same prepared query in  $queryBuilder, added the COUNT('uid') function.

Classes/Plugin/Collection.php
albig: You're right. I'll send a patch.

Unfortunately we have another bug above in the query. The orderBy with FIELD()-function seems not to be possible with doctrine :-(

sebastian-meyer (approved):

Classes/Plugin/Collection.php
sebastian-meyer: Do you provide a separate pull request for the `FIELD()` bug or should I keep this one open?

albig: I'll provide a separate patch. Actually, I'm not yet sure how to fix it.

PR #382, opened 2019-09-12
Several fixes on using the queryBuilder

albig:
There are still some few but important bugs introduced by #376.

This patch fixes the following:

* the list() function does not work with queryBuilder as it cannot work with associative arrays
* only the first metadata was indexed
* with a non-existing uid, the application stops with a SQL error in collection plugin

### DIFF
--- a/Classes/Common/Document.php
+++ b/Classes/Common/Document.php
@@ -451,10 +451,8 @@ abstract class Document {
                 ->setMaxResults(1)
                 ->execute();
 
-            $allResults = $result->fetchAll();
-
-            if (count($allResults) == 1) {
-                $documentFormat = $allResults[0]['document_format'];
+            if ($resArray = $result->fetch()) {
+                $documentFormat = $resArray['document_format'];
             }
         } else {
             // Get document format from content of remote document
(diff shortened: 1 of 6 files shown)


Classes/Common/Document.php
```diff
@@ -1540,10 +1532,13 @@ protected function __construct($uid, $pid, $preloadedDocument) {
...
 
-        $allResults = $result->fetchAll();
-
-        if (count($allResults) == 1) {
-            list ($this->uid, $this->pid, $this->recordId, $this->parentId, $this->thumbnail, $this->location) = $allResults[0];
+        if ($resArray = $result->fetch()) {
+            $this->uid = $resArray['uid'];
+            $this->pid = $resArray['pid'];
+            $this->recordId = $resArray['record_id'];
+            $this->uid = $resArray['uid'];
```
sebastian-meyer: This should be `$this->parentId` and probably something like `= $resArray['partof']`.

Classes/Common/DocumentTypeCheck.php
```diff
@@ -167,8 +167,8 @@ protected function loadDocument() {
                 ->execute();
 
             $allResults = $result->fetchAll();
```
sebastian-meyer: We don't need this line anymore, since you now fetch the result directly into `$resArray` below.

sebastian-meyer (changes requested):

PR #392, opened 2019-09-27
Workaround for toolbox plugin bug (if using custom template)

christopher-timm:
@sebastian-meyer: It is a bit different than discussed. But with this workaround there are no template changes needed.
I also removed the pi_wrapInBaseClass because it generates an extra div around each tool template.

### DIFF
--- a/Classes/Common/AbstractPlugin.php
+++ b/Classes/Common/AbstractPlugin.php
@@ -64,12 +64,18 @@ abstract class AbstractPlugin extends \TYPO3\CMS\Frontend\Plugin\AbstractPlugin 
      *
      * @return void
      */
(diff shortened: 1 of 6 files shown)


sebastian-meyer (commented):
Why did you remove `pi_wrapInBaseClass()`? An additional `div` doesn't hurt, but it provides explicit CSS selectors for every tool.

sebastian-meyer:
@chrizzor Just a reminder: Could you please revert removing `pi_wrapInBaseClass()`? It's the only issue preventing this from getting merged.

sebastian-meyer (approved):

sebastian-meyer:

PR #393, opened 2019-09-30
Add a new tool "search in document".

albig:
This adds a input field to the toolbox in case a fulltext file is
present. The search is executed only in this document. 20 results are
listed and you may page through all results in pages with 20 items.

The search is inspired by the SRU search in DFG-Viewer.

Please note: there is no design provided for this feature (as usual in
Kitodo.Presentation).

### DIFF
--- a/Classes/Common/Solr.php
+++ b/Classes/Common/Solr.php
@@ -110,7 +110,7 @@ class Solr {
         $helper = \TYPO3\CMS\Core\Utility\GeneralUtility::makeInstance(\Solarium\Core\Query\Helper::class);
         // Escape query phrase or term.
         if (preg_match('/^".*"$/', $query)) {
-            return '"'.$helper->escapePhrase(trim($query, '"')).'"';
+            return $helper->escapePhrase(trim($query, '"'));
         } else {
             return $helper->escapeTerm($query);
         }
--- a/Classes/Plugin/Tools/SearchInDocumentTool.php
+++ b/Classes/Plugin/Tools/SearchInDocumentTool.php
@@ -1,0 +1,117 @@
+<?php
+namespace Kitodo\Dlf\Plugin\Tools;
+
+/**
+ * (c) Kitodo. Key to digital objects e.V. <contact@kitodo.org>
+ *
+ * This file is part of the Kitodo and TYPO3 projects.
+ *
+ * @license GNU General Public License version 3 or later.
+ * For the full copyright and license information, please read the
+ * LICENSE.txt file that was distributed with this source code.
+ */
+
+use Kitodo\Dlf\Common\Helper;
+use TYPO3\CMS\Core\Utility\GeneralUtility;
+
+/**
+ * SearchInDocument tool for the plugin 'Toolbox' of the 'dlf' extension
+ *
+ * @author Sebastian Meyer <sebastian.meyer@slub-dresden.de>
+ * @author Alexander Bigga <alexander.bigga@slub-dresden.de>
+ * @package TYPO3
+ * @subpackage dlf
+ * @access public
+ */
+class SearchInDocumentTool extends \Kitodo\Dlf\Common\AbstractPlugin {
+    public $scriptRelPath = 'Classes/Plugin/Tools/SearchInDocumentTool.php';
+
+    /**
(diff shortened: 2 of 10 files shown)


Classes/Common/Solr.php
```diff
@@ -110,7 +110,7 @@ public static function escapeQuery($query) {
         $helper = \TYPO3\CMS\Core\Utility\GeneralUtility::makeInstance(\Solarium\Core\Query\Helper::class);
         // Escape query phrase or term.
         if (preg_match('/^".*"$/', $query)) {
-            return '"'.$helper->escapePhrase(trim($query, '"')).'"';
+            return $helper->escapePhrase(trim($query, '"'));
```
sebastian-meyer, line 113: Why did you remove the quotes? Without quotes phrases would become simple multi-term searches.

Classes/Plugin/Tools/SearchInDocumentTool.php
```diff
@@ -0,0 +1,124 @@
...
+        }
+
+        // Quit if no fulltext file is present
+        $fullTextFile = $this->doc->physicalStructureInfo[$this->doc->physicalStructure[$this->piVars['page']]]['files'][$this->conf['fileGrpFulltext']];
+        if (empty($fullTextFile)) {
+            return $content;
+        }
+
+        // Load template file.
+        $this->getTemplate();
```
sebastian-meyer: Template loading for tools has changed slightly (see #392).

```suggestion
        $this->getTemplate('###TEMPLATE###', '', TRUE);
```

sebastian-meyer (changes requested):

Classes/Common/Solr.php
sebastian-meyer: This affects every search, not only 'Search in Document'.

albig: This should have been another pull-request as this is a bug in Kitodo.Presentation since the switch to Solarium. At the moment, Kitodo.Presentation can't do phrase searches :-(

Example:

current patch
-----------------
The $query arrives at line 113 with this content:
`$query=""festung dresden""`
With trim($query, '"') we get
`$query = "festung dresden"`
With Solarium escapePhrase adds the Quotes again and we return at this point:
`return ""festung dresden""`

This results finally in the folowing query in case of the searchInDocument eID-script:
`http://solrserver:8983/solr/dlfCore2/select?wt=json&q=fulltext:(%22festung%20dresden%22)%20AND%20uid:11&hl=on&hl.fl=fulltext&fl=uid,id,page&hl.method=fastVector&start=0&rows=20`

Solr sees the following query parameter and executes a phrase search:
`"q":"fulltext:(\"festung dresden\") AND uid:11",`

Result textsnippet e.g.:
`"fulltext":["17. April 1945 Seite 19 Hermann Rahne Die »<em>Festung Dresden</em>« Seite 32 Christian Hermann Millionen Krie"]},`

current master
-------------------
With the additional quotes, the result would be:
`http://solrserver:8983/solr/dlfCore2/select?wt=json&q=fulltext:(%22%22festung%20dresden%22%22)%20AND%20uid:11&hl=on&hl.fl=fulltext&fl=uid,id,page&hl.method=fastVector&start=0&rows=20`

Solr sees the following query parameter and does a simple term search:
`"q":"fulltext:(\"\"festung dresden\"\") AND uid:11",`

Resulting textsnippet e.g.:
` "fulltext":["19 Hermann Rahne Die »<em>Festung</em> <em>Dresden</em>« von 1945 Am 16. April 1945 konnten die Dresdner Bürger in der"]},`

sebastian-meyer: So the problem seems to be: Why is the query double-quoted in the first place?

albig: It's not double quoted in the first place. That's only the output of xdebug.

The question is, why add the quotes another time? This is done already by escapePhrase(). This is what is does:
<pre>
public function escapePhrase($input)
{
return '"' . preg_replace('/("|\\\)/', '\\\$1', $input) . '"';
}
</pre>

sebastian-meyer: OK, got it!

sebastian-meyer (approved):

PR #394, opened 2019-10-02
License and Rights Information

sebastian-meyer:
This pull requests adds 5 additional metadata fields for documents to the database and the Solr index:

`license` holds an URI with license information
`terms` holds an URI with local terms of use
`restrictions` holds an URI with access restrictions
`out_of_print` holds text information about out of print works
`rights_info` holds general rights information

All fields are optional and repeatable. They were added to the `oai_dc` output of the OAI-PMH interface as well.

The implementation follows the guidelines of the [DINI AG KIM special interest group "Lizenzen"](https://wiki.dnb.de/display/DINIAGKIM/Lizenzen+Gruppe)


albig (approved):
Does out_of_print and rights_info needs really need a text field? Anyway, this should work.

We should backport this to 2.x.

sebastian-meyer:
From a METS/MODS perspective, all those fields are not limited to a certain length. So generally they all could exceed the limits of the current `VARCHAR` declaration in the schema. But since `out_of_print` and `rights_info` don't just hold URIs like the other fields, they are especially prone to hold quite long explanatory texts. So using `TEXT` is a safer approach. What do you recommend?

I also just found that I missed adding the fields to the Solr schema. I'll fix that before merging!

PR #404, opened 2019-10-15
Handle HTML entities in fulltext properly

### LINKED ISSUE #402: Escape special chars when handling fulltext data
Currently fulltext doesn't get properly escaped before output. Therefore fulltext could potentially contain malicious content like HTML tags or javascript. Simply treating the text with `htmlspecialchars()` would cause problems with result highlighting, so a more sophisticated approach is needed.
I propose using a Solr analyzer to escape at least `<` (and maybe more special characters) at index and query time.

sebastian-meyer:
This should fix #402.


sebastian-meyer:
@ByteParty Could you please have a look at this pull request? It should fix your reported problem with special chars like `<` in fulltext.

PR #408, opened 2019-10-21
Fix handling of tools' templates

### LINKED ISSUE #406: Overwriting the toolbox template is not possible
The tools template selection has been an issue serveral times already. The current code introduced with #392 a workaround which makes it impossible to overwrite the location of the tool template at all.

How to overwrite a Kitodo.Presenation Template?
-----------------------------------------------------------------
1. By Flexform
This is possible for all plugins except for the tools itself.

2. By TypoScript
This is possible for all plugins. It was possible for the tools, too.
E.g.
```
plugin.tx_dlf_fulltexttool {
  templateFile = EXT:myHelperExtension/Resources/Private/[...]/FulltextToolTemplate.tmpl
}
```
Current Behaviour
------------------------
The new code ignores the `templateFile`-configuration in case of tools and forces to the DLF default templates. This is inconsistent.

Solutions
------------
We had introduced to overcome this problem a new template marker `toolTemplateFile` in #129 which has been dropped on introducing AbstractPlugin::getTemplate() function.

Another idea would be to put all tool templates into the toolbox template and to distinguish the corresponding template part by the marker. e.g.`###TEMPLATE_TOOL_FULLTEXT###`.

We cannot use the fulltext-feature in DFG-Viewer for now because of this issue. Hoping for a quick solution.

sebastian-meyer:
This removes the special handling of tools' templates in `getTemplate()`. In exchange it introduces a default configuration for tool templates.

This fixes #406

### DIFF
--- a/Classes/Common/AbstractPlugin.php
+++ b/Classes/Common/AbstractPlugin.php
@@ -66,26 +66,16 @@ abstract class AbstractPlugin extends \TYPO3\CMS\Frontend\Plugin\AbstractPlugin 
      *
      * @return void
      */
-    protected function getTemplate($part = '###TEMPLATE###', $directory = '', $tool = FALSE) {
+    protected function getTemplate($part = '###TEMPLATE###') {
         $this->templateService = GeneralUtility::makeInstance(MarkerBasedTemplateService::class);
         if (!empty($this->conf['templateFile'])) {
             // Load template file from configuration.
-            if ($tool) {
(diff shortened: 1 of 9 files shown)


albig (approved):
Works for me as expected.

sebastian-meyer:
@chrizzor I tagged you for review because these changes might impact your work since they revert part of your pull request #392

PR #420, opened 2019-11-05
Fix scrutinizer issues

sebastian-meyer:
This fixes several issues reported by scrutinizer (see https://scrutinizer-ci.com/g/kitodo/kitodo-presentation/issues/master).

### DIFF
--- a/Classes/Command/IndexCommand.php
+++ b/Classes/Command/IndexCommand.php
@@ -87,13 +87,15 @@ class IndexCommand extends Command
             exit(1);
         }
 
-        if ($input->getOption('solr')) {
+        if (
+            !empty($input->getOption('solr'))
+            && !is_array($input->getOption('solr'))
+        ) {
             $allSolrCores = $this->getSolrCores($startingPoint);
             if (MathUtility::canBeInterpretedAsInteger($input->getOption('solr'))) {
                 $solrCoreUid = MathUtility::forceIntegerInRange((int) $input->getOption('solr'), 0);
             } else {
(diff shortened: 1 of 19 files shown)


Classes/Command/IndexCommand.php
```diff
@@ -87,13 +87,15 @@ protected function execute(InputInterface $input, OutputInterface $output)
             exit(1);
         }
 
-        if ($input->getOption('solr')) {
+        if (
+            !empty($input->getOption('solr'))
+            && !is_array($input->getOption('solr'))
```
albig, line 92: `$input->getOption('solr')` is never an array. This is garantied by symfony. 
To accept arrays, we must add `InputOption::VALUE_IS_ARRAY` which is not what we want

Classes/Command/IndexCommand.php
```diff
@@ -110,15 +112,19 @@ protected function execute(InputInterface $input, OutputInterface $output)
...
-            $io->error('ERROR: Required parameter --solr|-s is missing.');
+            $io->error('ERROR: Required parameter --solr|-s is missing or array.');
             exit(1);
         }
 
         if (
-            !MathUtility::canBeInterpretedAsInteger($input->getOption('doc'))
-            && !GeneralUtility::isValidUrl($input->getOption('doc'))
+            empty($input->getOption('doc'))
+            || is_array($input->getOption('doc'))
```
albig, line 121: Same here. doc cannot be an array.

albig (changes requested):

albig (approved):

PR #426, opened 2019-11-08
Always check for valid xpath results

sebastian-meyer:


albig (approved):
Works (again) for documents without physical structure (likenewspaper anchors of works with several volumes).

PR #435, opened 2019-11-13
Fix search truncation with '*' and '?'

### LINKED ISSUE #150: search truncation with * does not work any longer
As mentioned in issue#21 search truncation with \* does not work any longer (since fix#34).
Using '\*' still works - but 'fragment\*' does not work.
This kind of search is very important for our users.

sebastian-meyer:
This should fix #150

### DIFF
--- a/Classes/Common/Solr.php
+++ b/Classes/Common/Solr.php
@@ -121,7 +121,10 @@ class Solr
         if (preg_match('/^".*"$/', $query)) {
             return $helper->escapePhrase(trim($query, '"'));
         } else {
-            return $helper->escapeTerm($query);
+            // Using a modified escape function here to retain '*' and '?' for search truncation.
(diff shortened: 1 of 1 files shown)


albig (approved):
This works for a single term like "Dresd*" to find "Dresdner", "Dresden" etc.

This does not work for 'Dresdner Heft*" to find "Dresdner Hefte" as this is escaped to "Dresdner\ Heft*" which finds nothing in Solr. I think this is another story, as a query lilke 'Dresdner Heft*' (entered without quotes) must not go into else here.

sebastian-meyer:
I added another commit turning off escaping of whitespaces.

albig (approved):
Works now for 'Dresdner Hef*' or 'Dresdner Heft?'.

PR #436, opened 2019-11-13
Add support for seasonal newspapers and periodicals

sebastian-meyer:
This applies the changes and fixes from 2.x to master


sebastian-meyer:
There is one change in 2.x I didn't apply to master:
https://github.com/kitodo/kitodo-presentation/blob/2.x/common/class.tx_dlf_document.php#L359

I think we don't need this and it's rather ugly anyway.

albig (commented):
Looks quite good. I've currently no fully working setup to test this with TYPO3 8.7. Will continue tomorrow.

albig (approved):
Works now in my setup.

PR #438, opened 2019-11-14
Cleanup for release

sebastian-meyer:


albig (approved):
Document switch seems to work properly with this condition for the years overview:
`[userFunc = user_dlf_docTypeCheck(newspaper)] || [userFunc = user_dlf_docTypeCheck(ephemera)]`

PR #439, opened 2019-11-14
Fix docTypeCheck for non-indexed documents

sebastian-meyer:
The new docTypeCheck doesn't work for non-indexed documents (i. e. the DFG Viewer), because those documents don't bring their own PID for the metadata configuration. So we have to set a PID explicitly when calling the user function from Typoscript:
`[userFunc = user_dlf_docTypeCheck($type, $pid)]`

### DIFF
--- a/Classes/Hooks/UserFunc.php
+++ b/Classes/Hooks/UserFunc.php
@@ -64,9 +64,11 @@ class UserFunc
      *
      * @access public
      *
+     * @param int $cPid The PID for the metadata definitions
+     *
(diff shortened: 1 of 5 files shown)


sebastian-meyer:
For the DFG Viewer, the PID would be `config.storagePid`, so the user function should look like:
`[userFunc = user_dlf_docTypeCheck(year, {$config.storagePid})]`

albig (approved):
You're right. Have tested the last change of getDocumentType() obviously only with indexed documents.
This solution seems to work.

PR #443, opened 2019-11-15
Sort collection's entries by title initially

### LINKED ISSUE #442: Collection listview is not sorted on initial request
If you call a single collection view (collection plugin -> listview) the listing is sorted by UIDs but not by title as claimed by the "SortingForm".

On changing the settings in the sorting form (sort by or order by), the results get sorted.

This applies to 2.x and 3.0.

sebastian-meyer:
This should fix #442


albig (approved):
So easy :-) Thank you!

PR #452, opened 2020-01-03
Resize to options field of the tx_dlf_tokens table.

albig:
The options field keeps the serialized array of found UIDs of an OAI
set. With 6-digit UIDs it is only possible to have about 6500 objects in
an OAI set. In this case the serialized array is cutted an the OAI
plugin will fail with an exception.

Before commit 443aaff2cbb79870ae3f1ebecd3f7460f81e2812 this field was
textlong which might be really too long. This patch changes the field to
mediumtext wich should be able to keep about 1.6 million objects.


sebastian-meyer (approved):
I tried to avoid non-standard SQL field types, but since we are using Doctrine which handles differences between common SQL implementations quite well this shouldn't be a problem.

PR #476, opened 2020-02-10
Enhance the calendar plugin.

albig:
The calendar plugin is not usable without a TypoScrip condition or a
condition inside Fluid. The main() method is without any function so
far.

This patch adds the possibility to place the plugin on a separate page
like the pageview plugin.

The main method now does the type-check. Additionally it is possible to
configure an initial document. E.g. the anchor file of a newspaper. So
you may do a page with a calendar of a special newspaper or a single
year.

### DIFF
--- a/Classes/Plugin/Calendar.php
+++ b/Classes/Plugin/Calendar.php
@@ -49,6 +49,43 @@ class Calendar extends \Kitodo\Dlf\Common\AbstractPlugin
      */
     public function main($content, $conf)
     {
+        $this->init($conf);
+
+        // Set initial document (anchor or year file) if configured.
+        if (empty($this->piVars['id']) && !empty($this->conf['initialDocument'])) {
+            $this->piVars['id'] = $this->conf['initialDocument'];
+        }
+
+        // Load current document.
+        $this->loadDocument();
+        if ($this->doc === null) {
+            // Quit without doing anything if required variables are not set.
+            return $content;
+        }
+
+        $metadata = $this->doc->getTitledata($this->doc->cPid);
+        if (!empty($metadata['type'][0])) {
+            // Calendar plugin does not support IIIF (yet). Abort for all newspaper related types.
+            if (
+                $doc instanceof IiifManifest
+                && array_search($metadata['type'][0], ['newspaper', 'ephemera', 'year', 'issue']) !== false
+            ) {
+                return $type;
+            }
+            $type = $metadata['type'][0];
+        }
+
+        switch ($type) {
+            case 'newspaper':
+            case 'ephemera':
+                return $this->years($content, $conf);
+            case 'year':
+                return $this->calendar($content, $conf);
+            case 'issue':
+                default:
+                break;
+        }
+
         // Nothing to do here.
         return $content;
     }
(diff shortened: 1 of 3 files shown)


Configuration/Flexforms/Calendar.xml
```diff
@@ -40,6 +40,46 @@
...
+                                <minitems>0</minitems>
+                                <show_thumbs>1</show_thumbs>
+                                <wizards>
+                                    <suggest>
+                                        <type>suggest</type>
+                                    </suggest>
+                                </wizards>
+                            </config>
+                        </TCEforms>
+                    </initialDocument>
```
sebastian-meyer, line 82: I am not happy with this. Having just a single newspaper to present seems to be mostly an edge case. Most people don't need this and it can be achieved quite easily with pure typoscript as well:
`plugin.tx_dlf_calendar._DEFAULT_PI_VARS.id = [UID of document]`

I'd like to keep the flexform as clear as possible and suggest removing this feature. What do you think?

Classes/Plugin/Calendar.php
```diff
@@ -49,6 +49,43 @@ class Calendar extends \Kitodo\Dlf\Common\AbstractPlugin
...
+        }
+
+        $metadata = $this->doc->getTitledata($this->doc->cPid);
+        if (!empty($metadata['type'][0])) {
+            // Calendar plugin does not support IIIF (yet). Abort for all newspaper related types.
+            if (
+                $doc instanceof IiifManifest
+                && array_search($metadata['type'][0], ['newspaper', 'ephemera', 'year', 'issue']) !== false
+            ) {
+                return $type;
```
sebastian-meyer: This doesn't make any sense. Why return anything when `$this->doc` is an `IiifManifest` which we don't support? And `$type` isn't defined anyways... Maybe a copy & paste error from `docTypeCheck`?

Classes/Plugin/Calendar.php
```diff
@@ -49,6 +49,43 @@ class Calendar extends \Kitodo\Dlf\Common\AbstractPlugin
...
+        }
+
+        // Load current document.
+        $this->loadDocument();
+        if ($this->doc === null) {
+            // Quit without doing anything if required variables are not set.
+            return $content;
+        }
+
+        $metadata = $this->doc->getTitledata($this->doc->cPid);
```
sebastian-meyer: We don't need to give `$this->doc->cPid` as parameter, since this is the default. Also, `$this->doc->cPid` is *write-only* and should not be used directly.
```suggestion
        $metadata = $this->doc->getTitledata();
```

Classes/Plugin/Calendar.php
```diff
@@ -49,6 +49,43 @@ class Calendar extends \Kitodo\Dlf\Common\AbstractPlugin
...
+        if (!empty($metadata['type'][0])) {
+            // Calendar plugin does not support IIIF (yet). Abort for all newspaper related types.
+            if (
+                $doc instanceof IiifManifest
```
sebastian-meyer:
```suggestion
                $this->doc instanceof IiifManifest
```

sebastian-meyer (changes requested):
There are some errors and a general question regarding adding an edge case feature.

Classes/Plugin/Calendar.php
albig: This is just cut&copy code from `docTypeCheck`. Will send a new commit to fix this.

Configuration/Flexforms/Calendar.xml
albig: I would provide both ways. The first one via flexform is more visible and an editor can use it too, to show the calendar somewhere in TYPO3. 
Additionally, I can overwrite it by a TypoScript setting which is bit more catchy with the other settings, we are using. E.g.:
```
plugin.tx_dlf_calendar {
    pages = {$config.storagePid}
    targetPid.data = TSFE:page|uid
    initialDocument = 10
    templateFile = {$config.templateFileCalendar}
}
```

sebastian-meyer (approved):

PR #478, opened 2020-02-12
Proposal: Link to PURL in listview and calendar plugin.

albig:
This is a new feature which makes it possible to use the PURL to link
documents out of the search, listview and calendar plugins.

To enable this feature, you can use the TypoScript setting
  `plugin.tx_dlf.linkToPurl = 1`

Serveral questions appear during implementation:

* Should we add the setting to the extension configuration (globaly),
  too?
* Which plugins make it sense to add this feature? I could imagine the
  feed plugin, maybe the OAI-plugin?
* The highlight_word parameter will be appended to the PURL. Otherwise
  it get's lost from listview -> pageview. Is this a clean way? The we
  must enable the resolver to allow additional parameters like this.

### DIFF
--- a/Classes/Plugin/Calendar.php
+++ b/Classes/Plugin/Calendar.php
@@ -115,6 +115,7 @@ class Calendar extends \Kitodo\Dlf\Common\AbstractPlugin
             ->select(
                 'tx_dlf_documents.uid AS uid',
                 'tx_dlf_documents.title AS title',
+                'tx_dlf_documents.purl AS purl',
                 'tx_dlf_documents.year AS year',
                 'tx_dlf_documents.mets_label AS label',
                 'tx_dlf_documents.mets_orderlabel AS orderlabel'
@@ -144,7 +145,8 @@ class Calendar extends \Kitodo\Dlf\Common\AbstractPlugin
             $issues[] = [
                 'uid' => $resArray['uid'],
                 'title' => $title,
-                'year' => $resArray['year']
+                'year' => $resArray['year'],
+                'purl' => $resArray['purl'],
             ];
         }
         //  We need an array of issues with year => month => day number as key.
@@ -297,7 +299,7 @@ class Calendar extends \Kitodo\Dlf\Common\AbstractPlugin
                                         $dayLinkLabel = empty($issue['title']) ? strftime('%x', $currentDayTime) : $issue['title'];
                                         $linkConf = [
                                             'useCacheHash' => 1,
-                                            'parameter' => $this->conf['targetPid'],
+                                            'parameter' => $this->conf['linkToPurl'] ? $issue['purl'] : $this->conf['targetPid'],
                                             'additionalParams' => '&' . $this->prefixId . '[id]=' . urlencode($issue['uid']),
                                             'ATagParams' => ' class="title"',
                                         ];
@@ -384,6 +386,7 @@ class Calendar extends \Kitodo\Dlf\Common\AbstractPlugin
             ->select(
                 'tx_dlf_documents.uid AS uid',
                 'tx_dlf_documents.title AS title',
+                'tx_dlf_documents.purl AS purl',
                 'tx_dlf_documents.mets_label AS label',
                 'tx_dlf_documents.mets_orderlabel AS orderlabel'
             )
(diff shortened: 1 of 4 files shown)


Classes/Plugin/Calendar.php
```diff
@@ -297,7 +299,7 @@ protected function getCalendarYear($calendarIssuesByMonth, $year, $firstMonth =
                                         $dayLinkLabel = empty($issue['title']) ? strftime('%x', $currentDayTime) : $issue['title'];
                                         $linkConf = [
                                             'useCacheHash' => 1,
-                                            'parameter' => $this->conf['targetPid'],
+                                            'parameter' => $this->conf['linkToPurl'] ? $issue['purl'] : $this->conf['targetPid'],
```
sebastian-meyer, line 302: Before linking to `$issue['purl']` we should check if it isn't empty. Otherwise we have to fall back to using the internal linking.

```suggestion
                                            'parameter' => $this->conf['linkToPurl'] && !empty($issue['purl']) ? $issue['purl'] : $this->conf['targetPid'],
```

Also, when using the PURL the additional parameters are obsolete and should be omitted.

Classes/Plugin/ListView.php
```diff
@@ -163,7 +163,8 @@ protected function getEntry($number, $template)
                         }
                         $conf = [
                             'useCacheHash' => 1,
-                            'parameter' => $this->conf['targetPid'],
+                            'parameter' => $this->conf['linkToPurl'] ? ($this->list[$number]['purl'] . '/' . $this->list[$number]['page']) : $this->conf['targetPid'],
```
sebastian-meyer, line 166: This is a very SLUB-specific way of creating the PURL as it assumes that the resolver knows how to handle PURL with concatenated page numbers... 😞

Classes/Plugin/ListView.php
```diff
@@ -324,18 +325,27 @@ protected function getSubEntries($number, $template)
...
+                        }
                         if (!empty($this->piVars['logicalPage'])) {
                             $additionalParams['logicalPage'] = $this->piVars['logicalPage'];
                         }
                         $conf = [
                             // we don't want cHash in case of search parameters
                             'useCacheHash' => empty($this->list->metadata['searchString']) ? 1 : 0,
-                            'parameter' => $this->conf['targetPid'],
+                            'parameter' => $this->conf['linkToPurl']
+                                ? ($this->list[$number]['purl'] . '/' . $subpart['page'] . '?' . \TYPO3\CMS\Core\Utility\GeneralUtility::implodeArrayForUrl($this->prefixId, $additionalParams, '', true, false))
```
sebastian-meyer, line 346: See above: This works for the SLUB resolver, but not for anyone else.

Documentation/Plugins/Index.rst
```diff
@@ -19,8 +19,41 @@ Plugin Reference
...
+
+.. t3-field-list-table::
+ :header-rows: 1
+
+ - :Property:
+       Property
+   :Data Type:
+       Data type
+   :Default:
+        Default
```
sebastian-meyer, line 41:
```suggestion
       Default
```

sebastian-meyer (changes requested):

sebastian-meyer:
We need to make this more generic in order to be useful not only with the SLUB resolver.

> * Should we add the setting to the extension configuration (globaly),
>   too?

Using this as a global standard would be a very specific and rare edge-case, so we don't need to support this by explicitly adding a configuration option. If somebody needs it anyway, it can be easily accomplished with Typoscript as well.

> * Which plugins make it sense to add this feature? I could imagine the
>   feed plugin, maybe the OAI-plugin?

It would be useful for the `Feeds` plugin. The `OaiPmh` plugin doesn't link to a page view, so it's not applicable there.

> * The highlight_word parameter will be appended to the PURL. Otherwise
>   it get's lost from listview -> pageview. Is this a clean way? The we
>   must enable the resolver to allow additional parameters like this.

Although it would be a more complex solution, I suggest changing the TS option `plugin.tx_dlf.linkToPurl` from a binary on/off switch to some sort of link template. The template should contain markers for PURL, page number and search words. When such a template is given, it should be used for creating the links in `ListView`, `Calendar` and `Feeds`.

For example (for the SLUB resolver):
`plugin.tx_dlf.linkToPurl = ###PURL###/###PAGE_NUMBER###?highlight=###SEARCH_WORDS###`

albig:
Won't follow this idea anymore.

PR #489, opened 2020-03-04
Fix codacy security issues

sebastian-meyer:
Clean up javascript code and fix a lot of reported codacy issues.


albig (approved):
All known JavaScript works for me in pageview (images are loaded, fulltext is shown, word highlightning works, imagemanipulation works, ...).

PR #491, opened 2020-03-13
Listview numeration

christopher-timm:
This PR fixes the listview plugin numeration


albig (approved):
Works for me. 

E.g. on Page 2 with 30 items per page:

before:
```
<ol class="tx-dlf-listview-list">
    <li value="1">
[...]
```
after:
```
<ol class="tx-dlf-listview-list">
    <li value="31">
[...]
```

sebastian-meyer (approved):

PR #492, opened 2020-03-13
Facet collection configuration fix

christopher-timm:
This part controls the output of the collection facet.
With $collection[0] the facetCollectionArray only gets filled with NULL instead of the collection name. The PR should fix it.


christopher-timm:
Sry, was on tag 3.0.1 and not on master

PR #493, opened 2020-03-13
Fill orderlabel for sub-entries in TableOfContents plugin.

albig:
With newspaper/ephemera year files the fields title, year, volume,
volume_sorting are not filled. The TableOfContents plugin only shows the
structure type ("Year") on the anchor level.

The label we want to see is the mets_orderlabel in this case. But this
field is not queried in the plugin.

Filling the internal "orderlabel" field in the menuArray fixes this
issue. The "orderlabel" will be filled into the "title" field in
getMenuEntry().

Another issue is fixed in this patch: The ordering of the menu entries.
The sorting does not work if volume is not filled. In this case the
sorting by mets_orderlabel must be used.


sebastian-meyer (approved):
I've added `mets_label` as well in order to handle sub-entries the same way as toplevel entries. For toplevel entries `mets_label` is used instead of `title`. For sub-entries we can keep `title` as fallback for when `mets_label` isn't set.

PR #499, opened 2020-04-06
refactores $GLOBALS['TYPO3_DB'] to QueryBuilder

erik-konrad:
This pull request refactores $GLOBALS['TYPO3_DB'] to QueryBuilder functionality for TYPO3 9LTS compatibility.
* all methods are refactored for QueryBuilder usage
* Helper::whereClause() removed (check for hidden and deleted is default restriction configuration in querybuilder)
* general code cleanup

### DIFF
--- a/class.ext_update.php
+++ b/class.ext_update.php
@@ -40,17 +40,21 @@ class ext_update
      *
      * @return bool Should the update option be shown?
      */
-    public function access()
+    public function access(): bool
     {
         if (count($this->getMetadataConfig())) {
             return true;
-        } elseif ($this->oldIndexRelatedTableNames()) {
+        }
+        if ($this->oldIndexRelatedTableNames()) {
             return true;
-        } elseif ($this->solariumSolrUpdateRequired()) {
+        }
+        if ($this->solariumSolrUpdateRequired()) {
             return true;
-        } elseif (count($this->oldFormatClasses())) {
+        }
+        if (count($this->oldFormatClasses())) {
             return true;
-        } elseif ($this->hasNoFormatForDocument()) {
+        }
+        if ($this->hasNoFormatForDocument()) {
             return true;
         }
         return false;
@@ -63,7 +67,7 @@ class ext_update
      *
      * @return string The content that is displayed on the website
      */
-    public function main()
+    public function main(): string
     {
         // Load localization file.
         $GLOBALS['LANG']->includeLLFile('EXT:dlf/Resources/Private/Language/FlashMessages.xml');
@@ -86,7 +90,7 @@ class ext_update
         }
         return $this->content;
     }
-    
+
     /**
      * Get all outdated metadata configuration records
      *
@@ -94,7 +98,7 @@ class ext_update
      *
      * @return array Array of UIDs of outdated records
      */
-    protected function getMetadataConfig()
+    protected function getMetadataConfig(): array
     {
         $uids = [];
         // check if tx_dlf_metadata.xpath exists anyhow
@@ -107,15 +111,14 @@ class ext_update
 
         $rows = $result->fetchAll();
 
-        if((count($rows) === 0) || !in_array('xpath', array_keys($rows[0]))) {
+        if((count($rows) === 0) || !array_key_exists('xpath', $rows[0])) {
             return $uids;
-        } else {
-            foreach($rows as $row) {
-                if($row['format'] === 0 && $row['xpath']) {
-                    $uids[] = intval($row['uid']);
-                }
+        }
+        foreach($rows as $row) {
+            if($row['format'] === 0 && $row['xpath']) {
+                $uids[] = (int)$row['uid'];
             }
-        };
+        }
         return $uids;
     }
 
@@ -126,7 +129,7 @@ class ext_update
      *
      * @return bool
      */
-    protected function solariumSolrUpdateRequired()
+    protected function solariumSolrUpdateRequired(): bool
     {
         // Get all Solr cores that were not deleted.
         $queryBuilder = GeneralUtility::makeInstance(ConnectionPool::class)->getQueryBuilderForTable('tx_dlf_solrcores');
@@ -152,7 +155,7 @@ class ext_update
      *
      * @return array containing old format classes
      */
-    protected function oldFormatClasses()
+    protected function oldFormatClasses(): array
     {
         $oldRecords = [];
         // Get all records with outdated configuration.
@@ -178,7 +181,7 @@ class ext_update
      *
      * @return bool true if old index related columns exist
      */
-    protected function oldIndexRelatedTableNames()
+    protected function oldIndexRelatedTableNames(): bool
     {
         $queryBuilder = GeneralUtility::makeInstance(ConnectionPool::class)->getQueryBuilderForTable('INFORMATION_SCHEMA.COLUMNS');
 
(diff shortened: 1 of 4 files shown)


class.ext_update.php
```diff
@@ -116,26 +91,86 @@ public function main()
...
+            ->select('tx_dlf_formats.uid AS uid', 'tx_dlf_formats.type AS type')
+            ->from('tx_dlf_formats')
+            ->where('tx_dlf_formats.class IS NOT NULL AND tx_dlf_formats.class != "" AND tx_dlf_formats.class NOT LIKE "%\\\\\\\\%"')
+            ->execute();
+
+        while ($resArray = $result->fetch()) {
             $oldRecords[$resArray['uid']] = $resArray['type'];
         }
+
         return $oldRecords;
```
albig, line 177: Actually this Query doesn't work as expected. As "oldFormats", also current ones are found:
```
array (size=3)
  1 => string 'MODS' (length=4)
  2 => string 'TEIHDR' (length=6)
  3 => string 'ALTO' (length=4)
```
This is set by Kitodo.Presentation 3.0 and is working on some installations:
```
MariaDB [typo3]> select uid,type,class from tx_dlf_formats WHERE class IS NOT NULL AND class != "";
+-----+--------+--------------------------------+
| uid | type   | class                          |
+-----+--------+--------------------------------+
|   1 | MODS   | Kitodo\\Dlf\\Format\\Mods      |
|   2 | TEIHDR | Kitodo\\Dlf\\Format\\TeiHeader |
|   3 | ALTO   | Kitodo\\Dlf\\Format\\Alto      |
+-----+--------+--------------------------------+
3 rows in set (0.001 sec)
```

class.ext_update.php
```diff
@@ -38,61 +40,34 @@ class ext_update
...
+        }
+        if ($this->oldIndexRelatedTableNames()) {
             return true;
-        } elseif ($this->solariumSolrUpdateRequired()) {
+        }
+        if ($this->solariumSolrUpdateRequired()) {
             return true;
-        } elseif (count($this->oldFormatClasses())) {
+        }
+        if (count($this->oldFormatClasses())) {
```
albig, line 54: This test will always be run --> see below in https://github.com/kitodo/kitodo-presentation/pull/499/files#r404744493

class.ext_update.php
```diff
@@ -375,58 +452,27 @@ protected function doSolariumSolrUpdate()
...
-        $result = $GLOBALS['TYPO3_DB']->sql_query($sqlQuery);
+        $queryBuilder = GeneralUtility::makeInstance(ConnectionPool::class)->getQueryBuilderForTable('tx_dlf_documents');
+
+        $result = $queryBuilder
+            ->update('tx_dlf_documents')
+            ->where(
+                $queryBuilder->expr()->orX(
+                    $queryBuilder->expr()->eq('document_format', $queryBuilder->createNamedParameter(null)),
+                    $queryBuilder->expr()->eq('document_format', $queryBuilder->createNamedParameter(''))
+                ),
```
albig: This comma (`,`) breaks TYPO3 8.7 ;-)
```
syntax error, unexpected ')'
```

ext_emconf.php
```diff
@@ -13,12 +13,12 @@
 $EM_CONF[$_EXTKEY] = [
     'title' => 'Kitodo.Presentation',
     'description' => 'Base plugins, modules, services and API of the Digital Library Framework. It is part of the community-based Kitodo Digitization Suite.',
-    'version' => '3.0.0',
+    'version' => '3.0.1',
```
albig: Shouldn't we rise to 3.1.0 already? To avoid problem with next release?

albig (changes requested):
The one query may be wrong in current code, but should be fixed.

Without the mentioned comma, the code is running in 9.5 and 8.7. But I haven't tested the update queries itself.

class.ext_update.php
erik-konrad: fixed

class.ext_update.php
erik-konrad: also fixed

class.ext_update.php
erik-konrad: yes :-) thats the reason why TYPO3 is named TYPO3 :-D a simple syntax error. fixed.

ext_emconf.php
erik-konrad: done.

class.ext_update.php
```diff
@@ -116,26 +91,89 @@ public function main()
...
-        while ($resArray = $GLOBALS['TYPO3_DB']->sql_fetch_assoc($result)) {
+        $queryBuilder = GeneralUtility::makeInstance(ConnectionPool::class)->getQueryBuilderForTable('tx_dlf_formats');
+
+        $result = $queryBuilder
+            ->select('tx_dlf_formats.uid AS uid', 'tx_dlf_formats.type AS type')
+            ->from('tx_dlf_formats')
+            ->where(
+                $queryBuilder->expr()->isNotNull('tx_dlf_formats.class'),
+                $queryBuilder->expr()->neq('tx_dlf_formats.class', $queryBuilder->createNamedParameter('')),
+                $queryBuilder->expr()->notLike('tx_dlf_formats.class', $queryBuilder->createNamedParameter('%\\\\%'))
```
albig: This looks better but still does not work. The below values are correct and shouldn't be touched.
```
MariaDB [typo3]> select uid,type,class from tx_dlf_formats WHERE class IS NOT NULL AND class != "" AND class NOT LIKE '%\\\\%';
+-----+--------+--------------------------------+
| uid | type   | class                          |
+-----+--------+--------------------------------+
|   1 | MODS   | Kitodo\\Dlf\\Format\\Mods      |
|   2 | TEIHDR | Kitodo\\Dlf\\Format\\TeiHeader |
|   3 | ALTO   | Kitodo\\Dlf\\Format\\Alto      |
+-----+--------+--------------------------------+
3 rows in set (0.001 sec)
```
The check should only return old formats like "tx_dlf_alto" (or how this old ALTO class was named).
Maybe @sebastian-meyer could explain the history.

albig (changes requested):

class.ext_update.php
erik-konrad: Thats weird. If I run this query, I get the following result:

`
MariaDB [kitodo9]> select uid,type,class from tx_dlf_formats WHERE class IS NOT NULL AND class != "" AND class NOT LIKE '%\\\\%';
Empty set (0.000 sec)
`

erik-konrad: I changed the part `NOT LIKE '%\\\\%'` to `LIKE '%tx_dlf_%'`. Now this should work as expected.

albig (approved):
Works for me on 8.7 and 9.5 now.

class.ext_update.php
```diff
@@ -116,26 +91,89 @@ public function main()
...
-        while ($resArray = $GLOBALS['TYPO3_DB']->sql_fetch_assoc($result)) {
+        $queryBuilder = GeneralUtility::makeInstance(ConnectionPool::class)->getQueryBuilderForTable('tx_dlf_formats');
+
+        $result = $queryBuilder
+            ->select('tx_dlf_formats.uid AS uid', 'tx_dlf_formats.type AS type')
+            ->from('tx_dlf_formats')
+            ->where(
+                $queryBuilder->expr()->isNotNull('tx_dlf_formats.class'),
+                $queryBuilder->expr()->neq('tx_dlf_formats.class', $queryBuilder->createNamedParameter('')),
+                $queryBuilder->expr()->like('tx_dlf_formats.class', $queryBuilder->createNamedParameter('%tx_dlf_%'))
```
sebastian-meyer, line 170: This won't work for users which use custom classes for their format handling. But I don't know of anybody who does this currently, so we should be fine anyways.

sebastian-meyer (approved):

PR #500, opened 2020-04-08
Fix syntax error breaking TYPO3 8.7

albig:
Running current code causes a TYPO3 exception in TYPO3 8.7.

> syntax error, unexpected ')'


albig:
Sorry. Polluted PR.

PR #505, opened 2020-04-23
Add CORS header to file proxy

sebastian-meyer:
This adds a valid CORS header for files delivered via the `PageViewProxy`. This is needed when integrating the Page View with other applications on a different domain (i. e. the Zeitungsportal).

### DIFF
--- a/Classes/Plugin/Eid/PageViewProxy.php
+++ b/Classes/Plugin/Eid/PageViewProxy.php
@@ -69,6 +69,7 @@ class PageViewProxy
         $response = GeneralUtility::makeInstance(Response::class);
         if ($fetchedData) {
             $response->getBody()->write($fetchedData);
+            $response = $response->withHeader('Access-Control-Allow-Origin', '*');
             $response = $response->withHeader('Content-Type', finfo_buffer(finfo_open(FILEINFO_MIME), $fetchedData));
         }
         if ($header === 0 && !empty($lastModified)) {
--- a/Classes/Plugin/Eid/PageViewProxy.php
+++ b/Classes/Plugin/Eid/PageViewProxy.php
@@ -69,8 +69,8 @@ class PageViewProxy
         $response = GeneralUtility::makeInstance(Response::class);
         if ($fetchedData) {
             $response->getBody()->write($fetchedData);
-            $response = $response->withHeader('Access-Control-Allow-Origin', '*');
             $response = $response->withHeader('Access-Control-Allow-Methods', 'GET');
+            $response = $response->withHeader('Access-Control-Allow-Origin', $request->getHeader('Origin'));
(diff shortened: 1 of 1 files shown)


Classes/Plugin/Eid/PageViewProxy.php
```diff
@@ -69,6 +69,7 @@ public function main(ServerRequestInterface $request)
         $response = GeneralUtility::makeInstance(Response::class);
         if ($fetchedData) {
             $response->getBody()->write($fetchedData);
+            $response = $response->withHeader('Access-Control-Allow-Origin', '*');
```
albig: This will work, no doubt. 
But shouldn't we respect the CORS-settings from the source? Isn't this this a security issue? You may use the PageViewProxy for every data on the web and add CORS access to any domain. 
So maybe, we should make the origin domain configurable somehow and add `Access-Control-Allow-Method: GET` too, to limit to only GET requests?

albig (dismissed):
Approved with remarks.

Classes/Plugin/Eid/PageViewProxy.php
sebastian-meyer: We could restrict access to GET requests, but we don't know the origin.

albig: Don't we? The code will be included in some website on e.g.` newspaper.example.com`. So we know newspaper.example.com as origin and could grant the usage. But maybe, I'm wrong. Just a hint.

sebastian-meyer:
I've limited the `Access-Control-Allow-Methods` to `GET` and set the `Access-Control-Allow-Origin` to the request's `Origin` header. But the latter does not provide any additional security, since it is always set according to the requesting resource.

Classes/Plugin/Eid/PageViewProxy.php
sebastian-meyer: OK, I'll look into that. I hesitated going down that road, because it means having to configure the allowed domain somewhere and getting this information in the eID script.

sebastian-meyer: `Access-Control-Allow-Origin` may include only one origin. So we can't set this correctly for the Zeitungsportal, because some of the Javascript files are hosted on a different host than TYPO3. We would need to allow at least two different origins...

albig: Thank you for digging deeper into it. It's getting complicate if we want to make it really clean. Yes you could make it configurable with all the possible origin domains. The apache-way is described here: https://stackoverflow.com/a/1850482/2378158. It seems to be not recommended to provide a list of domains but to show a single domain if it is in a list of known domains.

But for now, I would keep your solution.

albig (approved):

PR #506, opened 2020-04-23
Make usage of absolute links configurable

sebastian-meyer:
In multi-domain environments like the Zeitungsportal we need the option to force all links to be absolute. Since TYPO3 9.x the Typoscript configuration option `config.absRefPrefix` only effects links generated by Typoscript, but doesn't work globally.

This Pull-Request introduces new extension-wide configuration options to force all generated links to be absolute. In addition it allows using HTTPS as default scheme for those URLs.

The default is to not use absolute links, so existing instances should behave exactly like before.

### DIFF
--- a/Classes/Common/AbstractPlugin.php
+++ b/Classes/Common/AbstractPlugin.php
@@ -217,7 +217,7 @@ abstract class AbstractPlugin extends \TYPO3\CMS\Frontend\Plugin\AbstractPlugin
     /**
      * Link string to the current page.
      * @see \TYPO3\CMS\Frontend\Plugin\AbstractPlugin->pi_linkTP()
-     * 
+     *
      * @access public
      *
      * @param string $str: The content string to wrap in <a> tags
--- a/Classes/Common/AbstractPlugin.php
+++ b/Classes/Common/AbstractPlugin.php
@@ -15,6 +15,7 @@ namespace Kitodo\Dlf\Common;
 use TYPO3\CMS\Core\Database\ConnectionPool;
 use TYPO3\CMS\Core\Service\MarkerBasedTemplateService;
 use TYPO3\CMS\Core\Utility\GeneralUtility;
+use TYPO3\CMS\Core\Utility\HttpUtility;
 
 /**
  * Abstract plugin class for the 'dlf' extension
@@ -88,7 +89,7 @@ abstract class AbstractPlugin extends \TYPO3\CMS\Frontend\Plugin\AbstractPlugin
      *
      * @access protected
      *
-     * @param array $conf: configuration array from TS-Template
+     * @param array $conf: Configuration array from TS-Template
      *
      * @return void
      */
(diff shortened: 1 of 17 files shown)


Classes/Common/AbstractPlugin.php
```diff
@@ -198,40 +199,68 @@ protected function loadDocument()
...
+    {
+        $parser = GeneralUtility::makeInstance(\TYPO3\CMS\Core\TypoScript\Parser\TypoScriptParser::class);
+        $parser->parse($string);
+        return $parser->setup;
+    }
+
+    /**
+     * Link string to the current page.
+     * @see \TYPO3\CMS\Frontend\Plugin\AbstractPlugin->pi_linkTP()
+     *
```
stweil:
```suggestion
     *
```

stweil: This fixes Codacy CI.

sebastian-meyer: Thanks!

albig (dismissed):
Actually this seems to work but keep in mind:
With TYPO9.5: 
* You need a site-configuration with a base including scheme+domain. Without this the links keep to be relative. Providing "/" as base will change nothing with these settings and won't make absolute URLs.
* If you have no HTTPS scheme provided in the site-configuration, it won't be force by this setting.
* Don't know, what is happening with TYPO3 8.7.

As it shouldn't break anything, we can merge this even if it won't work as expected in all situaltions.

sebastian-meyer:
That's how it is intended to work:
- If you have configured a base URL and HTTPS scheme in the `site-configuration`, those do not automatically lead to TYPO3 generating absolute URLs everywhere. The `typolink` function will still generate relative links as long as you don't set the `forceAbsoluteUrl` parameter. So that's where the new extension settings come into play.
- If you haven't configured a base URL and/or HTTPS scheme in the `site-configuration`, you can't expect TYPO3 to generate absolute URLs even with the new settings.

But you are right, I should make it clearer, that having a proper `site-configuration` is required for using the new extension settings.

albig (approved):

Classes/Common/AbstractPlugin.php
```diff
@@ -198,40 +199,68 @@ protected function loadDocument()
...
-            if (!$GLOBALS['TSFE']->config['config']['disablePrefixComment']) {
-                $content = "\n\n<!-- BEGIN: Content of extension '" . $this->extKey . "', plugin '" . Helper::getUnqualifiedClassName(get_class($this)) . "' -->\n\n" . $content . "\n\n<!-- END: Content of extension '" . $this->extKey . "', plugin '" . Helper::getUnqualifiedClassName(get_class($this)) . "' -->\n\n";
-            }
+        $conf = [];
+        if (!$cache) {
+            $conf['no_cache'] = true;
         }
-        return $content;
+        $conf['parameter'] = $altPageId ?: ($this->pi_tmpPageId ?: 'current');
+        $conf['additionalParams'] = $this->conf['parent.']['addParams'] . HttpUtility::buildQueryString($urlParameters, '&', true) . $this->pi_moreParams;
```
albig, line 237: buildQueryString() is unknown in TYPO3 8.7.

sebastian-meyer: Good catch! I'll fix it in a separate pull request.

PR #507, opened 2020-04-23
Add option to put JS in template

sebastian-meyer:
For the Zeitungsportal (or other widget-based integrations) we omit all header and footer rendering. Basically only the plugins' template-based HTML is rendered. For the Page View plugin this means all Javascript added to the page footer getting lost.

This pull request adds the option to include the Javascript into the template-based output instead. However, the default option remains to add it to the page footer.


sebastian-meyer:
Updated according to #509

albig (approved):
Pageview is working for me.
AudioPlayer is not tested.

PR #511, opened 2020-05-07
Use string value of header

sebastian-meyer:
`getHeader()` returns an array of header values, but we need a string representation to set the `Access-Control-Allow-Origin` header correctly. `getHeaderLine()` returns such a string.

### DIFF
--- a/Classes/Plugin/Eid/PageViewProxy.php
+++ b/Classes/Plugin/Eid/PageViewProxy.php
@@ -38,11 +38,11 @@ class PageViewProxy
     public function main(ServerRequestInterface $request)
     {
         // header parameter for getUrl(); allowed values 0,1,2; default 0
-        $header = (int)$request->getQueryParams()['header'];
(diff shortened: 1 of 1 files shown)


albig (dismissed):
This should work. What do you do, if the Origin header does not exist? The result would be 
`Access-Control-Allow-Origin: `. I think this is no problem as this should be treated as not set at all.

sebastian-meyer:
If there is no `Origin` header, it's not a valid CORS request anyway. But you are right, in this case we shouldn't return the `Access-Control-Allow-Origin` header at all, since returning an empty value is not valid either.

sebastian-meyer:
We now return the value of the `Origin` request header if it is set and `*` otherwise. This is the recommended way of handling CORS requests which are allowed under the same-origin policy (see https://developer.mozilla.org/de/docs/Web/HTTP/CORS).

albig (approved):

PR #514, opened 2020-05-11
Remove deprecated parameter

sebastian-meyer:
The `htmlspecialchars` parameter of `sL()`,  `getLL()`, `getLLL()` and `pi_getLL()` is deprecated in 8.x and removed in 9.x. This pull request removes the parameter and explicitly calls `htmlspecialchars()` on the result if necessary.


Classes/Plugin/Navigation.php
```diff
@@ -136,67 +136,67 @@ public function main($content, $conf)
...
-            $markerArray['###NEXT###'] = $this->makeLink($this->pi_getLL('nextPage', '', true), ['page' => $this->piVars['page'] + (1 + $this->piVars['double'])]);
+            $markerArray['###NEXT###'] = $this->makeLink(htmlspecialchars($this->pi_getLL('nextPage', '')), ['page' => $this->piVars['page'] + (1 + $this->piVars['double'])]);
         } else {
-            $markerArray['###NEXT###'] = '<span title="' . $this->pi_getLL('nextPage', '', true) . '">' . $this->pi_getLL('nextPage', '', true) . '</span>';
+            $markerArray['###NEXT###'] = '<span title="' . htmlspecialchars($this->pi_getLL('nextPage', '')) . '">' . htmlspecialchars($this->pi_getLL('nextPage', '')) . '</span>';
         }
         // Link forward X pages.
         if ($this->piVars['page'] <= ($this->doc->numPages - $pageSteps)) {
-            $markerArray['###FORWARD###'] = $this->makeLink(sprintf($this->pi_getLL('forwardXPages', '', true), $pageSteps), ['page' => $this->piVars['page'] + $pageSteps]);
+            $markerArray['###FORWARD###'] = $this->makeLink(htmlspecialchars(sprintf($this->pi_getLL('forwardXPages', '')), $pageSteps), ['page' => $this->piVars['page'] + $pageSteps]);
```
albig: There is wrong closing bracket `)` causing sprintf() to fail. The bracket has to be moved after `$pageSteps`.

albig (changes requested):

Classes/Plugin/Navigation.php
sebastian-meyer: Good catch! I've fixed it with 862b7db

albig (approved):

PR #515, opened 2020-05-13
Feed plugin db fix

erik-konrad:
* refactores the feed plugin for using querybuilder instead of $GLOBALS['TYPO3_DB']
* changes ext_emconf.php for better autoload vendor classes in non composer installation
* php deprecation fix in MetsDocument.php for PHP 7.4

### DIFF
--- a/Classes/Plugin/Feeds.php
+++ b/Classes/Plugin/Feeds.php
@@ -73,7 +73,7 @@ class Feeds extends \Kitodo\Dlf\Common\AbstractPlugin
 
         $allResults = $result->fetchAll();
 
-        if (count($allResults) == 1) {
+        if (count($allResults) === 1) {
             $resArray = $allResults[0];
             $channel->appendChild($rss->createElement('copyright', htmlspecialchars($resArray['label'], ENT_NOQUOTES, 'UTF-8')));
         }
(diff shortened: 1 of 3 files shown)


sebastian-meyer (approved):

Classes/Plugin/Feeds.php
```diff
@@ -88,28 +88,55 @@ public function main($content, $conf)
...
+                    'tx_dlf_documents.author AS author',
+                    'tx_dlf_documents.record_id AS guid',
+                    'tx_dlf_documents.tstamp AS tstamp',
+                    'tx_dlf_documents.crdate AS crdate'
+                )
+                ->from('tx_dlf_documents')
+                ->join(
+                    'tx_dlf_documents',
+                    'tx_dlf_relations',
+                    'tx_dlf_relations',
```
sebastian-meyer: Shouldn't this be something else like `tx_dlf_relations_join`?

Classes/Plugin/Feeds.php
```diff
@@ -88,28 +88,55 @@ public function main($content, $conf)
...
+                    $queryBuilder->expr()->eq('tx_dlf_documents.uid', $queryBuilder->quoteIdentifier('tx_dlf_relations.uid_local'))
+                )
+                ->join(
+                    'tx_dlf_relations',
+                    'tx_dlf_collections',
+                    'tx_dlf_collections',
```
sebastian-meyer, line 120: See above.

Classes/Plugin/Feeds.php
```diff
@@ -88,28 +88,55 @@ public function main($content, $conf)
...
+                    $queryBuilder->expr()->eq('tx_dlf_collections.uid', $queryBuilder->quoteIdentifier('tx_dlf_relations.uid_foreign'))
+                )
+                ->where(
+                    $queryBuilder->expr()->eq('tx_dlf_documents.pid', $queryBuilder->createNamedParameter((int)$this->conf['pages'])),
+                    $queryBuilder->expr()->eq('tx_dlf_relations.ident', $queryBuilder->createNamedParameter('docs_colls')),
```
sebastian-meyer: This would be `tx_dlf_relations_join.ident` then.

Classes/Plugin/Feeds.php
```diff
@@ -88,28 +88,55 @@ public function main($content, $conf)
...
+                    $queryBuilder->expr()->eq('tx_dlf_collections.pid', $queryBuilder->createNamedParameter((int)$this->conf['pages'])),
```
sebastian-meyer, line 126: And this `tx_dlf_collections_join.pid`.

sebastian-meyer (changes requested):
See my comments below. I am not quite sure how Doctrine handles joins, but according to other instances in the code, this seems odd here.
See for example: https://github.com/kitodo/kitodo-presentation/blob/669e9a09518cd677a57d9d4d63b9765e20a47179/Classes/Command/ReindexCommand.php#L219-L265

Classes/Plugin/Feeds.php
erik-konrad: The third parameter is only an identifier. You can call it what you want. If you dont want to use an identifier, you can take the table name. And a table named "tx_dlf_relations_join" does not exist. So I think this identifier is confusing for third party developers.

Classes/Plugin/Feeds.php
erik-konrad: See above. :)

Classes/Plugin/Feeds.php
sebastian-meyer: Isn't this a problem since we are using the identifier later in the `WHERE` clauses? How does Doctrine know we are referring to the joined table instead of the original?

erik-konrad: I can rename it. Thats not the problem. But I think the better identifier is "tx_dlf_documents_collections_mm" because this is TYPO3 standard name convention.

sebastian-meyer (approved):

PR #517, opened 2020-05-15
Make target container for image manipulation control configurable

sebastian-meyer:
The image manipulation tool of the toolbox is rendered into the first element with class `tx-dlf-toolbox` upon activation. Not only does this put the controls always to the bottom of the toolbox, it also always puts them into the first toolbox, which is not necessarily the one you actually added the tool to (if you have multiple instances of toolboxes like we have in the Zeitungsportal).

This pull requests changes this. Now the image manipulation controls get rendered by default into the first element with class `tx-dlf-imagemanipulationtool`, which is the container of the tool itself instead of the whole toolbox.
Moreover the target container is now configurable via Typoscript:
`plugin.tx_dlf_imagemanipulationtool.parentContainer = #elementId` (when using element id)
`plugin.tx_dlf_imagemanipulationtool.parentContainer = .elementClass` (when using element class)

To maintain the old behavior, the following Typoscript would be needed:
`plugin.tx_dlf_imagemanipulationtool.parentContainer = .tx-dlf-toolbox`


albig (approved):
Works for me.

PR #531, opened 2020-07-16
Prepare Release 2.3.0

sebastian-meyer:


composer.json
```diff
@@ -18,7 +18,7 @@
   ],
   "license": "GPL-3.0-or-later",
   "type": "typo3-cms-extension",
-  "version": "2.2.0",
```
albig, line 21: It's recommended to omit the version statement in composer.json files. But it doesn't hurt in this case, I think.

https://getcomposer.org/doc/04-schema.md#version

> Note: Packagist uses VCS repositories, so the statement above is very much true for Packagist as well. Specifying the version yourself will most likely end up creating problems at some point due to human error.

albig (dismissed):

composer.json
sebastian-meyer: You are right. I removed the version statement.

albig (approved):

PR #545, opened 2020-11-20
Fix search in documents for hierarchical documents like newspaper.

albig:
### Current Situation

The search plugin has the option to restrict the search in current
document and/or in current collection. This can be used to place the
plugin on a separate page to search only in one collection. Or you place
the plugin on the workview page and trigger a search in document.

Of course, the SearchInDocument plugin is better suited for the workview as the results are reported by via AJAX and the visitor will stay on the current page.

On the opposite, using the search plugin makes it possible to search
in all issues of a newspaper year or even a full newspaper title. At
least, this should be possible. The current implementation does not work
as expected.

#### Search plugin on workview (issue)

For example, you can place the plugin on level of a newspaper issue. If
you launch a search, the current implementation will search not only in
current issue but in all issues of the current year. This is a
frequently demanded feature but works only by hasard.

<pre>
newspaper title (anchor)
   |
   |-> year
        |
        |-> Issue
        |-> Issue
</pre>

The current search plugin does two mistakes:

1. Instead of the current document uid, the parent uid is set.
2. The search is done in given uid and all children.

#### Search plugin on workview (year/anchor)

If you place the search plugin on the workview with calendar view, the
search fails.

Same procedure as above. The search will be done in the anchor and it's
children (years). All these files have no fulltexts. So the result is
always empty.

## Proposed Implementation

The proposed implementation does change the searchIn document limitation
of the search plugin to search in current document _and_ all it's children.

This is done on using Solr join feature. To make this possible, the
partof field must use the docValues feature as "uid" does it already.


albig:
This is related to https://github.com/slub/slub_digitalcollections/issues/3

Classes/Common/Solr.php
```diff
@@ -324,12 +324,6 @@ public function search()
         $params['filterquery'] = isset($params['filterquery']) ? $params['filterquery'] : [];
         // Restrict the fields to the required ones.
         $params['fields'] = 'uid,id';
-        // Extend filter query to get all documents with the same uids.
```
albig, line 369: Actually, this breaks my changes and I cannot see any usecase for this code.

Classes/Plugin/Search.php
```diff
@@ -117,19 +117,26 @@ protected function addCurrentDocument()
             && \TYPO3\CMS\Core\Utility\MathUtility::canBeInterpretedAsInteger($this->piVars['id'])
         ) {
             $this->loadDocument();
-            // Get document's UID or parent ID.
+            // Get document's UID
             if ($this->doc->ready) {
-                return '<input type="hidden" name="' . $this->prefixId . '[id]" value="' . ($this->doc->parentId > 0 ? $this->doc->parentId : $this->doc->uid) . '" />';
```
albig, line 122: Why should be set the parentId here?

Classes/Common/Solr.php
sebastian-meyer: If I remember correctly, this is necessary to get all hits within a document. Since we split a METS document into several parts (i. e. pages containing fulltext) which are indexed as separate Solr documents, we have to merge these parts again when searching.

Classes/Plugin/Search.php
sebastian-meyer: In the past, users always wanted to search within an intellectual work, not just within an arbitrary part of it. At least that is true for multivolume works and multi-part manuscripts. That's why we always checked for a `parentId` and searched within all associated children (i. e. volumes).

I think, the correct way would be to get the actual document type and check if it is configured as "toplevel". If it is, it's ok to search just within that document and all of its descendants. If it is not, we need to get the ancestor (not just parent) "toplevel" document and search within that and all of its descendants (not just children).

Basically that's what the current implementation did, but without explicitly checking the document type and without respecting more than two hierarchical levels. Since newspapers have more than two hierarchical levels, it's not sufficient anymore to just check for the `parentId`. Instead we have to go up in the hierarchy until we find a "toplevel" ancestor and then perform the search with this "toplevel" ancestor's UID and all descendants.

Resources/Private/Language/Search.xml
```diff
@@ -28,7 +28,7 @@
             <label index="tt_content.pi_flexform.collections">Restrict search to these collections</label>
             <label index="tt_content.pi_flexform.searchIn">Restrict search to current document or collection?</label>
             <label index="tt_content.pi_flexform.searchIn.none">none</label>
-            <label index="tt_content.pi_flexform.searchIn.document">document only</label>
+            <label index="tt_content.pi_flexform.searchIn.document">document and children only</label>
```
sebastian-meyer: We should name this "descendants" instead of "children" to make clear, that it works for more than two hierarchical levels.

Resources/Private/Language/Search.xml
```diff
@@ -71,7 +71,7 @@
             <label index="tt_content.pi_flexform.collections">Suche auf diese Kollektionen einschränken</label>
             <label index="tt_content.pi_flexform.searchIn">Suche auf aktuelle/s Kollektion/Dokument einschränken?</label>
             <label index="tt_content.pi_flexform.searchIn.none">nein</label>
-            <label index="tt_content.pi_flexform.searchIn.document">nur Dokument</label>
+            <label index="tt_content.pi_flexform.searchIn.document">nur Dokument und Kindelemente</label>
```
sebastian-meyer: See above: "Kindelemente" implies that it works just for direct descendants but not for all.

sebastian-meyer (commented):
This is not a full review, but just a few comments to explain former implementations.

Classes/Common/Solr.php
albig: I think, I checked this and it is still working. But I will have another look at.

sebastian-meyer: @frank-ulrich-weber Do you remember why we added this join?

albig: Actually I couldn't find any changes having these settings. This `{!join from=uid to=uid}` is already prepended to the query itself. It seems to be not necessary to add this to the filter query again. In case of the proposed changes, it even destroys the filter query.

You're right. It is necessary to get all hits inside a document. E.g. `q={!join from=uid to=uid}fulltext:(satz)` and `fq=toplevel:true` will find all documents with the search word `satz` in fulltext. Without this join, you will get an empty result as no toplevel element has a fulltext field.

Classes/Plugin/Search.php
albig: I still find this a surprisingly behavior. If I search in a address book of one year of a town (volume), why should I get results of all other years? If I search in a periodical volume, why should I get results of all other volumes of this periodical? That's not what I would expect when I am search "in document".

The proposed way search now in all documents and descendants. What means:

1. multivolume works
  - search in all volumes, when on the multivolume document itself
  - search in one volume, when inside a volume
2. periodicals
  - search in all volumes, when on the top level document with type periodical
  - search in one volume, when inside a volume
3. newspaper/ephemera
  - **search in all issues of a title when on the anchor document**
  - **search in all issues of year when on the year document**
  - search in one issue when on the issue itself

The **bold** ones are possible with this patch. This didn't work on document level before. The search in one newspaper year was possible, when searching in one issue.

Classes/Plugin/Search.php
```diff
@@ -453,7 +459,10 @@ public function main($content, $conf)
                     !empty($this->piVars['id'])
                     && \TYPO3\CMS\Core\Utility\MathUtility::canBeInterpretedAsInteger($this->piVars['id'])
                 ) {
-                    $params['filterquery'][]['query'] = 'uid:(' . $this->piVars['id'] . ') OR partof:(' . $this->piVars['id'] . ')';
+                    // search in document and all descendants (valid for two level hierarchies like newspaper, too)
+                    $params['filterquery'][]['query'] = '_query_:"{!join from=uid to=partof}uid:{!join from=uid to=partof}uid:' . $this->piVars['id'] . '"' .
```
albig, line 463: This looks very ugly in the first moment. The subquery was necessary, because Solr does not process it otherwise. I'm no expert in Solr queries. It's a bit trial an error.

If we know the document type here, we could make this simpler for the corresponding type.

* anchors of newspapers:
`{!join from=uid to=partof}uid:{!join from=uid to=partof}uid:`
* years, multivolume, periodical
` {!join from=uid to=partof}uid:`
* all other documents
`uid:`

But on the other hand: It doesn't hurt to make this `OR` combination.

Of course, this works only up to three levels (anchor/year/issue). But we have no deeper usage at the moment. And for more complicate use cases, the search has to be rewritten anyhow.

Configuration/ApacheSolr/configsets/dlf/conf/schema.xml
```diff
@@ -107,7 +107,7 @@ limitations under the License.
         <!-- Image number where this document starts. -->
         <field name="page" type="int" indexed="false" stored="true" required="true" default="0" />
         <!-- Unique identifier for the parent document in the TYPO3 database. Only if this is a multi-volume work! -->
-        <field name="partof" type="int" indexed="true" stored="true" required="true" default="0" />
+        <field name="partof" type="int" indexed="true" stored="true" required="true" default="0" docValues="true" />
```
albig, line 110: Here I'm not sure about this. This was necessary during development. Now it seems to work even without it. But on the other side, it doesn't hurt and why should we treat it different to the uid field?

sebastian-meyer: `docValues` doesn't add any new features to a field, it's just a different way of storing the field's contents in the index, providing a faster retrieval for certain scenarios like sorting and faceting. Thus it shouldn't make any functional difference if this is enabled. (See https://solr.apache.org/guide/6_6/docvalues.html)

sebastian-meyer (approved):

PR #546, opened 2020-11-23
Display full text without need for additional button click

beatrycze-volk:


albig:
Why should we do this? This will break all current presentations.

beatrycze-volk:
That change was requested by Nationalzeitungsportal (https://jira.deutsche-digitale-bibliothek.de/browse/DDBNEXT-3330). I was also not sure if this change is going to be added to master or maybe it should go to some separated version which will be used only in that projet.

sebastian-meyer:
I'd suggest making this configurable. This could be done either by implementing an additional option or by duplicating the fulltext tool in a separate plugin to always show the fulltext.
I personally prefer the latter, but it's up to you.

beatrycze-volk:
Missing JS import

PR #548, opened 2020-11-26
Fix scrolling to the highlighted full text element

beatrycze-volk:
Times for waiting to scroll and for scrolling itself may need adjustments.

I have used waiting to avoid immediate pull away from the image to the text right after the hover.

### DIFF
--- a/Resources/Public/Javascript/PageView/FulltextControl.js
+++ b/Resources/Public/Javascript/PageView/FulltextControl.js
@@ -226,7 +226,7 @@ var dlfViewerFullTextControl = function(map, image, fulltextUrl) {
 
                         if (targetElem.length > 0 && !targetElem.hasClass('highlight')) {
                             targetElem.addClass('highlight');
-                            $('#tx-dlf-fulltextselection').scrollTo(targetElem, 50);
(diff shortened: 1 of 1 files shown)


Resources/Public/Javascript/PageView/FulltextControl.js
```diff
@@ -226,7 +226,7 @@ var dlfViewerFullTextControl = function(map, image, fulltextUrl) {
 
                         if (targetElem.length > 0 && !targetElem.hasClass('highlight')) {
                             targetElem.addClass('highlight');
-                            $('#tx-dlf-fulltextselection').scrollTo(targetElem, 50);
+                            setTimeout(this.scrollToText, 4000, targetElem);
```
sebastian-meyer: Timeout should be no longer than 1 second.

sebastian-meyer (changes requested):
Could you please reduce the timeout a bit?

Other than that, the pull request is fine. Thank you!

Resources/Public/Javascript/PageView/FulltextControl.js
beatrycze-volk: Done

sebastian-meyer (approved):

PR #549, opened 2020-12-02
Add plugin for download full text as txt file

beatrycze-volk:


beatrycze-volk:
Needs adjustment to work together with plugin for full text display.

PR #550, opened 2020-12-02
Update Solarium to 5.1.6

beatrycze-volk:
The highest possible now update of Solarium. Necessary change (https://solarium.readthedocs.io/en/stable/getting-started/):

> In the past, the V1 API endpoint solr was not added automatically, so most users set it as path on the endpoint. This bug was discovered with the addition of V2 API support. In almost every setup, the path has to be set to / instead of /solr with this release!

> For the same reason it is a must to explicit configure the core or collection.

Since version 5.2 needed is EventDispatcher which in TYPO3 is first available in ver. 10 (https://docs.typo3.org/m/typo3/reference-coreapi/master/en-us/ApiOverview/Hooks/EventDispatcher/Index.html).

### DIFF
--- a/Classes/Common/Solr.php
+++ b/Classes/Common/Solr.php
@@ -578,7 +578,7 @@ class Solr
                     'scheme' => $solrInfo['scheme'],
                     'host' => $solrInfo['host'],
                     'port' => $solrInfo['port'],
-                    'path' => '/' . $solrInfo['path'],
+                    'path' => '/' . $solrInfo['path'] . '/',
                     'core' => $core,
                     'username' => $solrInfo['username'],
                     'password' => $solrInfo['password'],
--- a/Classes/Common/Solr.php
+++ b/Classes/Common/Solr.php
@@ -248,7 +248,7 @@ class Solr
         // Set port if not set.
         $solrInfo['port'] = \TYPO3\CMS\Core\Utility\MathUtility::forceIntegerInRange($conf['solrPort'], 1, 65535, 8983);
         // Append core name to path.
-        $solrInfo['path'] = trim($conf['solrPath'], '/') . '/solr';
+        $solrInfo['path'] = trim($conf['solrPath'], '/');
         // Timeout
         $solrInfo['timeout'] = \TYPO3\CMS\Core\Utility\MathUtility::forceIntegerInRange($conf['solrTimeout'], 1, intval(ini_get('max_execution_time')), 10);
         return $solrInfo;
@@ -276,7 +276,7 @@ class Solr
             $host = $solrInfo['host'];
         }
         // Return entire request URL.
-        return $solrInfo['scheme'] . '://' . $host . ':' . $solrInfo['port'] . '/' . $solrInfo['path'] . '/' . $core;
+        return $solrInfo['scheme'] . '://' . $host . ':' . $solrInfo['port'] . '/' . $solrInfo['path'] . '/solr/' . $core;
     }
 
     /**
@@ -578,7 +578,7 @@ class Solr
                     'scheme' => $solrInfo['scheme'],
                     'host' => $solrInfo['host'],
                     'port' => $solrInfo['port'],
-                    'path' => '/',
+                    'path' => '/' . $solrInfo['path'],
                     'core' => $core,
                     'username' => $solrInfo['username'],
                     'password' => $solrInfo['password'],
--- a/Classes/Common/Solr.php
+++ b/Classes/Common/Solr.php
@@ -274,7 +274,7 @@ class Solr
             $host = $solrInfo['host'];
         }
         // Return entire request URL.
-        return $solrInfo['scheme'] . '://' . $host . ':' . $solrInfo['port'] . '/solr/' . $core;
+        return $solrInfo['scheme'] . '://' . $host . ':' . $solrInfo['port'] . '/' . $solrInfo['path'] . '/' . $core;
     }
 
     /**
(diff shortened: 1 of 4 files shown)


Classes/Common/Solr.php
```diff
@@ -276,7 +274,7 @@ public static function getSolrUrl($core = '')
             $host = $solrInfo['host'];
         }
         // Return entire request URL.
-        return $solrInfo['scheme'] . '://' . $host . ':' . $solrInfo['port'] . '/' . $solrInfo['path'] . '/' . $core;
+        return $solrInfo['scheme'] . '://' . $host . ':' . $solrInfo['port'] . '/solr/' . $core;
```
sebastian-meyer: See next comment. This should stay unchanged.

```suggestion
        return $solrInfo['scheme'] . '://' . $host . ':' . $solrInfo['port'] . '/' . $solrInfo['path'] . '/' . $core;
```

Classes/Common/Solr.php
```diff
@@ -247,8 +247,6 @@ public static function getSolrConnectionInfo()
         $solrInfo['password'] = $conf['solrPass'];
         // Set port if not set.
         $solrInfo['port'] = \TYPO3\CMS\Core\Utility\MathUtility::forceIntegerInRange($conf['solrPort'], 1, 65535, 8983);
-        // Append core name to path.
-        $solrInfo['path'] = trim($conf['solrPath'], '/');
```
sebastian-meyer, line 251: We still need `$solrInfo['path']`. As far as I understand the changes in Solarium, just the API path segment `/solr` is added automatically by Solarium, but if the Solr server runs in a sub-directory of the web server's document root, we still have to provide this path (although in most cases it will be `/`). So just the `solr`-part is removed from `$conf['solrPath']` and has to be added here (because this bypasses Solarium).

`$solrInfo['path'] = trim($conf['solrPath'], '/') . '/solr';`

Classes/Common/Solr.php
```diff
@@ -578,8 +576,10 @@ protected function __construct($core)
                     'scheme' => $solrInfo['scheme'],
                     'host' => $solrInfo['host'],
                     'port' => $solrInfo['port'],
-                    'path' => '/' . $solrInfo['path'] . '/',
+                    'path' => '/',
                     'core' => $core,
+                    // For Solr Cloud you need to provide a collection instead of core:
+                    // 'collection' => $core
```
sebastian-meyer: That's good to know and a nice addition, but currently we don't support Solr Cloud. If we do, we have to change the way we automatically create new cores as well.

Resources/Private/Language/Labels.xml
```diff
@@ -192,7 +192,7 @@
             <label index="config.solrHttps">Use https: (default is "FALSE")</label>
             <label index="config.solrHost">Solr Server Host: (default is "localhost")</label>
             <label index="config.solrPort">Solr Server Port: (default is "8983")</label>
-            <label index="config.solrPath">Solr Server Path: (default is "/solr/")</label>
+            <label index="config.solrPath">Solr Server Path: (default is "/")</label>
```
sebastian-meyer:
```suggestion
            <label index="config.solrPath">Solr Server Path: without API endpoint "/solr" (default is "/")</label>
```

Resources/Private/Language/Labels.xml
```diff
@@ -375,7 +375,7 @@
             <label index="config.solrHttps">Https verwenden: (Standard ist "FALSE")</label>
             <label index="config.solrHost">Solr Server Host: (Standard ist "localhost")</label>
             <label index="config.solrPort">Solr Server Port: (Standard ist "8983")</label>
-            <label index="config.solrPath">Solr Server Pfad: (Standard ist "/solr/")</label>
+            <label index="config.solrPath">Solr Server Pfad: (Standard ist "/")</label>
```
sebastian-meyer:
```suggestion
            <label index="config.solrPath">Solr Server Pfad: ohne API-Endpunkt "/solr" (Standard ist "/")</label>
```

sebastian-meyer (changes requested):
Please see my comments below.

sebastian-meyer (approved):
Sorry, I confused the constructor (which actually uses Solarium to establish the Solr connection) and `getSolrUrl()` (which is used to create new cores without using Solarium). Some of my review comments were garbage... :o/

I've added a few changes and now it works for me. Does it for you as well?

Resources/Private/Language/Labels.xml
```diff
@@ -192,7 +192,7 @@
             <label index="config.solrHttps">Use https: (default is "FALSE")</label>
             <label index="config.solrHost">Solr Server Host: (default is "localhost")</label>
             <label index="config.solrPort">Solr Server Port: (default is "8983")</label>
-            <label index="config.solrPath">Solr Server Path: (default is "/solr/")</label>
+            <label index="config.solrPath">Solr Server Path: without API endpoint "/solr" (default is "/")</label>
```
beatrycze-volk, line 195: I'm not getting this comment. When it is /solr?

sebastian-meyer: If your Solr runs on `localhost:8983/some-path/solr` in the past, you had to set `config.solrPath` to `/some-path/solr`.
Since newer versions of Solarium add the `/solr` part automatically (because it is considered to be the API endpoint, but not part of the application's path), you have to omit this part and just set `config.solrPath` to `/some-path`.
I added the comment to (hopefully) make this clear - especially for users who are used to the old way...

beatrycze-volk:
Yes, it looks ok.

PR #551, opened 2020-12-03
Duplicate full text plugin for always active full text

beatrycze-volk:
PR #546 was closed because of missing JS file import in PageView. Here change between plugins happens by replacing it.


beatrycze-volk:
Needs adjustment to work together with plugin for full text download.

PR #553, opened 2020-12-06
Make index_name editable

### LINKED ISSUE #554: Error in DataHandler
## Description

There is a serious bug in `Classes/Hooks/DataHandler.php` when handling the `index_name` on record updates. Instead of keeping the `index_name` unchanged by overwriting a given `index_name` in the `$fieldArray` with the current value from database as intended, the database query actually is a copy of the routine handling the `index_autocomplete` and `index_indexed` fields (which sets these fields even though they are not allowed by TCA). Thus it is now possible to change the `index_name` in production and run into `DataHandler` exceptions...

## Screenshots and Examples

https://github.com/kitodo/kitodo-presentation/blob/a12df1d4e195fa2fdb387d831fbabfa2135c088f/Classes/Hooks/DataHandler.php#L209-L222

## Environment

 - OS version: any
 - RDBMS version: any
 - Apache Solr version: N/A
 - TYPO3 version: any
 - PHP version: any

## Additional Context

See pull request #553

sebastian-meyer:
There is a serious bug in `Classes/Hooks/DataHandler.php` when handling the `index_name` on record updates. Instead of keeping the `index_name` unchanged by overwriting a given `index_name` in the `$fieldArray` with the current value from database as intended, the database query actually is a copy of the routine handling the `index_autocomplete` and `index_indexed` fields (which sets these fields even though they are not allowed by TCA). Thus it is now possible to change the `index_name` in production and run into `DataHandler` exceptions...

https://github.com/kitodo/kitodo-presentation/blob/a12df1d4e195fa2fdb387d831fbabfa2135c088f/Classes/Hooks/DataHandler.php#L209-L222

BUT: In theory editing the `index_name` actually is a nice feature - if you know what you are doing and can handle the side effects. So, instead of just fixing the bug, I "repaired" the feature and made it behave more consistently. What this pull request does:

- It makes the `index_name` field for tables `tx_dlf_collections`, `tx_dlf_libraries`, `tx_dlf_metadata` and `tx_dlf_structures` editable.
- Since this is potentially dangerous to do in production, the field itself is configured as `readOnly` in TCA and not shown in the backend.
- Instead there is an additional field `index_name_edit` in TCA, which has to be user-activated explicitly before its value can be changed (thus hopefully preventing accidental changes). It shows the current `index_name` as default value and overwrites the `index_name` field if set (via hook `processDatamap_preProcessFieldArray`).
- The above mentioned flawed code in `Classes/Hooks/DataHandler.php` was removed.


Configuration/TCA/tx_dlf_collections.php
```diff
@@ -120,11 +120,26 @@
...
+            'exclude' => 1,
+            'l10n_mode' => 'exclude',
+            'label' => 'LLL:EXT:dlf/Resources/Private/Language/Labels.xml:tx_dlf_collections.index_name',
+            'config' => [
+                'type' => 'input',
+                'size' => 30,
+                'max' => 255,
+                'eval' => 'null',
+                'default' => null,
+                'mode' => 'useOrOverridePlaceholder',
```
albig: That's a good idea to use a virtual field with `mode=useOrOverridePlaceholder`. 

What I don't like at this feature is the empty default value if you activate the input field. Before you see the placeholder - but shortend. If you just want to change a typo or whatever you have to get somewhere (but from where?) the full index_name. index_name might be quite complexe in case of collections.

That's a missing feature in TYPO3 - not here. The same is with images metatags where this mode is used, too. Editors must get crazy, if they really fill out title, alternative, caption.

Maybe you have an idea for a simple display of the current `index_name` above, to copy into the `index_name_edit` field?

albig (changes requested):
Should work but some essential detail is missing. 

It's not possible to create new records in the tables with index_name_edit!

> 2: SQL error: 'Unknown column 'index_name_edit' in 'field list'' (tx_dlf_structures:NEW5fec3ff075098859175500)

Or if you try to translate a record:

> An exception occurred while executing 'SELECT `uid`, `toplevel`, `index_name`, `index_name_edit`, `oai_name`, `thumbnail`, `pid` FROM `tx_dlf_structures` WHERE (`uid` = ?) AND (`tx_dlf_structures`.`deleted` = 0)' with params [281]: Unknown column 'index_name_edit' in 'field list'

Maybe - haven't testet it now - the solution is to unset `$fieldArray['index_name_edit']` in `processDatamap_postProcessFieldArray()`, too. A (quite old - maybe outdated) note about this is found in slub_events: https://github.com/slub/slub_events/blob/master/Classes/Slots/HookPostProcessing.php#L101

sebastian-meyer:
Unfortunately this doesn't work as I intended. Although I could `unset()` the field `index_name_edit` in `processDatamap_postProcessFieldArray()` as well, this wouldn't fix the problem with translation. In fact, the only backend field type, which does not need a corresponding database field is `type = 'none'`, but this can't be configured to be editable.

I'll revert my changes, make the original fields editable and try to find an option to display some sort of warning message for changing the value.

sebastian-meyer:
OK, another approach: We can't use `fieldInformation`, although logically that would be the right solution. But `fieldInformation` for some reason doesn't allow all the HTML tags coming with Flash Messages - while `fieldWizard` does. This is slightly annoying as `fieldInformation` is shown above the input field (which would've been nicer), while `fieldWizard` is shown below.

Configuration/TCA/tx_dlf_collections.php
sebastian-meyer: With my new approach this isn't an issue anymore as you can now edit the original field which always shows its initial value.

albig (approved):
Seems to work.

The `fieldWizard` content is very obvious. You're right, it would be nicer above. But you're right, `fieldInformation` is very limited. The only supported HTML tags are: "`<a><br><br/><div><em><i><p><strong><span><code>`". The [documentation](https://docs.typo3.org/m/typo3/reference-coreapi/master/en-us/ApiOverview/FormEngine/Rendering/Index.html#formengine-rendering-nodeexpansion) has an example for the usage:

> “Hey, this field was auto-filled for you by an automatic translation wizard. Maybe you want to check the content”

But the orange flash message is better here, I think.

PR #563, opened 2020-12-09
Add full text download tool and duplicate full text display for always active display

beatrycze-volk:
PR contains new plugin for download full text and changes to full text plugin for always active display.

Changes are possible to test on the test system.


sebastian-meyer:
On second thought: Wouldn't it be better to not duplicate the fulltext control tool, but instead add a configuration option to turn it on or off when loading? Something like:

If you set the Typoscript configuration `plugin.tx_dlf_fulltexttool.launchState = on` then the `FulltextControl` tool is turned on on initialization, otherwise (or when setting `plugin.tx_dlf_fulltexttool.launchState = off`) it isn't.

`plugin.tx_dlf_fulltexttool.launchState = off` would be the default and makes the fulltext tool behave just like it does now.

`plugin.tx_dlf_fulltexttool.launchState = on` would turn the fulltext on, but still provide a switch to turn it off.

To meet the requirements of the Zeitungsportal, we would set `plugin.tx_dlf_fulltexttool.launchState = off` and hide the off-switch via CSS.

What do you think?

beatrycze-volk:
Adding the switch is more clean solution than duplicating the code. I will change it.

sebastian-meyer (approved):
Looks good!

Configuration/TypoScript/Toolbox/setup.txt
```diff
@@ -1,5 +1,7 @@
 plugin.tx_dlf_annotationtool.templateFile = EXT:dlf/Resources/Private/Templates/AnnotationTool.tmpl
+plugin.tx_dlf_fulltexttool.launchState = off
```
albig, line 2: Sorry, I cannot reproduce what this setting does:

* Without launchState: nothing happens. Ok.
* with launchState=off: the fulltext is shown initially and on every click on the button it is marked once more. So the color gets darker and darker (in the screenshot its getting more and more red)
* with launchState=on: the old behaviour shows up

So what is launchState=off good for?
fulltext-launchstate

PR #567, opened 2020-12-14
Don't add second slash to solr path if it was default /

### LINKED ISSUE #566: Bugs introduced with usage of Solarium 5.2.0 by #550
## Description

The change by #550 to Solarium 5.2.0 does not work:

1. Feedback in extension settings wrong
Sachsen Digital  TYPO3 CMS 9 5 23
The solrPath should be set to "/" but the flash feedback reports that the solr connection fails. This is wrong.

2. Collection Plugin cannot connect
The collection plugin fails with the famous error message `Call to a member function search_raw() on null`. As always this means there is no valid solr instance.

3. Indexing does'nt work
Does nothing without feedback. Normal behaviour if Solr ist not reachable.

4. New Solr cores cannot be created in backend
Does nothing without feedback. Normal behaviour if Solr ist not reachable.

## Reproduction

Try the steps above.

## Environment

 - Apache Solr version: 7.7.3
 - Solarium 5.2.0
 - TYPO3 version: 9.5.0

## Additional Context

Fix for 2., 3., and 4. is in `Kitodo\Dlf\Common\Solr::__construct()`. The leading and trailing slash around path is wrong. The` $this->service->ping($ping);` fails with it.

beatrycze-volk:
Fix for #566

### DIFF
--- a/Classes/Common/Solr.php
+++ b/Classes/Common/Solr.php
@@ -578,7 +578,7 @@ class Solr
                     'scheme' => $solrInfo['scheme'],
                     'host' => $solrInfo['host'],
                     'port' => $solrInfo['port'],
-                    'path' => '/' . $solrInfo['path'] . '/',
+                    'path' => $solrInfo['path'] !== '' ? '/' . $solrInfo['path'] . '/' : '/',
                     'core' => $core,
                     'username' => $solrInfo['username'],
                     'password' => $solrInfo['password'],


sebastian-meyer:
To fix issue number 1 in #566 we need to change `Classes/Hooks/ConfigurationForm.php` around line 60:

```
        if (!empty($solrInfo['path'])) {
            $solrInfo['path'] .= '/';
        }
        $url = $solrInfo['scheme'] . '://' . $host . ':' . $solrInfo['port'] . '/' . $solrInfo['path'] . 'solr/admin/cores?wt=xml';
```

albig:
This only fixes issue 2. and 3. 

Creating Solr cores is still not possible and fails.

sebastian-meyer:
I suggest changing `Solr::getSolrConnectionInfo()` in a way, that `$solrInfo['path']` always has a trailing `/`:

`$solrInfo['path'] = trim($conf['solrPath'], '/') . '/';`

This requires changing (almost) every other method using this information, but we wouldn't need to distinguish between having a "real" path or just `/` to avoid double slashes. I'll close this pull request and open a new one with the proposed changes.

albig:
Creating Solr-Core failes in `Kitodo\Dlf\Hooks\DataHandler`.` $solrInfo['path']` is empty here and the url probably not valid.

https://github.com/kitodo/kitodo-presentation/blob/master/Classes/Hooks/DataHandler.php#L125

Issue #454, opened 2020-01-08
[FUND] Solr Improvements


sebastian-meyer:
### Solr Improvements
- [x] Initialize/update Solr schema via REST API (instead of providing a schema file which has to be deployed manually) (#1369)
- [x] Update Solr configuration `solrconfig.xml` to be compatible with newest version (#1122)
- [x] Revise Solr configuration `solrconfig.xml` for better performance and less deviation from default configuration (#1122)
- [x] Use default suggester component instead of dedicated `select` request handler for autocompletion (#1289)
- [x] Support for bulk importing, soft commits, etc. (for generally better performance) (#1285)
- [x] Use newest version of https://dbmdz.github.io/solr-ocrhighlighting/ for fulltext search and highlighting (after upgrading to Solr 9.x) (#1122)
- [x] Replace index-time field boosting with query-time boosting (#396)
- [x] Move *optimize* from indexing to TYPO3 scheduler task (#1285)
- [x] Add manual scheduler tasks for (re-)indexing of documents and/or collections (#1119)

csidirop:
I believe https://github.com/kitodo/kitodo-presentation/pull/892 addressed point 2 _Update Solr configuration_ and point 6 _Newest version of ocr-ocrhighlighting_

sebastian-meyer:
You are right about point 6! I forgot to check the mark above.

But point 2 is still valid, I believe. `solrconfig.xml` states version `7.4.0` while the most current version of Solr is `9.1.1`. A diff between the default `solrconfig.xml` from Solr 9.1.1 and our `solrconfig.xml` shows some differences which should be examined in order to maintain compatibility with the newest version.

sebastian-meyer:
Even point 6 gets valid again, when we actually upgrade to Solr 9.x. Currently we are delivering the OCR highlighting plugin for Solr 7/8, so we also need to change that...

michaelkubina:
Do you mean by _"Support for bulk importing, **soft commits**, etc. (for generally better performance)"_ something like the atomic update mechanism? (https://solr.apache.org/guide/solr/latest/indexing-guide/partial-document-updates.html#atomic-updates)

If not, it would be something worth considering for your list. Updating only those (for the most part dynamic) metadata fields that actually have been changed in the backend, instead of reindexing the whole documents including the fulltexts, would give an huge performance boost.

[...]

PR #569, opened 2020-12-20
Finish Solarium upgrade

### LINKED ISSUE #566: Bugs introduced with usage of Solarium 5.2.0 by #550

sebastian-meyer:
This pull requests finishes the Solarium upgrade from #550:

Use Solarium everywhere we communicate with Solr:
- [x] Fetching documents via `\Kitodo\Dlf\Common\DocumentList`
- [x] Indexing documents via `\Kitodo\Dlf\Common\Indexer`
- [x] Checking Solr connection via `\Kitodo\Dlf\Hooks\ConfigurationForm`
- [x] Adding new **or deleting** Solr cores via `\Kitodo\Dlf\Hooks\DataHandler`
- [x] Fetching documents via `\Kitodo\Dlf\Plugin\Collection`
- [x] Search queries via `\Kitodo\Dlf\Plugin\OaiPmh`
- [x] Search queries via `\Kitodo\Dlf\Plugin\Search`
- [x] Search queries via `\Kitodo\Dlf\Plugin\Eid\SearchInDocument`
- [x] Autocomplete queries via `\Kitodo\Dlf\Plugin\Eid\SearchSuggest`
- [x] Checking for Solr updates and recreating cores via `class.ext_update.php`

This should fix #566

### DIFF
--- a/Classes/Hooks/DataHandler.php
+++ b/Classes/Hooks/DataHandler.php
@@ -287,10 +287,10 @@ class DataHandler
             in_array($command, ['move', 'delete', 'undelete'])
             && $table == 'tx_dlf_documents'
         ) {
-            // Get Solr-Core.
+            // Get Solr core.
             $queryBuilder = GeneralUtility::makeInstance(ConnectionPool::class)
                 ->getQueryBuilderForTable('tx_dlf_solrcores');
-            // tx_dlf_documents is already deleted at this point --> include deleted documents in query
+            // Record in "tx_dlf_documents" is already deleted at this point.
             $queryBuilder
                 ->getRestrictions()
                 ->removeByType(DeletedRestriction::class);
@@ -347,6 +347,54 @@ class DataHandler
                         }
                         break;
                 }
+            }
+        }
+        if (
+            $command === 'delete'
+            && $table == 'tx_dlf_solrcores'
+        ) {
+            // Delete core from Apache Solr as well.
+            $queryBuilder = GeneralUtility::makeInstance(ConnectionPool::class)
+                ->getQueryBuilderForTable('tx_dlf_solrcores');
+            // Record in "tx_dlf_solrcores" is already deleted at this point.
+            $queryBuilder
+                ->getRestrictions()
+                ->removeByType(DeletedRestriction::class);
+
+            $result = $queryBuilder
+                ->select(
+                    'tx_dlf_solrcores.index_name AS core'
+                )
+                ->from('tx_dlf_solrcores')
+                ->where($queryBuilder->expr()->eq('tx_dlf_solrcores.uid', intval($id)))
+                ->setMaxResults(1)
+                ->execute();
+
+            $allResults = $result->fetchAll();
+
+            if (count($allResults) == 1) {
+                $resArray = $allResults[0];
+                // Establish Solr connection.
+                $solr = Solr::getInstance();
+                if ($solr->ready) {
+                    // Delete Solr core.
+                    $query = $solr->service->createCoreAdmin();
+                    $action = $query->createUnload();
+                    $action->setCore($resArray['core']);
+                    $action->setDeleteDataDir(true);
+                    $action->setDeleteIndex(true);
+                    $action->setDeleteInstanceDir(true);
+                    $query->setAction($action);
+                    try {
+                        $response = $solr->service->coreAdmin($query);
+                        if ($response->getWasSuccessful()) {
+                            return;
+                        }
+                    } catch (\Exception $e) {
+                        // Nothing to do here.
+                    }
+                }
+                Helper::devLog('Core ' . $resArray['core'] . ' could not be deleted from Apache Solr', DEVLOG_SEVERITY_WARNING);
             }
         }
     }
--- a/Classes/Hooks/DataHandler.php
+++ b/Classes/Hooks/DataHandler.php
@@ -246,7 +246,8 @@ class DataHandler
                             $resArray = $allResults[0];
                             if ($resArray['hidden']) {
                                 // Establish Solr connection.
-                                if ($solr = Solr::getInstance($resArray['core'])) {
+                                $solr = Solr::getInstance($resArray['core']);
+                                if ($solr->ready) {
                                     // Delete Solr document.
                                     $updateQuery = $solr->service->createUpdate();
                                     $updateQuery->addDeleteQuery('uid:' . $id);
@@ -325,7 +326,8 @@ class DataHandler
                     case 'move':
                     case 'delete':
                         // Establish Solr connection.
-                        if ($solr = Solr::getInstance($resArray['core'])) {
+                        $solr = Solr::getInstance($resArray['core']);
+                        if ($solr->ready) {
                             // Delete Solr document.
                             $updateQuery = $solr->service->createUpdate();
                             $updateQuery->addDeleteQuery('uid:' . $id);
--- a/Classes/Hooks/DataHandler.php
+++ b/Classes/Hooks/DataHandler.php
@@ -91,53 +91,13 @@ class DataHandler
                     break;
                     // Field post-processing for table "tx_dlf_solrcores".
                 case 'tx_dlf_solrcores':
-                    $queryBuilder = GeneralUtility::makeInstance(ConnectionPool::class)
-                        ->getQueryBuilderForTable('tx_dlf_solrcores');
-
-                    // Get number of existing cores.
-                    $result = $queryBuilder
-                        ->select('*')
-                        ->from('tx_dlf_solrcores')
-                        ->execute();
-
-                    // Get first unused core number.
-                    $coreNumber = Solr::solrGetCoreNumber(count($result->fetchAll()));
-                    // Get Solr credentials.
-                    $conf = unserialize($GLOBALS['TYPO3_CONF_VARS']['EXT']['extConf']['dlf']);
-                    $solrInfo = Solr::getSolrConnectionInfo();
-                    // Prepend username and password to hostname.
-                    if (
-                        $solrInfo['username']
-                        && $solrInfo['password']
-                    ) {
-                        $host = $solrInfo['username'] . ':' . $solrInfo['password'] . '@' . $solrInfo['host'];
-                    } else {
-                        $host = $solrInfo['host'];
+                    // Create new Solr core.
+                    $fieldArray['index_name'] = Solr::createCore();
+                    if (empty($fieldArray['index_name'])) {
+                        Helper::devLog('Could not create new Apache Solr core', DEVLOG_SEVERITY_ERROR);
+                        // Solr core could not be created, thus unset field array.
+                        $fieldArray = [];
                     }
-                    $context = stream_context_create([
-                        'http' => [
-                            'method' => 'GET',
-                            'user_agent' => ($conf['useragent'] ? $conf['useragent'] : ini_get('user_agent'))
-                        ]
-                    ]);
-                    // Build request for adding new Solr core.
-                    // @see http://wiki.apache.org/solr/CoreAdmin
-                    $url = $solrInfo['scheme'] . '://' . $host . ':' . $solrInfo['port'] . '/' . $solrInfo['path'] . '/admin/cores?wt=xml&action=CREATE&name=dlfCore' . $coreNumber . '&instanceDir=dlfCore' . $coreNumber . '&dataDir=data&configSet=dlf';
-                    $response = @simplexml_load_string(file_get_contents($url, false, $context));
-                    // Process response.
-                    if ($response) {
-                        $solrStatus = $response->xpath('//lst[@name="responseHeader"]/int[@name="status"]');
-                        if (
-                            is_array($solrStatus)
-                            && $solrStatus[0] == 0
-                        ) {
-                            $fieldArray['index_name'] = 'dlfCore' . $coreNumber;
-                            return;
-                        }
-                    }
-                    Helper::devLog('Could not create new Apache Solr core "dlfCore' . $coreNumber . '"', DEVLOG_SEVERITY_ERROR);
-                    // Solr core could not be created, thus unset field array.
-                    $fieldArray = [];
                     break;
             }
         } elseif ($status == 'update') {
--- a/Classes/Hooks/DataHandler.php
+++ b/Classes/Hooks/DataHandler.php
@@ -353,48 +353,52 @@ class DataHandler
             $command === 'delete'
             && $table == 'tx_dlf_solrcores'
         ) {
-            // Delete core from Apache Solr as well.
-            $queryBuilder = GeneralUtility::makeInstance(ConnectionPool::class)
-                ->getQueryBuilderForTable('tx_dlf_solrcores');
-            // Record in "tx_dlf_solrcores" is already deleted at this point.
-            $queryBuilder
-                ->getRestrictions()
-                ->removeByType(DeletedRestriction::class);
+            // Is core deletion allowed in extension configuration?
+            $extConf = unserialize($GLOBALS['TYPO3_CONF_VARS']['EXT']['extConf']['dlf']);
+            if (!empty($extConf['solrAllowCoreDelete'])) {
+                // Delete core from Apache Solr as well.
+                $queryBuilder = GeneralUtility::makeInstance(ConnectionPool::class)
+                    ->getQueryBuilderForTable('tx_dlf_solrcores');
+                // Record in "tx_dlf_solrcores" is already deleted at this point.
+                $queryBuilder
+                    ->getRestrictions()
+                    ->removeByType(DeletedRestriction::class);
 
-            $result = $queryBuilder
-                ->select(
-                    'tx_dlf_solrcores.index_name AS core'
-                )
-                ->from('tx_dlf_solrcores')
-                ->where($queryBuilder->expr()->eq('tx_dlf_solrcores.uid', intval($id)))
-                ->setMaxResults(1)
-                ->execute();
+                $result = $queryBuilder
+                    ->select(
+                        'tx_dlf_solrcores.index_name AS core'
+                    )
+                    ->from('tx_dlf_solrcores')
+                    ->where($queryBuilder->expr()->eq('tx_dlf_solrcores.uid', intval($id)))
+                    ->setMaxResults(1)
+                    ->execute();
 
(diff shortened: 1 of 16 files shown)


sebastian-meyer:
~~Even though the first commit message suggests otherwise, this doesn't upgrade Solarium to 5.2, but keeps the version introduced by @beatrycze-volk.~~
I actually updated the Solarium dependency to `^5.2` with f16209708e0ee4d2dfc0d197751fbe4aae6ed53f.

beatrycze-volk (dismissed):

Classes/Plugin/Eid/SearchInDocument.php
```diff
@@ -38,29 +38,48 @@ class SearchInDocument
...
+        // Perform Solr query.
+        $solr = Solr::getInstance($core);
+        if ($solr->ready) {
+            $query = $solr->service->createSelect();
+            $query->setFields(['id', 'uid', 'page']);
+            $query->setQuery('fulltext:(' . Solr::escapeQuery((string) $parameters['q']) . ') AND uid:' . intval($parameters['uid']));
+            $query->setStart($count)->setRows(20);
+            $hl = $query->getHighlighting();
+            $hl->setFields('fulltext');
+            $hl->setMethod('fastVector');
```
albig: This method exists only from solarium 6.0.2.
Maybe, before it was possible with `$hl->setUsefastvectorhighlighter(true);`? Could'n test this because of non-ready solr above.

Classes/Plugin/Eid/SearchInDocument.php
```diff
@@ -38,29 +38,48 @@ class SearchInDocument
...
-            $query = (string)$parameters['q'];
-            $uid = (string)$parameters['uid'];
-            $start = (string)$parameters['start'];
-            $url = trim(Solr::getSolrUrl($core), '/') . '/select?wt=json&q=fulltext:(' . Solr::escapeQuery($query) . ')%20AND%20uid:' . $uid
-                . '&hl=on&hl.fl=fulltext&fl=uid,id,page&hl.method=fastVector'
-                . '&start=' . $start . '&rows=20';
-            $output = GeneralUtility::getUrl($url);
+        // Perform Solr query.
+        $solr = Solr::getInstance($core);
+        if ($solr->ready) {
```
albig, line 55: solr instance is not ready at this point.

$solr->config->path is empty. Should be "/".
$solr->config->timeout is 0. Should be 10.
Both is not set in the extension configuration.

Classes/Plugin/Eid/SearchInDocument.php
```diff
@@ -38,29 +38,48 @@ class SearchInDocument
...
+            $results = $solr->service->select($query);
+            $output['numFound'] = $results->getNumFound();
+            $highlighting = $results->getHighlighting();
+            foreach ($results as $result) {
+                $snippet = $highlighting->getResult($result->id);
+                $document = [
+                    'id' => $result->id,
+                    'uid' => $result->uid,
+                    'page' => $result->page,
+                    'snippet' => $snippet['fulltext']
```
albig: Cannot use object of type Solarium\Component\Result\Highlighting\Result as array

Classes/Hooks/DataHandler.php
```diff
@@ -387,5 +349,53 @@ public function processCmdmap_postProcess($command, $table, $id)
...
+            if (count($allResults) == 1) {
+                $resArray = $allResults[0];
+                // Establish Solr connection.
+                $solr = Solr::getInstance();
+                if ($solr->ready) {
+                    // Delete Solr core.
+                    $query = $solr->service->createCoreAdmin();
+                    $action = $query->createUnload();
+                    $action->setCore($resArray['core']);
+                    $action->setDeleteDataDir(true);
```
albig: Honeystly. I wouldn't add this feature. This can destroy the indexing work of several months by a wrong click in the backend. Or you move installation to another system and do not think about chaning the Solr configuration.
Deleting a solr core is a super admin job. So the person knows what he/she is doing and can do this even better in the Solr admin page.
If we add this feature to TYPO3, we need at least one of the following:
1. Possibility to disable this feature
2. Better do it in some backend-module
3. Add a dialog to confirm the deletion

albig (changes requested):
The status of the solr connection seems to work in the extentension configuration. The searchInDocument-plugin fails. If this is fixed, I would check the autosuggestion and the other features.
The feature "delete core" is nice but very dangerious. I wouldn't implement it like this.

Classes/Plugin/Eid/SearchInDocument.php
sebastian-meyer: You're right! The method is called `setUseFastVectorHighlighter()` in Solarium <6.x.

Classes/Plugin/Eid/SearchInDocument.php
sebastian-meyer: Ah, `Solarium\Component\Result\Highlighting\Result` implements `\IteratorAggregate` and `\Countable`, but not `\ArrayAccess` for some reason...

Classes/Hooks/DataHandler.php
sebastian-meyer: I've added the first option: In order to automatically delete a Solr core when deleting the corresponding TYPO3 record, you have to set `solrAllowCoreDelete` to `true` in extension configuration (default is `false` to prevent accidents).

Classes/Plugin/Eid/SearchInDocument.php
sebastian-meyer: > $solr->config->path is empty. Should be "/".

This is actually correct. A single `/` is added in `Solr::__construct()` at line 640.

> $solr->config->timeout is 0. Should be 10.

Hm, that was a tricky one. In line 333 of `/Classes/Common/Solr.php` the timeout is set as:
`MathUtility::forceIntegerInRange($conf['solrTimeout'], 1, intval(ini_get('max_execution_time')), 10)`

So, if `$conf['timeout']` is not set in the extension configuration, a default value of `10` is set, but only if this value is between `1` and `max_execution_time`.

The tricky part is, that if you are using something like XDebug or a development environment which has `max_execution_time` disabled, `ini_get('max_execution_time')` will return `0`, thus forcing the timeout to be `0` as well.

I've changed this so that the timeout is limited to 30 seconds if `max_execution_time` is 0.

Resources/Public/Javascript/Search/SearchInDocument.js
```diff
@@ -67,55 +64,61 @@ $(document).ready(function() {
...
-              var resultItems = [];
-              var resultList = '<div class="results-active-indicator"></div><ul>';
-              if (data.error) {
-                  resultList += '<li class="error">' + data.error + '</li>';
-              } else {
-                  for (var i=0; i < data.response.docs.length; i++) {
+                var resultItems = [];
+                var resultList = '<div class="results-active-indicator"></div><ul>';
+                var start = -1;
+                var results = JSON.parse(data);
```
albig:
```suggestion
                var results = data;
```

albig: `JSON.parse(data)` fails with the following message: _SyntaxError: Unexpected token o in JSON at position 1_.

It's because `data` is already a JavaScript object and no JSON at this point anymore.

See: https://stackoverflow.com/questions/38380462/syntaxerror-unexpected-token-o-in-json-at-position-1

Resources/Public/Javascript/Search/SearchInDocument.js
```diff
@@ -67,55 +64,61 @@ $(document).ready(function() {
...
-                  }
-                  if (resultItems.length > 0) {
-                    // sort by page as this cannot be done with current solr schema
-                    resultItems.sort(function(a, b){return a-b});
-                    resultItems.forEach(function(item, index){
+                        if (element['snippet'].length > 0) {
+                            resultItems[element['page']] = '<span class="structure">'
+                                + $('#tx-dlf-search-in-document-label-page').text() + ' ' + element['page']
+                                + '</span><br />'
+                                + '<span ="textsnippet">' +
```
albig:
```suggestion
                                + '<span ="textsnippet">' 
```

albig: One `+` too much destroys the output.

Classes/Plugin/Eid/SearchInDocument.php
albig: Ahh, ok. So xdebug just spoofed me? Sorry for that. But without I wouldn't find the other bugs ;-)

Classes/Plugin/Eid/SearchInDocument.php
```diff
@@ -38,29 +37,48 @@ class SearchInDocument
...
+        // Perform Solr query.
+        $solr = Solr::getInstance($core);
+        if ($solr->ready) {
+            $query = $solr->service->createSelect();
+            $query->setFields(['id', 'uid', 'page']);
+            $query->setQuery('fulltext:(' . Solr::escapeQuery((string) $parameters['q']) . ') AND uid:' . intval($parameters['uid']));
+            $query->setStart($count)->setRows(20);
+            $hl = $query->getHighlighting();
+            $hl->setFields(['fulltext']);
+            $hl->setUseFastVectorHighlighter(true);
```
albig, line 62: Perfect. This does the job with Solarium < 6.

Classes/Plugin/Eid/SearchInDocument.php
albig: This works now.

Resources/Public/Javascript/Search/SearchInDocument.js
sebastian-meyer: I hate Javascript...

Resources/Public/Javascript/Search/SearchInDocument.js
sebastian-meyer: Did I mention I hate Javascript?

sebastian-meyer: And why did no static code analyzer and syntax checker find this?

Classes/Hooks/DataHandler.php
albig: Perfect. This works as expected. The core is only deleted if this option is activated.

Resources/Public/Javascript/Search/SearchInDocument.js
albig: Don't know. Even the browser didn't complain but scrambels the output. There was some `NaNhttp://` instead of the `<a>`-tag.

sebastian-meyer: I've fixed this with a29a7ca

Resources/Public/Javascript/Search/SearchInDocument.js
sebastian-meyer: I've (hopefully) fixed this with eaf0096

Resources/Public/Javascript/Search/Suggester.js
```diff
@@ -31,14 +31,14 @@ $(
 					function(data) {
 						var result = [];
 						var option = "";
-						$("arr[name='suggestion'] str", data).each(function(i) {
+						data.each(function(i) {
```
albig: Uncaught TypeError: data.each is not a function

albig: But data is empty. Will dig deeper into it.

albig: The old code executed the following query (from Solr log):
```
  webapp=/solr path=/suggest params={q=dresde&wt=xml} status=0 QTime=2
```
--> 1 Result

The new code executes the following query:
```
webapp=/solr path=/suggest params={json.nl=flat&omitHeader=true&suggest.build=false&suggest.q=dresd&suggest.count=10&suggest.reload=false&suggest=true&suggest.dictionary=suggest&wt=json} status=0 QTime=0
```
--> 0 Results (of course on same core)

albig: I think the problem is, that we configure in solrconfig.xml the SpellCheckComponent as way to get a autosuggest feature. But we use the Suggester component. I assume, this is the better & newer way.

From https://lucene.apache.org/solr/guide/7_7/suggester.html: 

> Although it is possible to use the Spell Checking functionality to power autosuggest behavior, Solr has a dedicated SuggestComponent designed for this functionality.

So we get a spellchecking-response, if we do a query like:
```
/suggest?q=dresde&wt=json&omitHeader=true
{
  "spellcheck":{
    "suggestions":[
      "dresde",{
        "numFound":2,
        "startOffset":0,
        "endOffset":6,
        "suggestion":["dresden-leipziger schnellpressenfabrik ag, coswig",
          "dresdener pflasterstein-werke aktiengesellschaft, dresden-leubnitz"]}]}}
```
But if we use the SuggestComponent:
```
/suggest?suggest.build=false&suggest.q=dresde&&suggest=true&suggest.dictionary=suggest&wt=json
{
  "responseHeader":{
    "status":0,
    "QTime":0}}
```
(Mind the missing `q=` paramter!

### Conclusion

There are two errors in the code:
1. We cannot use the SuggestComponent if we do not modify the solrconfig.xml
2. For emty results (empty array) the JavaScript code fails. There must be a check if  we may iterate with each or not.

albig (changes requested):
The autosuggest feature doesn't work like this. The other features seems to work as expected.

Resources/Public/Javascript/Search/Suggester.js
sebastian-meyer: Well, I actually forgot to use the `solrconfig.xml` from Kitodo when testing the autocompletion, but used the default Solr configuration and some hand-picked sample data instead.
I'll set up a proper testing environment and take a look at this again - but not today. ;)

sebastian-meyer: I replaced the suggester query with the old dedicated `/suggest` request handler. It's bad design, but the only way without losing backwards compatibility...

Classes/Plugin/Eid/SearchSuggest.php
```diff
@@ -39,25 +38,34 @@ class SearchSuggest
...
+        $solr = Solr::getInstance($core);
+        if ($solr->ready) {
+            $query = $solr->service->createSelect();
+            $query->setHandler('suggest');
+            $query->setQuery(Solr::escapeQuery((string) $parameters['q']));
+            $query->setRows(0);
+            $results = $solr->service->select($query)->getResponse()->getBody();
+            $result = json_decode($results);
+            foreach ($result->spellcheck->suggestions as $suggestions) {
+                foreach ($suggestions->suggestion as $suggestion) {
```
albig: This does through an exception because the first array element is no object.

That's because of the result of json_decode().
Response from Solr:
`/suggest?q=dresde&json.nl=flat&omitHeader=true&fl=*,score&start=0&rows=0&wt=json`
```
{
  "spellcheck":{
    "suggestions":[
      "dresde",{
        "numFound":2,
        "startOffset":0,
        "endOffset":6,
        "suggestion":["dresdener pflasterstein-werke aktiengesellschaft, dresden-leubnitz",
          "dresden-leipziger schnellpressenfabrik ag, coswig"]}]}}
```
Result of json_decode():
```
object(stdClass)#3 (1) {
  ["spellcheck"]=>
  object(stdClass)#2 (1) {
    ["suggestions"]=>
    array(2) {
      [0]=>
      string(6) "dresde"
      [1]=>
      object(stdClass)#1 (4) {
        ["numFound"]=>
        int(2)
        ["startOffset"]=>
        int(0)
        ["endOffset"]=>
        int(6)
        ["suggestion"]=>
        array(2) {
          [0]=>
          string(66) "dresdener pflasterstein-werke aktiengesellschaft, dresden-leubnitz"
          [1]=>
          string(49) "dresden-leipziger schnellpressenfabrik ag, coswig"
        }
      }
    }
  }
}

```
If we use json_decode with the optional paramter to get an associative array, the result would be:
```
array(1) {
  ["spellcheck"]=>
  array(1) {
    ["suggestions"]=>
    array(2) {
      [0]=>
      string(6) "dresde"
      [1]=>
      array(4) {
        ["numFound"]=>
        int(2)
        ["startOffset"]=>
        int(0)
        ["endOffset"]=>
        int(6)
        ["suggestion"]=>
        array(2) {
          [0]=>
          string(66) "dresdener pflasterstein-werke aktiengesellschaft, dresden-leubnitz"
          [1]=>
          string(49) "dresden-leipziger schnellpressenfabrik ag, coswig"
        }
      }
    }
  }
}

```
So we must check for a valid object or array before calling the foreach() loop.

Classes/Plugin/Eid/SearchSuggest.php
```diff
@@ -39,25 +38,34 @@ class SearchSuggest
...
+                    $output[] = $suggestion;
+                }
```
albig:
```suggestion
                if (is_object($suggestions)) {
                    foreach ($suggestions->suggestion as $suggestion) {
                        $output[] = $suggestion;
                    }
                }
```

albig (changes requested):
Despite of the last change request, this seems to work.

So solrconfig.xml has to be reviewed and probably updated anyhow. Maybe we should do this in a coming major release and change to the suggester component then.

There are still two Scrutinizer warnings, I do not understand. It seems to work nevertheless.

sebastian-meyer:
I added the task to rewrite the autocompletion to use the proper suggester component to #454.

albig (approved):

PR #571, opened 2020-12-23
Use numerical value for activation of the full text initially and add on/off behavior on link click

### LINKED ISSUE #570: Fulltext switch doesn't work as expected
@albig reports:

Sorry, I cannot reproduce what this setting does:

* Without launchState: nothing happens. Ok.
* with launchState=off: the fulltext is shown initially and on every click on the button it is marked once more. So the color gets darker and darker (in the screenshot its getting more and more red)
* with launchState=on: the old behaviour shows up

So what is launchState=off good for?
fulltext-launchstate

_Originally posted by @albig in https://github.com/kitodo/kitodo-presentation/pull/563#discussion_r547549674_

beatrycze-volk:
Fix for #570


Resources/Public/Javascript/PageView/FulltextControl.js
```diff
@@ -433,6 +433,10 @@ dlfViewerFullTextControl.prototype.enableFulltextSelect = function() {
         .attr('title', this.dic['fulltext-off']);
     }
 
+    if(this.dic['launch-state'] === 'off') {
+        $("#tx-dlf-tools-fulltext").css('visibility', 'hidden');
+    }
+
```
sebastian-meyer: This should be removed. We want to disable the link only in the Zeitungsportal, so this should be added to some custom CSS.

sebastian-meyer:
The bug report suggests, that this feature isn't working as expected. You didn't address this with your pull request, but just changed the default value of `launchState`. Is this a misunderstanding or couldn't you reproduce the issue?

Resources/Public/Javascript/PageView/FulltextControl.js
beatrycze-volk: I have understood that `off` is a custom thing for Zeitungsportal, so if the full text is always visible (off) the link should be invisble. If setting is set up to `on`, it should be all like before.

sebastian-meyer: There are actually three different cases:

* If someone wants to have fulltexts initially disabled when loading a page, but available to be turned on by the user: `launchState = off`.
* If someone wants to have fulltexts initially shown when loading a page, but possible for the user to turn it off: `launchState = on`.

Those two general cases essentially work the same: Fulltext can be switched on and off by a user. The only difference is, if the fulltext is initially turned on or off when loading the page and without any user interaction.

Then there is a third case, which we only need for the Zeitungsportal:

* We want the fulltext to be turned on by default (`launchState = on`), but the user shouldn't be able to turn it off. We don't need any code in Kitodo.Presentation to achieve this, as we can simply use the second option from above and hide the link. The latter is easy, since we can add some custom CSS to the page or just omit the placeholder from the tool's template.

beatrycze-volk: Oh, I have understood it a bit differently. So actually it needs additional checkbox which is visible in the frontend and Zeitungsportal needs on their side hide it by css. Also logic of the switches needs to be reveresed.

sebastian-meyer: The checkbox isn't needed since the `launchState` in Typoscript is sufficient, but apart from that you're right.

beatrycze-volk: Checkbox removed, logic reveresd, clicking on 'Volltext' makes nothing as all click actions are removed from this text.

albig (changes requested):
Unfortunately, it still does not work as expected.

I observe the follwing behaviour:

### 1. Setting of launchState not available
`plugin.tx_dlf_fulltexttool.launchState = `

Nothing happens. No fulltext is open initially. No fulltext and no reaction (even no error) on click on the activate link.

### 2. Setting of launchState set to "on"
`plugin.tx_dlf_fulltexttool.launchState = on`

* fulltext is active on load
* no reaction on click on the link --> deactivation is not possible

Markup on load:
```
<a class="select switchoff fulltext-visible active" id="tx-dlf-tools-fulltext" title="Volltext" data-dic="fulltext:Volltext;fulltext-on:Volltext an;fulltext-off:Volltext aus;launch-state:on">Volltext</a>
```

### 3. Setting of launchState set to "off"
`plugin.tx_dlf_fulltexttool.launchState = off`

Markup on load:
```
<a class="select switchoff" id="tx-dlf-tools-fulltext" title="Volltext an" data-dic="fulltext:Volltext;fulltext-on:Volltext an;fulltext-off:Volltext aus;launch-state:off">Volltext an</a>
```
* Fulltext is not shown on initial load
* Click on link activates the fulltext
* click on link deactivates the fulltext again

Markup after activation of fulltext
```
<a class="select switchoff fulltext-visible active" id="tx-dlf-tools-fulltext" title="Volltext aus" data-dic="fulltext:Volltext;fulltext-on:Volltext an;fulltext-off:Volltext aus;launch-state:off">Volltext aus</a>
```

The behaviour of `launchState = off` is the old one. This should be the same as if launchState is not set.

Last: There is a possible security issue in this solution. You can set the launchState to whatever you want. Maybe it would better to work with a boolean type and check forr it? In this case another name could be better like `ActivateFulltextInitially = 1`.

beatrycze-volk:
> Unfortunately, it still does not work as expected.
> 
> I observe the follwing behaviour:
> 
> ### 1. Setting of launchState not available
> `plugin.tx_dlf_fulltexttool.launchState = `
> 
> Nothing happens. No fulltext is open initially. No fulltext and no reaction (even no error) on click on the activate link.
> 

This shouldn't be empty, it has default value setup in setup.txt: `plugin.tx_dlf_fulltexttool.launchState = off`.

> ### 2. Setting of launchState set to "on"
> `plugin.tx_dlf_fulltexttool.launchState = on`
> 
> * fulltext is active on load
> * no reaction on click on the link --> deactivation is not possible
> 

But is it not what was expected here? Full text always active, behavior can be change by changing `plugin.tx_dlf_fulltexttool.launchState` to off. Should click on the link change the behavior to the option `on`? 
> Markup on load:
> 
> ```
> <a class="select switchoff fulltext-visible active" id="tx-dlf-tools-fulltext" title="Volltext" data-dic="fulltext:Volltext;fulltext-on:Volltext an;fulltext-off:Volltext aus;launch-state:on">Volltext</a>
> ```
> 
> ### 3. Setting of launchState set to "off"
> `plugin.tx_dlf_fulltexttool.launchState = off`
> 
> Markup on load:
> 
> ```
> <a class="select switchoff" id="tx-dlf-tools-fulltext" title="Volltext an" data-dic="fulltext:Volltext;fulltext-on:Volltext an;fulltext-off:Volltext aus;launch-state:off">Volltext an</a>
> ```
> 
> * Fulltext is not shown on initial load
> * Click on link activates the fulltext
> * click on link deactivates the fulltext again
> 
> Markup after activation of fulltext
> 
> ```
> <a class="select switchoff fulltext-visible active" id="tx-dlf-tools-fulltext" title="Volltext aus" data-dic="fulltext:Volltext;fulltext-on:Volltext an;fulltext-off:Volltext aus;launch-state:off">Volltext aus</a>
> ```
> 
> The behaviour of `launchState = off` is the old one. This should be the same as if launchState is not set.

Yes, off is exactly the old one. There was nothing change on it.
> 
> Last: There is a possible security issue in this solution. You can set the launchState to whatever you want. Maybe it would better to work with a boolean type and check forr it? In this case another name could be better like `ActivateFulltextInitially = 1`.

I will change to number values.

albig:
> > ### 1. Setting of launchState not available
> 
> This shouldn't be empty, it has default value setup in setup.txt: `plugin.tx_dlf_fulltexttool.launchState = off`.

Yes, but there are possibilities to loose, ignore, forget, whatever this TypoScript setting. We use the tools usually not inside the toolbox anymore. E.g. in DFG-Viewer. So the behaviour should fall back to some default state if this setting is missing.

> > ### 2. Setting of launchState set to "on"

> But is it not what was expected here? Full text always active, behavior can be change by changing `plugin.tx_dlf_fulltexttool.launchState` to off. Should click on the link change the behavior to the option `on`?

That's not what the name of this option implies: `launchState=on` == "state of fulltext on launch is on (visible)". This doesn't mean that I cannot disable the fulltext.

Or like @sebastian-meyer  wrote above: 

> If someone wants to have fulltexts initially shown when loading a page, but possible for the user to turn it off: launchState = on.

Classes/Plugin/Tools/FulltextTool.php
```diff
@@ -75,7 +75,7 @@ public function main($content, $conf)
         $this->getTemplate();
         $fullTextFile = $this->doc->physicalStructureInfo[$this->doc->physicalStructure[$this->piVars['page']]]['files'][$this->conf['fileGrpFulltext']];
         if (!empty($fullTextFile)) {
-            $markerArray['###FULLTEXT_SELECT###'] = '<a class="select switchoff" id="tx-dlf-tools-fulltext" title="" data-dic="fulltext:' . htmlspecialchars($this->pi_getLL('fulltext', '')) . ';fulltext-on:' . htmlspecialchars($this->pi_getLL('fulltext-on', '')) . ';fulltext-off:' . htmlspecialchars($this->pi_getLL('fulltext-off', '')) . ';activate-full-text-initially:' . $this->conf['activateFullTextInitially'] .'">&nbsp;</a>';
+            $markerArray['###FULLTEXT_SELECT###'] = '<a class="select switchoff" id="tx-dlf-tools-fulltext" title="" data-dic="fulltext:' . htmlspecialchars($this->pi_getLL('fulltext', '')) . ';fulltext-on:' . htmlspecialchars($this->pi_getLL('fulltext-on', '')) . ';fulltext-off:' . htmlspecialchars($this->pi_getLL('fulltext-off', '')) . ';activate-full-text-initially:' . \TYPO3\CMS\Core\Utility\MathUtility::forceIntegerInRange($this->conf['activateFullTextInitially'], 0, 1, 0) .'">&nbsp;</a>';
```
albig: This is just to make sure, only `0` or `1` will go into the markup. Everything else get's `0`.

albig (dismissed):
This works now for my three use-cases:

1. no TypoScript setting --> old behaviour
2. `activateFullTextInitially=0` --> old behaviour
3. `activateFullTextInitially=1` --> fulltext is shown initially but can be closed

sebastian-meyer (approved):

PR #575, opened 2020-12-29
Add CLI command for OAI harvesting

sebastian-meyer:
This adds a new CLI command `kitodo:harvest` in order to add documents by harvesting an OAI-PMH interface.

In order to use the command, you first have to configure a library in the backend (`tx_dlf_libraries`), setting at least a `label` and `oai_base`. The latter should be a valid OAI-PMH base URL (e.g. https://digital.slub-dresden.de/oai/).

Then you can add/index documents by harvesting via:
`./vendor/bin/typo3 kitodo:harvest --lib=<UID> --pid=<PID> --solr=<CORE> --from=<timestamp> --until=<timestamp> --set=<set>`

The parameters mean:
* `<UID>` is the UID of the library record with the OAI interface that should be harvested.
* `<PID>` is the page UID of the library record and therefore the page the documents are added to.
* `<CORE>` UID or *index_name* of the Solr core to use (which has to have the same `<PID>` as above).
* `<timestamp>` is a timestamp in the format `YYYY-MM-DD`. The parameters `from` and `until` limit harvesting to the given period, e.g. for incremental updates. Those parameters are optional.
* `<set>` is the name of a OAI set. The parameter limits harvesting to the given set. This parameter is optional.

**Attention:** This introduces changes to the TCA configuration of `tx_dlf_libraries.oai_base` and changes it from `int(11)` to `varchar(255)` (therefore requiring a database update). Although this field was never fully implemented in its originally intended way, there still is the possibility that someone actually set this to a TYPO3 page UID. With the proposed changes, the field now holds the base URL to a library's OAI-PMH interface. Having the field set to a PID won't break anything, but it may be confusing.

---

While implementing the harvester I stumbled upon a few bugs and small inconveniences I tried to fix along the way:
* It is now possible to explicitly set an `owner` for documents when using the `kitodo:index` and `kitodo:reindex` commands. This is done by setting the optional parameter `--owner|-o` with the *index_name* or *UID* of a library record. The document will then by associated with this library regardless of the configuration for the metadata field `owner`. When omitted, the old behavior is unchanged.
* ~~It is now possible to configure the target element the OpenLayers OverviewMap should be rendered into. Just set `plugin.tx_dlf_pageview.overviewMapElementId` to the `@id` attribute of the target element. The default is unchanged. (This change was provided by @erik-konrad for the DDB Zeitungsportal.)~~ Reverted this change in favor of #584 which provides the same functionality.
* Since collections can be configured as search queries nowadays, it is pointless to have `tx_dlf_documents.collections` defined as a required field. I changed the TCA configuration for `minitems` accordingly to remove the annoying backend warnings.

---

> This pull request is sponsored by [Open Culture Consulting](https://www.opencultureconsulting.com).


Classes/Common/MetsDocument.php
```diff
@@ -775,18 +793,6 @@ protected function getDocument()
...
-     *
-     * @access protected
-     *
-     * @return int The PID of the metadata definitions
-     */
-    protected function _getCPid()
-    {
-        return $this->cPid;
-    }
-
```
sebastian-meyer, line 799: Moved to parent class `\Kitodo\Dlf\Common\Document`.

sebastian-meyer:
Sorry for the mix-up of different features in one pull request. When starting development of the harvest command I wasn't aware of older changes in the branch I had made for the Zeitungsportal...

sebastian-meyer:
I reverted changes which were not part of the OAI harvesting feature. Those changes are instead included in #584.
Hopefully this makes reviewing easier and allows merging of both pull requests without conflicts.

Classes/Command/ReindexCommand.php
```diff
@@ -127,6 +133,16 @@ protected function execute(InputInterface $input, OutputInterface $output)
             exit(1);
         }
 
+        if (!empty($input->getOption('owner'))) {
+            if (MathUtility::canBeInterpretedAsInteger($input->getOption('owner'))) {
```
albig, line 137: Please add 

`use TYPO3\CMS\Core\Utility\MathUtility;`

It's not part of ReindexCommand.php, yet.

Classes/Common/Helper.php
```diff
@@ -393,6 +393,59 @@ public static function getIndexNameFromUid($uid, $table, $pid = -1)
...
+     */
+    public static function getLabelFromUid($uid, $table, $pid = -1)
+    {
+        // Sanitize input.
+        $uid = max(intval($uid), 0);
+        if (
+            !$uid
+            || !in_array($table, ['tx_dlf_collections', 'tx_dlf_libraries', 'tx_dlf_metadata', 'tx_dlf_structures', 'tx_dlf_solrcores'])
+        ) {
+            self::devLog('Invalid UID "' . $uid . '" or table "' . $table . '"', DEVLOG_SEVERITY_ERROR);
```
albig: We removed` self:devLog()` in the meantime. Please use `self:log()`.

albig (changes requested):
My first review on this pull request.

Please have a look on my comments.

Please add your documentation to https://github.com/kitodo/kitodo-presentation/blob/master/Documentation/User/Index.rst

Classes/Command/ReindexCommand.php
sebastian-meyer: Done! And while at it I also removed the no longer needed use statements for Symfony's base `Command` class since all Kitodo commands now have their own `BaseCommand` class.

Classes/Common/Helper.php
sebastian-meyer: Done!

sebastian-meyer:
Thanks for your review! I made all requested changes and added documentation.

Classes/Command/ReindexCommand.php
albig: That's new. Thank you for adjusting.

Configuration/TCA/tx_dlf_documents.php
```diff
@@ -385,7 +385,7 @@
                 'foreign_table_where' => 'AND tx_dlf_collections.pid=###CURRENT_PID### AND tx_dlf_collections.sys_language_uid IN (-1,0) ORDER BY tx_dlf_collections.label',
                 'size' => 5,
                 'autoSizeMax' => 15,
-                'minitems' => 1,
+                'minitems' => 0,
```
albig, line 388: I'm not sure about this change. Of course I can live with a required field less. But are you sure, all plugins work with an empty collections field?

What's the relation with the collections field and the OAI harvesting feature?

Documentation/User/Index.rst
```diff
@@ -197,3 +196,119 @@ collections or even to reindex all documents on the given page.::
        Show each processed documents uid and location with timestamp and
        amount of processed/all documents.
    :Example:
+
```
albig, line 229: Thank you.

Please add the "--owner" command to the kitodo:index and kitodo:reindex command descriptions, too.

albig (changes requested):
One question and a small request.

Configuration/TCA/tx_dlf_documents.php
sebastian-meyer: As far as I checked, there are no plugins requiring the collection field. The OAI plugin may use collections as OAI sets, but that's optional.

The worst thing that could happen is some documents being "invisible" in plugins like the collection plugin if they have an empty collection field and are not part of any query-based custom collection. But that would be the expected behaviour anyway, I guess?

Documentation/User/Index.rst
sebastian-meyer: Ah, sorry. I missed that!

Configuration/TCA/tx_dlf_documents.php
albig: I agree. This is the expected behaviour.

I found only the feeds plugin. New documents without collection are not listed in the RSS feed. But this should be discussed on the plugin side.

albig (approved):
Thank you for this new feature. It worked (almost) out of the box.

I run into problems with our own OAI provider. That's nice to have a CLI tool now, to "test" it.

I wonder a little about the OAI plugin itself. You can select the library. But this seems to be used only for the `Identify` verb. Later on, all documents in `tx_dlf_documents` with a collection are shown. Not sure, if this is intended. But this would be a new issue.

PR #576, opened 2021-01-04
Add core status widget

sebastian-meyer:
This pull request adds a small status message for Solr cores. When editing a `tx_dlf_solrcores` record in the backend, the status of the respective core is fetched from Solr and shown as flash message.

This is done by implementing a new `FieldInformation` renderType for input fields.

---

In addition this pull request refactors the recently introduced warning message for changing the `index_name` of records (#553) from `FieldWizard` to `FieldInformation` as well. Originally this was implemented as `FieldWizard` because `FieldInformation` doesn't allow all HTML tags of a flash message. But in fact, only the `<h4>` tag of the flash message's title is disallowed, so by just omitting the title we can still use flash messages as `FieldInformation`.
Since a static warning message should logically be a `FieldInformation` I decided to refactor this.

### DIFF
--- a/Classes/Hooks/Form/FieldInformation/SolrCoreStatus.php
+++ b/Classes/Hooks/Form/FieldInformation/SolrCoreStatus.php
@@ -10,7 +10,7 @@
  * LICENSE.txt file that was distributed with this source code.
  */
 
-namespace Kitodo\Dlf\Hooks\Form\FieldWizard;
+namespace Kitodo\Dlf\Hooks\Form\FieldInformation;
 
 use Kitodo\Dlf\Common\Helper;
 use Kitodo\Dlf\Common\Solr;
--- a/Classes/Hooks/Form/FieldInformation/SolrCoreStatus.php
+++ b/Classes/Hooks/Form/FieldInformation/SolrCoreStatus.php
@@ -32,6 +32,7 @@ class SolrCoreStatus extends AbstractNode
      * @access public
      *
      * @return array As defined in initializeResultArray() of AbstractNode
+     *               Allowed tags are: "<a><br><br/><div><em><i><p><strong><span><code>"
      */
     public function render(): array
     {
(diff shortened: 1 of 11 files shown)


Classes/Hooks/Form/FieldInformation/SolrCoreStatus.php
```diff
@@ -0,0 +1,78 @@
...
+                // Get core data.
+                $coreAdminQuery = $solr->service->createCoreAdmin();
+                $action = $coreAdminQuery->createStatus();
+                $action->setCore($core);
+                $coreAdminQuery->setAction($action);
+                $response = $solr->service->coreAdmin($coreAdminQuery)->getStatusResult();
+                $uptimeInSeconds = floor($response->getUptime() / 1000);
+                $uptime = floor($uptimeInSeconds / 3600) . gmdate(":i:s.", $uptimeInSeconds % 3600) . $response->getUptime() % 1000;
+                $numDocuments = $response->getNumberOfDocuments();
+                $startTime = $response->getStartTime() ? date_format($response->getStartTime(), 'c') : 'N/A';
```
albig: "Start Time" is confusing in my opinion. I would prefer "Creation Time" because it''s the creation time of the core. Not the start time of the Solr. instance.

Classes/Hooks/Form/FieldInformation/SolrCoreStatus.php
```diff
@@ -0,0 +1,78 @@
...
+            // Get Solr instance.
+            $solr = Solr::getInstance($core);
+            if ($solr->ready) {
+                // Get core data.
+                $coreAdminQuery = $solr->service->createCoreAdmin();
+                $action = $coreAdminQuery->createStatus();
+                $action->setCore($core);
+                $coreAdminQuery->setAction($action);
+                $response = $solr->service->coreAdmin($coreAdminQuery)->getStatusResult();
+                $uptimeInSeconds = floor($response->getUptime() / 1000);
```
albig: Would be easier to read if adding the amount of days. E.g. DD HH:MM:SS.

I think, this  is what the Solr-Core Admin-Panel writes as "startTime: 16 days ago".

Uptime is better than "Start Time". But The "Star Time" below seems to be the creation time. I'm a little confused.

albig (changes requested):
It's working. Thank you and it's helpful.

The given times are a little confusing. Please check again.

This is what TYPO3 tells me:
	
```
Start Time: 2021-01-20T14:50:16+00:00
Uptime: 22:08:43.511
Last Modified: 2021-01-20T15:35:42+00:00
Number of Documents: 2832
```

This is, what Solr tells me about the Core:

```
startTime:about 22 hours ago
lastModified:about 22 hours ago
Num Docs:2832
```

Classes/Hooks/Form/FieldInformation/SolrCoreStatus.php
sebastian-meyer: Actually `startTime` is the time the core was loaded. This can be the creation date or any later time when a core reload or Solr restart happened.
Thus `startTime` and `Uptime` are redundant: `Uptime` is just `NOW - startTime`.

Classes/Hooks/Form/FieldInformation/SolrCoreStatus.php
sebastian-meyer: Yeah, I agree that adding the days would be convenient, but it's not easy to calculate: https://stackoverflow.com/questions/12918158/timestamp-to-days-hours-minutes

sebastian-meyer: I agree that it would be nice to have the days as well, but it's complicated to calculate. There are so many ways and each has its disadvantages: https://stackoverflow.com/questions/8273804/convert-seconds-into-days-hours-minutes-and-seconds
I think it's just not worth it...

sebastian-meyer:
I changed the date output to be localized in order to make it less confusing. But I suggest keeping the `Uptime` as it is.

sebastian-meyer:
Thanks for your proposal to use the Linux uptime format. I merged your pull request.

albig (approved):

PR #577, opened 2021-01-12
Remove double submit of the search form

beatrycze-volk:
Method `resetStart` is called on the submit button, so at the end `submit` method is called two times.


albig (approved):
Should work. Thank you.

PR #579, opened 2021-01-19
Add support for Solarium 4.2

### LINKED ISSUE #578: Composer dependencies of solarium client
The solarium updates creates an incompatibility to other extensions e.g. ext:solr. For systems using both extensions this is problematic. So no composer updates/installations will work anymore.
https://github.com/TYPO3-Solr/ext-solr/blob/master/composer.json#L33
https://github.com/kitodo/kitodo-presentation/blob/master/composer.json#L38

sebastian-meyer:
This adds support for Solarium 4.2 and Solarium 5.2.

Fixes #578.

### DIFF
--- a/Classes/Common/Solr.php
+++ b/Classes/Common/Solr.php
@@ -624,13 +624,15 @@ class Solr
         $this->loadSolrConnectionInfo();
         // Configure connection adapter.
         $adapter = GeneralUtility::makeInstance(\Solarium\Core\Client\Adapter\Http::class);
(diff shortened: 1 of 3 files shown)


albig (approved):
Seems to work for me with Solarium 4.2 and 5.2.

Please @ByteParty have a look, too.

ByteParty (approved):
after i changed the solr path in the extension config, it works with solarium 4.2

PR #580, opened 2021-01-19
Use AES-256-CTR and SHA256

sebastian-meyer:
Currently Kitodo.Presentation uses a blowfish algorithm (`BF-CFB`) for encrypting/decrypting form data. This is considered weak nowadays (see [SonarCloud Analysis](https://sonarcloud.io/project/issues?id=kitodo-presentation&resolved=false&severities=BLOCKER%2CCRITICAL%2CMAJOR%2CMINOR&sinceLeakPeriod=true&types=VULNERABILITY)).

This pull request changes the defaults to `AES-256-CTR` as ciphering algorithm and `SHA256` as hashing algorithm.

Furthermore it removes the additional hashed value, because this opens a potential security hole and provides no real value.

### DIFF
--- a/Classes/Common/Helper.php
+++ b/Classes/Common/Helper.php
@@ -189,7 +189,7 @@ class Helper
         $data = substr($encrypted, openssl_cipher_iv_length(self::$cipherAlgorithm));
         $key = openssl_digest($GLOBALS['TYPO3_CONF_VARS']['SYS']['encryptionKey'], self::$hashAlgorithm, true);
         // Decrypt data.
-        $decrypted = openssl_decrypt($data, self::$cipherAlgorithm, $key, OPENSSL_RAW_DATA, $iv);
+        $decrypted = openssl_decrypt($data, self::$cipherAlgorithm, $key, 0, $iv);
         return $decrypted;
     }
 
@@ -282,7 +282,7 @@ class Helper
         $iv = openssl_random_pseudo_bytes(openssl_cipher_iv_length(self::$cipherAlgorithm));
         $key = openssl_digest($GLOBALS['TYPO3_CONF_VARS']['SYS']['encryptionKey'], self::$hashAlgorithm, true);
         // Encrypt data.
-        $encrypted = openssl_encrypt($string, self::$cipherAlgorithm, $key, OPENSSL_RAW_DATA, $iv);
+        $encrypted = openssl_encrypt($string, self::$cipherAlgorithm, $key, 0, $iv);
         // Merge initialisation vector and encrypted data.
         if ($encrypted !== false) {
             $encrypted = $iv . $encrypted;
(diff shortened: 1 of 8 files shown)


Classes/Common/Helper.php
```diff
@@ -214,26 +236,58 @@ public static function devLog($message, $severity = 0)
...
         if (empty($GLOBALS['TYPO3_CONF_VARS']['SYS']['encryptionKey'])) {
             self::devLog('No encryption key set in TYPO3 configuration', DEVLOG_SEVERITY_ERROR);
-            return;
+            return false;
+        }
+        // Generate random initialisation vector.
+        $iv = openssl_random_pseudo_bytes(openssl_cipher_iv_length(self::$cipherAlgorithm));
+        $key = openssl_digest($GLOBALS['TYPO3_CONF_VARS']['SYS']['encryptionKey'], self::$hashAlgorithm, true);
+        // Encrypt data.
+        $encrypted = openssl_encrypt($string, self::$cipherAlgorithm, $key, OPENSSL_RAW_DATA, $iv);
```
albig, line 286: This fails with the following exception (including binary data):

<pre>
PHP Warning: openssl_encrypt(): A tag should be provided when using AEAD mode in /var/www/extensions/dlf/Classes/Common/Helper.php line 285

at openssl_encrypt('dlfCore1', 'aes-256-gcm', '��NÒ�&���|��m���UuV�P�|�|uG(@', 1, 'Ie�:UsZ�')
</pre>

What am I missing? I've just have the search plugin on the page.

albig (changes requested):
I cannot test this.

Actually this is a lot of code to do some core functionality. Maybe we should better invest the time to swith to fluid templates?

Classes/Common/Helper.php
sebastian-meyer: Sorry, my fault! I experimented with different cipher algorithms and didn't push the final choice to upstream... This should be fixed now.

Explanation: The `*-gcm` and `*-ccm` cipher methods use authentication (AEAD mode) and therefore need a tag. But we don't need authentication, so I changed the cipher method to `AES-256-CTR`. This is the same as `AES-256-GCM`, but without authentication.

sebastian-meyer:
> Actually this is a lot of code to do some core functionality. Maybe we should better invest the time to swith to fluid templates?

How would using fluid templates make this obsolete?

Classes/Common/Helper.php
sebastian-meyer: Although this worked fine, I added base64-encoding to get rid of the binary data in HTML. I think it's less confusing now.

albig:
> > Actually this is a lot of code to do some core functionality. Maybe we should better invest the time to swith to fluid templates?
> 
> How would using fluid templates make this obsolete?

Protecting forms with hidden fields is daily business of the form-ViewHelper. Nothing you should think about on extension development.

albig (approved):
Works out-of-the box now. Thank you.

PR #584, opened 2021-01-26
Update OpenLayers to 6.5.0

sebastian-meyer:
This pull request updates OpenLayers used in `PageView` plugin to version 6.5.0.

This is still work in progress, but all the basic features should be working now:
- [x] loading images from static, IIP, Zoomify, and IIIF sources
- [x] displaying two images simultaneous in double-page mode
- [x] standard map interactions (drag, rotate, pan, zoom - with mouse, keyboard and touch)
- [x] standard map controls (zoom buttons, zoom slider, zoom-to-extent button, overview map, fullscreen mode, reset rotation)
As before they can be selected in the plugin's settings in the backend.
- [x] added custom buttons for 90 degree rotation to rotate control
- [x] rendering map controls in custom HTML elements outside the map (as needed for the Zeitungsportal)
This can be configured by setting `plugin.tx_dlf_pageview.<controlName>ElementId` to the `@id` of the target HTML element, i. e. `plugin.tx_dlf_pageview.overviewMapElementId = tx-dlf-overviewmap`.
- [x] removed static map controls from `Navigation` plugin, since they can now be placed dynamically by using the above mentioned `plugin.tx_dlf_pageview.<controlName>ElementId`
If you want to keep the old behavior, just add `<div id="tx-dlf-<controlName>"></div>` to the `Navigation` template and set `plugin.tx_dlf_pageview.<controlName>ElementId = tx-dlf-<controlName>` for every map control you want to keep (the legacy configuration would be `Zoom`, `FullScreen`, `Rotate`).
- [x] added a new image manipulation control (and removed the old `ImageManipulationTool`)
The new control solely uses CSS filters for image manipulation and therefore doesn't need WebGL anymore.
- [x] added a new magnifier control (and removed the old implementation that was bundled with the `Basket` feature)

Still working on:
- [ ] displaying fulltext and annotations
- [ ] highlighting search hits

**Attention:** Since OpenLayers 6.5.0 doesn't support Internet Explorer anyway, I've dropped support completely and used a lot of newer ECMAScript features. We could try to support IE by using Polyfills, but we'd need _a lot_ of them.


sebastian-meyer:
Temporarily working dev system: https://vdl.opencultureconsulting.com/

sebastian-meyer:
@beatrycze-volk The overview map in this branch works the way it's supposed to in the Zeitungsportal.

Classes/Plugin/OaiPmh.php
```diff
@@ -352,7 +352,7 @@ public function main($content, $conf)
             $stylesheet = \TYPO3\CMS\Core\Utility\GeneralUtility::locationHeaderUrl($this->conf['stylesheet']);
         } else {
             // Use default stylesheet if no custom stylesheet is given.
-            $stylesheet = \TYPO3\CMS\Core\Utility\GeneralUtility::locationHeaderUrl(\TYPO3\CMS\Core\Utility\PathUtility::stripPathSitePrefix(\TYPO3\CMS\Core\Utility\ExtensionManagementUtility::extPath($this->extKey)) . 'Resources/Public/Stylesheets/OaiPmh.xsl');
+            $stylesheet = \TYPO3\CMS\Core\Utility\GeneralUtility::locationHeaderUrl(\TYPO3\CMS\Core\Utility\PathUtility::stripPathSitePrefix(\TYPO3\CMS\Core\Utility\ExtensionManagementUtility::extPath($this->extKey)) . 'Resources/Public/Stylesheets/OaiPmh/OaiPmh.xsl');
```
beatrycze-volk: Maybe using here short class names would make this line a bit easier to read.

sebastian-meyer: Thanks for the hint! I've changed it.

sebastian-meyer:
@beatrycze-volk + @albig: I am done now with everything except the fulltext stuff. I've tested this cross-browser and it works in everything except IE. A working test system can be found [here](https://vdl.opencultureconsulting.com). How should I go on? Merge this and implement fulltext support in a new branch? This would make collaboration easier (since @beatrycze-volk is working on fulltext features as well and has made some changes in `master` already), but it would temporarily break fulltext support in `master`.

In general, I suggest re-implementing the fulltext features in order to combine current ALTOv2 support with streamlined IIIF annotations support, new ALTOv3 support and the upcoming `solr_ocrhighlighting` support.
- I'd like to get rid of the JS ALTO parser and instead build a clean eID API which supports ALTO (v2 + v3) as well as IIIF annotations (and could be enhanced to support TEI).
- Fulltext parsing would then be done by the backend in a streamlined way (for indexing and for display) and the javascript in the  frontend would just have to handle the HTML provided by the backend (regardless of the fulltext source).
- Support for `solr_ocrhighlighting` could then be implemented in the backend encoding the image coordinates in the output HTML (i. e. as `data-*` attributes).

What do you think?

Resources/Public/Javascript/PageView/ImageManipulationControl.js
```diff
@@ -8,435 +8,235 @@
...
+
+        /**
+         * @type {Object}
+         * @private
+         */
+        this.filterDefaults_ = {
+            contrast: 1,
+            saturate: 1,
+            brightness: 1,
+            'hue-rotate': '0turn',
```
beatrycze-volk: Why hue-rotate filter is inside the ' ' and others not?

sebastian-meyer: Otherwise the `-` would be interpreted as operator.

Resources/Public/Javascript/PageView/PageView.js
```diff
@@ -8,692 +8,348 @@
...
+                target: this.settings.controlTargets.FullScreen || undefined,
+                tipLabel: this.settings.controlTitles.FullScreen || undefined
+            }));
         }
-        if (annotationControl !== undefined) {
-            $(imageManipulationControl).on("activate-imagemanipulation", $.proxy(annotationControl.deactivate, annotationControl));
-            $(annotationControl).on("activate-annotations", $.proxy(imageManipulationControl.deactivate, imageManipulationControl));
+        if (this.settings.controls.includes('ImageManipulation')) {
+            this.olx.controls.push(new ImageManipulationControl({
+                // autoOpen: false,
```
beatrycze-volk, line 116: Why this lines are outcommented?

sebastian-meyer: For every new map control I gave all possible options, but commented those out, which use the defaults anyway. That way every option is documented, but no unnecessary option parsing happens.

The reason is, that the new map controls can be used without Kitodo.Presentation as stand-alone OpenLayers controls as well. So I thought it would be a good idea to have all options documented. (I am still looking for the best way to release the controls separately.)

Do you suggest removing the outcommented options?

beatrycze-volk: No, they can stay. I just wanted to know why they are there ;)

beatrycze-volk:
> @beatrycze-volk + @albig: I am done now with everything except the fulltext stuff. I've tested this cross-browser and it works in everything except IE. A working test system can be found [here](https://vdl.opencultureconsulting.com). How should I go on? Merge this and implement fulltext support in a new branch? This would make collaboration easier (since @beatrycze-volk is working on fulltext features as well and has made some changes in `master` already), but it would temporarily break fulltext support in `master`.
> 

I would be ok with temporal break. The only question is: How long it would be?

> In general, I suggest re-implementing the fulltext features in order to combine current ALTOv2 support with streamlined IIIF annotations support, new ALTOv3 support and the upcoming `solr_ocrhighlighting` support.
> 
> * I'd like to get rid of the JS ALTO parser and instead build a clean eID API which supports ALTO (v2 + v3) as well as IIIF annotations (and could be enhanced to support TEI).

Would not be better if we first restore to work and next start to update from working state?

> * Fulltext parsing would then be done by the backend in a streamlined way (for indexing and for display) and the javascript in the  frontend would just have to handle the HTML provided by the backend (regardless of the fulltext source).

Sounds cool. Could you give more details on it?
> * Support for `solr_ocrhighlighting` could then be implemented in the backend encoding the image coordinates in the output HTML (i. e. as `data-*` attributes).
> 

Actually I started from implementing more in backend - some PHP classes are here (https://github.com/beatrycze-volk/kitodo-presentation/commit/f5969487b57a99a20858bdec14aac4558b0a5d2a)

> What do you think?

I'm fully for moving functionalities to backend and minimize amount of JavaScript processing.

sebastian-meyer:
I've further investigated how to refactor the fulltext feature.

First I analysed the current situation and identified the following issues and features we need to support:

- Currently the code for the fulltext feature is split between the `PageView` plugin and the `FulltextTool` of the `Toolbox` plugin (like most of the map features). This is messy and really bad for readability and maintenance.
- Fulltext has to be shown besides the image, correlating each text line with its counter-part in the image and vice-versa (by highlighting polygons while hovering).
- There needs to be a way to render the fulltext outside of the map extent, e. g. in the sidebar of the Zeitungsportal viewer. If it is rendered inside the map extent, the image width has to be used as a offset to place the text besides the image.

I am thinking of handling those issues and refactoring the fulltext feature like this:

- All code should be focused in the `PageView` plugin, implementing the map features mostly in Javascript as [OpenLayers map controls](https://openlayers.org/en/latest/apidoc/module-ol_control_Control-Control.html). I already reworked the `ImageManipulation` tool as well as the `Magnify` and `Rotate` features into OpenLayers map controls. I suggest reimplementing the fulltext feature as an OpenLayers map control in the same way (which already allows rendering outside the map extent).
- I suggest handling the fulltext as GeoJSON, thereby encoding each textblock, line and word as a `Feature` with its coordinates as `Polygon` or `LineString` geometry and - for words - its property `content` as the word itself. This way we could use the default OpenLayers `VectorLayer` with a `GeoJSON` source for visualization and even position the words exactly like in the image (see [this example](https://openlayers.org/en/latest/examples/street-labels.html) where a [GeoJSON text](https://openlayers.org/en/latest/examples/data/geojson/vienna-streets.geojson) layer is used to show street names).
- (We could even make using the image width as an offset configurable, thereby allowing the rendering of the text right on the image. This could be especially useful when introducing a slider which changes the layer's opacity.)
- The transformation into GeoJSON should be handled by an eID handler server-side. That way we could easily support different fulltext formats like ALTO and TEI simply by providing transformations for those formats (which is rather straight-forward since those XML formats are structurally similar to GeoJSON). There is no need anymore for format-specific implementations in the PageView itself.
- The current features for highlighting text passages by hovering could be realised by registering event handlers for the GeoJSON features of the OpenLayers map. OpenLayers already provides methods for event registration and dispatching.

Some advantages of this approach would be:

- We are mostly using features already provided by OpenLayers, minimizing the need for own code. This makes maintenance easier and should be more future-proof.
- The handling of GeoJSON is extensible, making future feature additions like showing geo-referenced maps, adding more TEI-features (e. g. for annotations), etc. easier.
- The resource-heavy transformation of fulltext source files is done server-side, circumventing potential performance issues with old clients or outdated browsers.

Before starting implementation I'd appreciate your feedback, @albig and @beatrycze-volk!

sebastian-meyer:
@albig, @beatrycze-volk: Any comments about this? I'd like to finish this soon, because since you are working on viewer details as well, this is getting harder and harder to merge without conflicts...

beatrycze-volk:
_All code should be focused in the PageView plugin, implementing the map features mostly in Javascript as OpenLayers map controls. I already reworked the ImageManipulation tool as well as the Magnify and Rotate features into OpenLayers map controls. I suggest reimplementing the fulltext feature as an OpenLayers map control in the same way (which already allows rendering outside the map extent)._

I think it is the great idea!

_We are mostly using features already provided by OpenLayers, minimizing the need for own code. This makes maintenance easier and should be more future-proof._

It sounds really good.

sebastian-meyer:
I am afraid I won't be able to finish this pull request any time soon. It would be a shame if all the work would just get dismissed, but on the other hand, the fulltext features are still missing (while the rest is finished and tested).

How should we handle this?

albig:
Sorry, I hadn't the time to look deepter into this. My goal is to merge the PR #587 first which is getting closer but still does not fully work [for my setup](https://github.com/albig/ddev-kitodo-presentation/tree/ocr-highlighting-plugin). This review work will be continued mid of July 2021.

After this, the OpenLayers PR would be the next. Maybe it's possible to split this PR into smaller parts, like wie tried with #587 to reduce the time for testing and review?

sebastian-meyer:
Splitting this pull request up is impossible, because you can't use two different versions of OpenLayers at the same time. As soon as I update OpenLayers the whole viewer breaks, so I have to rework everything else at the same time.

But: Reviewing this shouldn't be to complicated, because most of the changes only affect the viewer part. I don't recommend reviewing this as a line by line comparison, because it's a complete reimplementation and not just a set of changes to the old files. I think it's easier to ignore the old files and just review the new code.

I'd be happy to give you some introduction to the changes and some insights into my thought process. Maybe that would be helpful to make reviewing this easier.

albig:
Closing in favor of #763

Issue #502, opened 2020-04-08
Solr highlighting doesn't work with certain search terms.


albig:
## Description

On using the fulltext search in Kitodo.Presentation we experience the following behaviour:

- Search for "gut", or "Leipzig" returns results and highlighting textsnippets
- Search for "böse" or "Dresden" returns results but **no** highlighting textsnippets

## Reproduction

Steps to reproduce the behaviour:

1. Go to [Börsenblatt digital: Recherche](https://www.boersenblatt-digital.de/recherche)
2. Search for "gut", "böse", "Leipzig", "Dresden"

## Expected Behavior

If a search result is found, the textsnippet must be found, too.

## Environment

 - OS version: [Debian Linux 10.3]
 - RDBMS version: [MariaDB 10.3]
 - Apache Solr version: [e.g. 7.7]
 - TYPO3 version: [e.g. 9.5]
 - PHP version: [e.g. 7.3]

## Additional Context

The Solr-query which is performed by Kitodo.Presentation through [Solarium](https://github.com/solariumphp/solarium) is the following:

```
webapp=/solr path=/select params={json.nl=flat&hl=true&fl=uid,id,toplevel,thumbnail,page,type,title_tsi,volume_usu,author_tsi,place_tsi,year_usi,localtermsofuse_usi,useandreproduction_usi&start=0&sort=score+desc&fq=uid:15044&rows=10000&q=fulltext:(böse)+OR+toplevel:true&hl.useFastVectorHighlighter=true&omitHeader=true&hl.q=böse&hl.fl=fulltext&wt=json} hits=3 status=0 QTime=4
```
The search term is found twice inside: 

`q=fulltext:(böse)...` and `hl.q=böse`.

I assume, this is the answer why the VectorHighlighter does not work as expected. If you look at the query with `debugQuery=on` you see, that the initial search term gets modified to perform the search:

```
{
  "response":{"numFound":3,"start":0,"docs":[
      {
        "id":"15044PHYS_0047",
        "uid":15044,
        "page":47,
        "thumbnail":"https://digital.slub-dresden.de/data/kitodo/Brsfded_39946221X-19080219/Brsfded_39946221X-19080219_tif/jpegs/00000047.tif.small.jpg",
        "toplevel":false,
        "type":"page"},
      {
        "id":"15044PHYS_0035",
        "uid":15044,
        "page":35,
[...]

PR #587, opened 2021-02-11
Use Solr OCR Highlighting Plugin in Search in Document Plugin

beatrycze-volk:
The main aim of this PR is to implement  Solr OCR Highlighting Plugin inside the existing Search in Document Plugin.

Tasks:

- [x] adjust SOLR schema
- [x] remap index fields or possible usage of indexes with different schema than default one
- [x] remap input fields names to make query parameters configurable
- [x] add backend classes for handling results returned from SOLR index
- [x] update SearchInDocumentTool.php and file and its template to pass all needed parameters
- [x] rewrite SearchInDocument,js file to handle search and processing of the result
   - [x] display result snippets
   - [x] highlight searched word in the full text
   - [x] highlight searched word in the image
- [x] update documentation for TypoScript config

Changes which should be discussed:
- [x] Access-Control-Allow-Origin
Currently in PageViewProxy it is set. It causes problem when this header is also configured in .htaccess.
- [x] replace of submit button with standard function button in search form
For unknown reason submit form button with prevent page reload was not working for the partner, so it got changes to standard button which calls a method. It work for the partner now, but it has some implications for e.g. pressing Enter after writing search term simply submits form without executing search function. One of the possible solutions would be also map the key to the function, but it is not sure that there are no other implications not detected currently.

### DIFF
--- a/Classes/Common/Document.php
+++ b/Classes/Common/Document.php
@@ -652,7 +652,7 @@ abstract class Document
      */
     public abstract function getFullText($id);
 
-        /**
+    /**
      * This extracts the OCR full text for a physical structure node / IIIF Manifest / Canvas from an
      * XML full text representation (currently only ALTO). For IIIF manifests, ALTO documents have
      * to be given in the Canvas' / Manifest's "seeAlso" property.
@@ -700,6 +700,15 @@ abstract class Document
         return $fullText;
     }
 
+    /**
+     * Get format of the OCR full text
+     *
+     * @access private
+     *
+     * @param string $fileContent: content of the XML file
+     *
+     * @return string The format of the OCR full text
+     */
     private function getTextFormat($fileContent)
     {
         // Turn off libxml's error logging.
--- a/Classes/Common/Document.php
+++ b/Classes/Common/Document.php
@@ -670,7 +670,7 @@ abstract class Document
         // ... physical structure ...
         $this->_getPhysicalStructure();
         // ... and extension configuration.
-        $extConf = unserialize($GLOBALS['TYPO3_CONF_VARS']['EXT']['extConf'][self::$extKey]);
+        $extConf = GeneralUtility::makeInstance(ExtensionConfiguration::class)->get(self::$extKey);
         $fileGrpsFulltext = GeneralUtility::trimExplode(',', $extConf['fileGrpFulltext']);
         if (!empty($this->physicalStructureInfo[$id])) {
             while ($fileGrpFulltext = array_shift($fileGrpsFulltext)) {
--- a/Classes/Common/Document.php
+++ b/Classes/Common/Document.php
@@ -1291,10 +1291,14 @@ abstract class Document
         }
         // Add document to index.
         if ($core) {
-            //TODO: handling if this method returns failure
-            Indexer::add($this, $core);
+            //TODO: change return of this method to true on success and false on failure
+            $hasErrors = Indexer::add($this, $core);
+            if ($hasErrors) {
+                return false;
+            }
         } else {
             $this->logger->notice('Invalid UID "' . $core . '" for Solr core');
+            return false;
         }
         return true;
     }
--- a/Classes/Common/Document.php
+++ b/Classes/Common/Document.php
@@ -677,20 +677,9 @@ abstract class Document
             while ($fileGrpFulltext = array_shift($fileGrpsFulltext)) {
                 if (!empty($this->physicalStructureInfo[$id]['files'][$fileGrpFulltext])) {
                     // Get full text file.
-                    $file = GeneralUtility::getUrl($this->getFileLocation($this->physicalStructureInfo[$id]['files'][$fileGrpFulltext]));
-                    if ($file !== false) {
-                        // Turn off libxml's error logging.
-                        $libxmlErrors = libxml_use_internal_errors(true);
-                        // Disables the functionality to allow external entities to be loaded when parsing the XML, must be kept.
-                        $previousValueOfEntityLoader = libxml_disable_entity_loader(true);
-                        // Load XML from file.
-                        $rawTextXml = simplexml_load_string($file);
-                        // Reset entity loader setting.
-                        libxml_disable_entity_loader($previousValueOfEntityLoader);
-                        // Reset libxml's error logging.
-                        libxml_use_internal_errors($libxmlErrors);
-                        // Get the root element's name as text format.
-                        $textFormat = strtoupper($rawTextXml->getName());
+                    $fileContent = GeneralUtility::getUrl($this->getFileLocation($this->physicalStructureInfo[$id]['files'][$fileGrpFulltext]));
+                    if ($fileContent !== false) {
+                        $textFormat = $this->getTextFormat($fileContent);
                     } else {
                         $this->logger->warning('Couldn\'t load full text file for structure node @ID "' . $id . '"');
                         return $fullText;
@@ -704,105 +693,28 @@ abstract class Document
         }
         // Is this text format supported?
         // This part actually differs from previous version of indexed OCR
-        if (!empty($file) && !empty($this->formats[$textFormat])) {
-            if (!empty($this->formats[$textFormat]['class'])) {
-                $fullText = $file;
-            }
+        if (!empty($fileContent) && !empty($this->formats[$textFormat])) {
+            $fullText = $fileContent;
         } else {
             $this->logger->warning('Unsupported text format "' . $textFormat . '" in physical node with @ID "' . $id . '"');
         }
         return $fullText;
     }
 
-    /**
-     * This extracts the raw text for a physical structure node / IIIF Manifest / Canvas. Text might be
-     * given as ALTO for METS or as annotations or ALTO for IIIF resources. If IIIF plain text annotations
-     * with the motivation "painting" should be treated as full text representations, the extension has to be
-     * configured accordingly.
-     *
-     * @access public
-     *
-     * @abstract
-     *
-     * @param string $id: The @ID attribute of the physical structure node (METS) or the @id property
-     * of the Manifest / Range (IIIF)
-     *
-     * @return string The physical structure node's / IIIF resource's raw text
-     */
-    //TODO: check if this method is still needed somewhere, if not simply replace with getFullText
-    public abstract function getRawText($id);
-
-    /**
-     * This extracts the raw text for a physical structure node / IIIF Manifest / Canvas from an
-     * XML fulltext representation (currently only ALTO). For IIIF manifests, ALTO documents have
-     * to be given in the Canvas' / Manifest's "seeAlso" property.
-     *
-     * @param string $id: The @ID attribute of the physical structure node (METS) or the @id property
-     * of the Manifest / Range (IIIF)
-     *
-     * @return string The physical structure node's / IIIF resource's raw text from XML
-     */
-    protected function getRawTextFromXml($id)
+    private function getTextFormat($fileContent)
     {
-        $rawText = '';
-        // Load available text formats, ...
-        $this->loadFormats();
-        // ... physical structure ...
-        $this->_getPhysicalStructure();
-        // ... and extension configuration.
-        $extConf = GeneralUtility::makeInstance(ExtensionConfiguration::class)->get(self::$extKey);
-        $fileGrpsFulltext = GeneralUtility::trimExplode(',', $extConf['fileGrpFulltext']);
-        if (!empty($this->physicalStructureInfo[$id])) {
-            while ($fileGrpFulltext = array_shift($fileGrpsFulltext)) {
-                if (!empty($this->physicalStructureInfo[$id]['files'][$fileGrpFulltext])) {
-                    // Get fulltext file.
-                    $file = GeneralUtility::getUrl($this->getFileLocation($this->physicalStructureInfo[$id]['files'][$fileGrpFulltext]));
-                    if ($file !== false) {
-                        // Turn off libxml's error logging.
-                        $libxmlErrors = libxml_use_internal_errors(true);
-                        // Disables the functionality to allow external entities to be loaded when parsing the XML, must be kept.
-                        $previousValueOfEntityLoader = libxml_disable_entity_loader(true);
-                        // Load XML from file.
-                        $rawTextXml = simplexml_load_string($file);
-                        // Reset entity loader setting.
-                        libxml_disable_entity_loader($previousValueOfEntityLoader);
-                        // Reset libxml's error logging.
-                        libxml_use_internal_errors($libxmlErrors);
-                        // Get the root element's name as text format.
-                        $textFormat = strtoupper($rawTextXml->getName());
-                    } else {
-                        $this->logger->warning('Couldn\'t load fulltext file for structure node @ID "' . $id . '"');
-                        return $rawText;
-                    }
-                    break;
-                }
-            }
-        } else {
-            $this->logger->warning('Invalid structure node @ID "' . $id . '"');
-            return $rawText;
-        }
-        // Is this text format supported?
-        if (
-            !empty($rawTextXml)
-            && !empty($this->formats[$textFormat])
-        ) {
-            if (!empty($this->formats[$textFormat]['class'])) {
-                $class = $this->formats[$textFormat]['class'];
-                // Get the raw text from class.
-                if (
-                    class_exists($class)
-                    && ($obj = GeneralUtility::makeInstance($class)) instanceof FulltextInterface
-                ) {
-                    $rawText = $obj->getRawText($rawTextXml);
-                    $this->rawTextArray[$id] = $rawText;
-                } else {
-                    $this->logger->warning('Invalid class/method "' . $class . '->getRawText()" for text format "' . $textFormat . '"');
-                }
-            }
-        } else {
-            $this->logger->warning('Unsupported text format "' . $textFormat . '" in physical node with @ID "' . $id . '"');
-        }
-        return $rawText;
+        // Turn off libxml's error logging.
+        $libxmlErrors = libxml_use_internal_errors(true);
+        // Disables the functionality to allow external entities to be loaded when parsing the XML, must be kept.
+        $previousValueOfEntityLoader = libxml_disable_entity_loader(true);
+        // Load XML from file.
+        $rawTextXml = simplexml_load_string($fileContent);
+        // Reset entity loader setting.
+        libxml_disable_entity_loader($previousValueOfEntityLoader);
+        // Reset libxml's error logging.
+        libxml_use_internal_errors($libxmlErrors);
+        // Get the root element's name as text format.
+        return strtoupper($rawTextXml->getName());
     }
 
     /**
--- a/Classes/Common/Document.php
+++ b/Classes/Common/Document.php
@@ -570,9 +570,8 @@ abstract class Document
             if (!empty($extConf['caching'])) {
                 Helper::saveToSession(self::$registry, get_class($instance));
             }
-            $instance->logger = GeneralUtility::makeInstance(LogManager::class)->getLogger(__CLASS__);
+            $instance->logger = GeneralUtility::makeInstance(LogManager::class)->getLogger(get_class($instance));
         }
-
         // Return new instance.
         return $instance;
     }
--- a/Classes/Common/Document.php
+++ b/Classes/Common/Document.php
@@ -570,8 +570,9 @@ abstract class Document
             if (!empty($extConf['caching'])) {
                 Helper::saveToSession(self::$registry, get_class($instance));
             }
+            $instance->logger = GeneralUtility::makeInstance(LogManager::class)->getLogger(__CLASS__);
         }
-        $instance->logger = GeneralUtility::makeInstance(LogManager::class)->getLogger(get_class($instance));
+
         // Return new instance.
         return $instance;
     }
@@ -638,6 +639,82 @@ abstract class Document
     }
 
     /**
+     * This extracts the OCR full text for a physical structure node / IIIF Manifest / Canvas. Text might be
+     * given as ALTO for METS or as annotations or ALTO for IIIF resources.
+     *
+     * @access public
+     *
+     * @abstract
+     *
+     * @param string $id: The @ID attribute of the physical structure node (METS) or the @id property
+     * of the Manifest / Range (IIIF)
+     *
+     * @return string The OCR full text
+     */
+    public abstract function getFullText($id);
+
+        /**
+     * This extracts the OCR full text for a physical structure node / IIIF Manifest / Canvas from an
+     * XML full text representation (currently only ALTO). For IIIF manifests, ALTO documents have
+     * to be given in the Canvas' / Manifest's "seeAlso" property.
+     *
+     * @param string $id: The @ID attribute of the physical structure node (METS) or the @id property
+     * of the Manifest / Range (IIIF)
+     *
+     * @return string The OCR full text
+     */
+    protected function getFullTextFromXml($id)
+    {
+        $fullText = '';
+        // Load available text formats, ...
+        $this->loadFormats();
+        // ... physical structure ...
+        $this->_getPhysicalStructure();
+        // ... and extension configuration.
+        $extConf = unserialize($GLOBALS['TYPO3_CONF_VARS']['EXT']['extConf'][self::$extKey]);
+        $fileGrpsFulltext = GeneralUtility::trimExplode(',', $extConf['fileGrpFulltext']);
+        if (!empty($this->physicalStructureInfo[$id])) {
+            while ($fileGrpFulltext = array_shift($fileGrpsFulltext)) {
+                if (!empty($this->physicalStructureInfo[$id]['files'][$fileGrpFulltext])) {
+                    // Get full text file.
+                    $file = GeneralUtility::getUrl($this->getFileLocation($this->physicalStructureInfo[$id]['files'][$fileGrpFulltext]));
+                    if ($file !== false) {
+                        // Turn off libxml's error logging.
+                        $libxmlErrors = libxml_use_internal_errors(true);
+                        // Disables the functionality to allow external entities to be loaded when parsing the XML, must be kept.
+                        $previousValueOfEntityLoader = libxml_disable_entity_loader(true);
+                        // Load XML from file.
+                        $rawTextXml = simplexml_load_string($file);
+                        // Reset entity loader setting.
+                        libxml_disable_entity_loader($previousValueOfEntityLoader);
+                        // Reset libxml's error logging.
+                        libxml_use_internal_errors($libxmlErrors);
+                        // Get the root element's name as text format.
+                        $textFormat = strtoupper($rawTextXml->getName());
+                    } else {
+                        $this->logger->warning('Couldn\'t load full text file for structure node @ID "' . $id . '"');
+                        return $fullText;
+                    }
+                    break;
+                }
+            }
+        } else {
+            $this->logger->warning('Invalid structure node @ID "' . $id . '"');
+            return $fullText;
+        }
+        // Is this text format supported?
+        // This part actually differs from previous version of indexed OCR
+        if (!empty($file) && !empty($this->formats[$textFormat])) {
+            if (!empty($this->formats[$textFormat]['class'])) {
+                $fullText = $file;
+            }
+        } else {
+            $this->logger->warning('Unsupported text format "' . $textFormat . '" in physical node with @ID "' . $id . '"');
+        }
+        return $fullText;
+    }
+
+    /**
      * This extracts the raw text for a physical structure node / IIIF Manifest / Canvas. Text might be
      * given as ALTO for METS or as annotations or ALTO for IIIF resources. If IIIF plain text annotations
      * with the motivation "painting" should be treated as full text representations, the extension has to be
@@ -652,6 +729,7 @@ abstract class Document
      *
      * @return string The physical structure node's / IIIF resource's raw text
      */
+    //TODO: check if this method is still needed somewhere, if not simply replace with getFullText
     public abstract function getRawText($id);
 
     /**
@@ -1301,6 +1379,7 @@ abstract class Document
         }
         // Add document to index.
         if ($core) {
+            //TODO: handling if this method returns failure
             Indexer::add($this, $core);
         } else {
             $this->logger->notice('Invalid UID "' . $core . '" for Solr core');
--- a/Classes/Common/Document.php
+++ b/Classes/Common/Document.php
@@ -28,7 +28,7 @@ use Ubl\Iiif\Tools\IiifHelper;
  * @package TYPO3
  * @subpackage dlf
  * @access public
- * @property-write int $cPid This holds the PID for the configuration
+ * @property int $cPid This holds the PID for the configuration
  * @property-read bool $hasFulltext Are there any fulltext files available?
  * @property-read string $location This holds the documents location
  * @property-read array $metadataArray This holds the documents' parsed metadata array
@@ -1001,10 +1001,11 @@ abstract class Document
      *
      * @param int $pid: The PID of the saved record
      * @param int $core: The UID of the Solr core for indexing
+     * @param int|string $owner: UID or index_name of owner to set while indexing
      *
      * @return bool true on success or false on failure
      */
-    public function save($pid = 0, $core = 0)
+    public function save($pid = 0, $core = 0, $owner = null)
     {
         if (\TYPO3_MODE !== 'BE') {
             $this->logger->error('Saving a document is only allowed in the backend');
@@ -1130,53 +1131,57 @@ abstract class Document
             ->getQueryBuilderForTable('tx_dlf_libraries');
 
         // Get UID for owner.
-        $owner = !empty($metadata['owner'][0]) ? $metadata['owner'][0] : 'default';
-
-        $result = $queryBuilder
-            ->select('tx_dlf_libraries.uid AS uid')
-            ->from('tx_dlf_libraries')
-            ->where(
-                $queryBuilder->expr()->eq('tx_dlf_libraries.pid', intval($pid)),
-                $queryBuilder->expr()->eq('tx_dlf_libraries.index_name', $queryBuilder->expr()->literal($owner)),
-                Helper::whereExpression('tx_dlf_libraries')
-            )
-            ->setMaxResults(1)
-            ->execute();
-
-        if ($resArray = $result->fetch()) {
-            $ownerUid = $resArray['uid'];
-        } else {
-            // Insert new library.
-            $libNewUid = uniqid('NEW');
-            $libData['tx_dlf_libraries'][$libNewUid] = [
-                'pid' => $pid,
-                'label' => $owner,
-                'index_name' => $owner,
-                'website' => '',
-                'contact' => '',
-                'image' => '',
-                'oai_label' => '',
-                'oai_base' => '',
-                'opac_label' => '',
-                'opac_base' => '',
-                'union_label' => '',
-                'union_base' => '',
-            ];
-            $substUid = Helper::processDBasAdmin($libData);
-            // Add new library's UID.
-            $ownerUid = $substUid[$libNewUid];
-            if (!(\TYPO3_REQUESTTYPE & \TYPO3_REQUESTTYPE_CLI)) {
-                Helper::addMessage(
-                    htmlspecialchars(sprintf(Helper::getMessage('flash.newLibrary'), $owner, $ownerUid)),
-                    Helper::getMessage('flash.attention', true),
-                    \TYPO3\CMS\Core\Messaging\FlashMessage::INFO,
-                    true
-                );
-            }
+        if (empty($owner)) {
+            $owner = empty($metadata['owner'][0]) ? $metadata['owner'][0] : 'default';
         }
-        $metadata['owner'][0] = $ownerUid;
+        if (!MathUtility::canBeInterpretedAsInteger($owner)) {
+            $result = $queryBuilder
+                ->select('tx_dlf_libraries.uid AS uid')
+                ->from('tx_dlf_libraries')
+                ->where(
+                    $queryBuilder->expr()->eq('tx_dlf_libraries.pid', intval($pid)),
+                    $queryBuilder->expr()->eq('tx_dlf_libraries.index_name', $queryBuilder->expr()->literal($owner)),
+                    Helper::whereExpression('tx_dlf_libraries')
+                )
+                ->setMaxResults(1)
+                ->execute();
+
+            if ($resArray = $result->fetch()) {
+                $ownerUid = $resArray['uid'];
+            } else {
+                // Insert new library.
+                $libNewUid = uniqid('NEW');
+                $libData['tx_dlf_libraries'][$libNewUid] = [
+                    'pid' => $pid,
+                    'label' => $owner,
+                    'index_name' => $owner,
+                    'website' => '',
+                    'contact' => '',
+                    'image' => '',
+                    'oai_label' => '',
+                    'oai_base' => '',
+                    'opac_label' => '',
+                    'opac_base' => '',
+                    'union_label' => '',
+                    'union_base' => '',
+                ];
+                $substUid = Helper::processDBasAdmin($libData);
+                // Add new library's UID.
+                $ownerUid = $substUid[$libNewUid];
+                if (!(\TYPO3_REQUESTTYPE & \TYPO3_REQUESTTYPE_CLI)) {
+                    Helper::addMessage(
+                        htmlspecialchars(sprintf(Helper::getMessage('flash.newLibrary'), $owner, $ownerUid)),
+                        Helper::getMessage('flash.attention', true),
+                        \TYPO3\CMS\Core\Messaging\FlashMessage::INFO,
+                        true
+                    );
+                }
+            }
+            $owner = $ownerUid;
+        }
+        $metadata['owner'][0] = $owner;
         // Get UID of parent document.
-        $partof = $this->getParentDocumentUidForSaving($pid, $core);
+        $partof = $this->getParentDocumentUidForSaving($pid, $core, $owner);
         // Use the date of publication or title as alternative sorting metric for parts of multi-part works.
         if (!empty($partof)) {
             if (
@@ -1322,7 +1327,19 @@ abstract class Document
      *
      * @return int The parent document's id.
      */
-    protected abstract function getParentDocumentUidForSaving($pid, $core);
+    protected abstract function getParentDocumentUidForSaving($pid, $core, $owner);
+
+    /**
+     * This returns $this->cPid via __get()
+     *
+     * @access protected
+     *
+     * @return int The PID of the metadata definitions
+     */
+    protected function _getCPid()
+    {
+        return $this->cPid;
+    }
 
     /**
      * This returns $this->hasFulltext via __get()
--- a/Classes/Common/Document.php
+++ b/Classes/Common/Document.php
@@ -693,11 +693,29 @@ abstract class Document
         // Is this text format supported?
         // This part actually differs from previous version of indexed OCR
         if (!empty($fileContent) && !empty($this->formats[$textFormat])) {
-            $fullText = $fileContent;
+            $fullText = $this->getFullTextWithoutImages($fileContent);
         } else {
             $this->logger->warning('Unsupported text format "' . $textFormat . '" in physical node with @ID "' . $id . '"');
         }
         return $fullText;
+    }
+
+     /**
+     * Get content of the OCR full text file without images
+     *
+     * @access private
+     *
+     * @param string $fileContent: content of the XML file
+     *
+     * @return string The content of the OCR full text file without images
+     */
+    private function getFullTextWithoutImages($fileContent)
+    {
+        $objectXml = $this->getFullTextAsObjectXML($fileContent);
+        if (isset($objectXml->Layout->Page->PrintSpace->Illustration)) {
+            unset($objectXml->Layout->Page->PrintSpace->Illustration);
+        }
+        return $objectXml->asXML();
     }
 
     /**
@@ -711,18 +729,33 @@ abstract class Document
      */
     private function getTextFormat($fileContent)
     {
+        // Get the root element's name as text format.
+        return strtoupper($this->getFullTextAsObjectXML($fileContent)->getName());
+    }
+
+    /**
+     * Get content of the OCR full text file
+     *
+     * @access private
+     *
+     * @param string $fileContent: content of the XML file
+     *
+     * @return \SimpleXMLElement content of the XML file as object
+     */
+    private function getFullTextAsObjectXML($fileContent)
+    {
         // Turn off libxml's error logging.
         $libxmlErrors = libxml_use_internal_errors(true);
         // Disables the functionality to allow external entities to be loaded when parsing the XML, must be kept.
         $previousValueOfEntityLoader = libxml_disable_entity_loader(true);
         // Load XML from file.
-        $rawTextXml = simplexml_load_string($fileContent);
+        $objectXml = simplexml_load_string($fileContent);
         // Reset entity loader setting.
         libxml_disable_entity_loader($previousValueOfEntityLoader);
         // Reset libxml's error logging.
         libxml_use_internal_errors($libxmlErrors);
-        // Get the root element's name as text format.
-        return strtoupper($rawTextXml->getName());
+        // Get the root element
+        return $objectXml;
     }
 
     /**
--- a/Classes/Common/DocumentList.php
+++ b/Classes/Common/DocumentList.php
@@ -273,7 +273,7 @@ class DocumentList implements \ArrayAccess, \Countable, \Iterator, LoggerAwareIn
         // Set additional query parameters.
         // Set reasonable limit for safety reasons.
         // We don't expect to get more than 10.000 hits per UID.
-        $query->setStart(0)->setRows(1000);
+        $query->setStart(0)->setRows(10000);
         // Take over existing filter queries.
         $filterQueries = isset($this->metadata['options']['params']['filterquery']) ? $this->metadata['options']['params']['filterquery'] : [];
         // Extend filter query to get all documents with the same UID.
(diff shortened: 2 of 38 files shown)


albig:
What do I have to do to install/activate the OCR Highlighting Plugin? Does the code work, if it's not installed?

beatrycze-volk:
> What do I have to do to install/activate the OCR Highlighting Plugin? Does the code work, if it's not installed?

https://dbmdz.github.io/solr-ocrhighlighting/installation/

I would assume that it doesn't work if not installed. I have only tested with SOLR index which had the plugin installed.

albig:
> > What do I have to do to install/activate the OCR Highlighting Plugin? Does the code work, if it's not installed?
> 
> https://dbmdz.github.io/solr-ocrhighlighting/installation/
> 
> I would assume that it doesn't work if not installed. I have only tested with SOLR index which had the plugin installed.

Ok. Than we have to include this information to the documentation. 

And I will setup a testsystem with this Solr plugin. May take some days.

beatrycze-volk:
Should I mark it as ready to merge or we also decide that some changes can't be merged in and I will create new branch in which I will implement new stuff without changing old ones (I mean way of submitting the form and cross origin access)?

albig:
> https://dbmdz.github.io/solr-ocrhighlighting/installation/
> 
> I would assume that it doesn't work if not installed. I have only tested with SOLR index which had the plugin installed.

You are right. It doesn't work out-of-the box. The search plugin still finds no snippets of course.The searchInDocument fails with exception.

I still don't like the idea to make the solr-ocrhighlighting plugin obligatory. And I will investigate howto upgrade an existing installation.

Classes/Common/SolrSearchResult/ResultDocument.php
```diff
@@ -0,0 +1,259 @@
...
+     *
+     * @return void
+     */
+    public function __construct($record, $highlighting, $fields) {
+        $this->id = $record[$fields['id']];
+        $this->uid = $record[$fields['uid']];
+        $this->page = $record[$fields['page']];
+
+        $highlightingForRecord = $highlighting[$record[$fields['id']]][$fields['fulltext']];
+        $this->snippetsForRecord = $highlightingForRecord['snippets'];
```
albig: This maybe null and the getArrayByIndex() function will fail late.
```suggestion
        $this->snippetsForRecord = is_array($highlightingForRecord['snippets']) ? $highlightingForRecord['snippets'] : [];
```

Classes/Common/SolrSearchResult/ResultDocument.php
```diff
@@ -0,0 +1,259 @@
...
+     */
+    public function getHighlightsIds() {
+        $highlightsIds = [];
+        foreach ($this->highlights as $highlight) {
+            array_push($highlightsIds, $highlight->getId());
+        }
+        return $highlightsIds;
+    }
+
+    private function parseSnippets() {
```
albig: Please add PHPDoc comments to private functions, too.

Configuration/ApacheSolr/configsets/dlf/conf/solrconfig.xml
```diff
@@ -754,6 +756,21 @@
...
+    highlighting component, but **after** the query component.
+  -->
+  <requestHandler name="/select" class="solr.SearchHandler">
+      <arr name="components">
+          <str>query</str>
+          <str>ocrHighlight</str>
+          <str>highlight</str>
+      </arr>
+  </requestHandler>
+
```
albig: The registration of the searchComponent ist missing in this solrconfig.xml. See https://dbmdz.github.io/solr-ocrhighlighting/installation/

Without this, it's not possible to create a core.

```
<!-- Add a new named search component that takes care of highlighting OCR field values. -->
  <searchComponent class="de.digitalcollections.solrocr.solr.OcrHighlightComponent"
                   name="ocrHighlight" />
```

albig (changes requested):
I didn't manage to get the Solr working with
* Solr 8.8
* solr-ocrhighlighting 0.5.0 (current release)
* schema.xml and solrconfig.xml from this PR
* updated extension configuration for solr (using default values)

On indexing a document, I end up with the following error

<pre>
org.apache.solr.common.SolrException: Exception writing document id 36LOG_0003 to the index; possible analysis error.
	at org.apache.solr.update.DirectUpdateHandler2.addDoc(DirectUpdateHandler2.java:256)
	at org.apache.solr.update.processor.RunUpdateProcessorFactory$RunUpdateProcessor.processAdd(RunUpdateProcessorFactory.java:73)
...
Caused by: java.lang.RuntimeException: Could not parse pointer: 
	at de.digitalcollections.solrocr.model.SourcePointer.parse(SourcePointer.java:81)
	at de.digitalcollections.solrocr.lucene.filters.ExternalUtf8ContentFilterFactory.create(ExternalUtf8ContentFilterFactory.java:42)
	at org.apache.solr.analysis.TokenizerChain.initReader(TokenizerChain.java:97)
...
</pre>

Please check again and tell me your system setup to reproduce.

Classes/Common/SolrSearchResult/ResultDocument.php
beatrycze-volk: Shouldn't this ternary operator have something between ?: or is it something specific to is_array and it doesn't need assignment?

albig: I think, you're right. This would return "true" in case of a valid array. I've updated my suggestion.

Classes/Plugin/Tools/SearchInDocumentTool.php
```diff
@@ -113,14 +105,68 @@ public function main($content, $conf)
...
+     *
+     * @access protected
+     *
+     * @return string with current document id
+     */
+    protected function getCurrentDocumentId()
+    {
+        $id = $this->doc->uid;
+
+        if (!empty($this->conf['documentIdUrlSchema'])) {
```
albig, line 160: What's the purpose of this? Could you please provide an example string to verify the string splitting magic below?

beatrycze-volk: Sometimes id of the document is actually not an id but a link to xml file. Then to search in the index the full link can't be used so it is needed to extract real id out of the full link.

Example:
https://dev-ddb.fiz-karlsruhe.de/api/items/F45KF24U7UIIVCUJJBV334PD7IPE7JBW/source/record -> https://dev-ddb.fiz-karlsruhe.de/api/items/*id*/source/record

sebastian-meyer: I think we need to distinguish more cases here:
- The normal case would be documents which are properly indexed in the TYPO3 database. Those don't need any special treatment for their ID anyway.
- The "DFG-Viewer" case would be documents which are not indexed in the database **nor in the Solr index**. Normally those can be identified by the fact that they have an URL as ID.
- Then there is the special "Zeitungsportal" case where we have documents, which are not indexed in the database, but are indexed in the Solr index. Those have an URL as ID as well, but need special treatment to extract the indexed ID from the URL.

I suggest using the TYPO3 hook mechanism to handle all those cases:
- If a document has a numerical ID, no further treatment is necessary.
- If a document has a string ID, this string is processed through a hook.
  - If the hook returns an integer, the document is treated as indexed.
  - If the hook returns the string unchanged, the document is considered unindexed and ignored when searching. (This would be the default if no hook is registered, e. g. in "DFG-Viewer mode".)

The advantages would be:
- We don't need to add Zeitungportal-specific logic to the main extension. Instead the hook can be part of the supplemental Zeitungsportal extension.
- The solution is extensible, because you can register more than one hook to handle different kinds of identifiers. So anybody with a setup of a pre-existing Solr index could map their document IDs to the index IDs and re-use their Solr index. (This could even be interesting for SLUB, when we think about consolidating our different indices - CMR, Find, Kitodo, ...)
- Processing the ID string wouldn't be limited to splitting a string, because a hook could include more complex logic like regular expressions or even database lookups.

beatrycze-volk: > I suggest using the TYPO3 hook mechanism to handle all those cases:

Do you mean https://typo3.org/article/how-to-use-existing-hooks-in-your-own-extension ?

> If a document has a string ID, this string is processed through a hook.
> If the hook returns an integer, the document is treated as indexed.
> If the hook returns the string unchanged, the document is considered unindexed and ignored when searching. (This would be the default if no hook is registered, e. g. in "DFG-Viewer mode".)

But how to convert string ID to an integer?

sebastian-meyer: We actually already have some hooks in Kitodo.Presentation. A good example is the way how we extract the record ID from METS:
Normally we would take `/mets:mets[@OBJID]` as the document's unique record identifier: https://github.com/kitodo/kitodo-presentation/blob/f12eede1596112dc359678768e9a9463f0791788/Classes/Common/MetsDocument.php#L147-L150
But if this attribute is not set (which is the case for old versions of Kitodo.Production), we use hooks to have a fallback. The hook is registered in `ext_localconf.php`: https://github.com/kitodo/kitodo-presentation/blob/f12eede1596112dc359678768e9a9463f0791788/ext_localconf.php#L219 and called in `Classes/Common/MetsDocument.php`: https://github.com/kitodo/kitodo-presentation/blob/f12eede1596112dc359678768e9a9463f0791788/Classes/Common/MetsDocument.php#L151-L158
The hook itself can be found in `Classes/Hooks/KitodoProductionHacks.php`: https://github.com/kitodo/kitodo-presentation/blob/f12eede1596112dc359678768e9a9463f0791788/Classes/Hooks/KitodoProductionHacks.php#L37-L68

albig: Hook is a possibility. I think the modern way would be to make some "middleware". But I haven't done this before. Hook support itself seem to be not deprecated, although lots of core hooks are.

The DFG-Viewer case is not relevant in the SearchInDocument tool. This tool won't work with it anyway. The DFG-Viewer has it's own SRU search plugin, which requires a SRU server of course.

I suggest to first test and documentate this PR including indexing (doesn't work yet) and searching for Kitodo.Presentation alone. If this works, we can discuss further at this point.

beatrycze-volk:
@albig: Is your error similar to something like that:

`org.apache.solr.common.SolrException: Exception writing document id ocrdoc-1 to the index; possible analysis error: Illegal char <:> at index 5: https://digital.slub-dresden.de/data/kitodo/TheDarea_416971482-19100223/TheDarea_416971482-19100223_ocr/00000001.xml
	at org.apache.solr.update.DirectUpdateHandler2.addDoc(DirectUpdateHandler2.java:247)
	at org.apache.solr.update.processor.RunUpdateProcessor.processAdd(RunUpdateProcessorFactory.java:76)`

?

After research I have found out that indexing when XML files are stored in data folder of the core works correctly but when they are passed from external URL the error appears.

I have checked in the GItHub page of the plugin and found https://github.com/dbmdz/solr-ocrhighlighting/issues/49:

> Currently we only support loading field values/OCR fragments from the file system.

Classes/Common/SolrSearchResult/ResultDocument.php
beatrycze-volk: Added

beatrycze-volk:
I'm sorry for this few rebase pushes, I have made a bit of mess and needed to fix it. Now all commits should be correct.

albig:
> @albig: Is your error similar to something like that:
> 
> `org.apache.solr.common.SolrException: Exception writing document id ocrdoc-1 to the index; possible analysis error: Illegal char <:> at index 5: https://digital.slub-dresden.de/data/kitodo/TheDarea_416971482-19100223/TheDarea_416971482-19100223_ocr/00000001.xml at org.apache.solr.update.DirectUpdateHandler2.addDoc(DirectUpdateHandler2.java:247) at org.apache.solr.update.processor.RunUpdateProcessor.processAdd(RunUpdateProcessorFactory.java:76)`
> 
> ?
> 
> After research I have found out that indexing when XML files are stored in data folder of the core works correctly but when they are passed from external URL the error appears.
> 
> I have checked in the GItHub page of the plugin and found [dbmdz/solr-ocrhighlighting#49](https://github.com/dbmdz/solr-ocrhighlighting/issues/49):
> 
> > Currently we only support loading field values/OCR fragments from the file system.

Yes, I think this was the error I've seen. So we have to wait for the fix in the ocrhighlighting plugin?

beatrycze-volk:
> Yes, I think this was the error I've seen. So we have to wait for the fix in the ocrhighlighting plugin?

Yes, of course if they are going to fix it actually. That's not so sure after reading the conversation in this issue.

Other possiblity is to store content of the xml files instead pointers to those files.

Classes/Common/Document.php
```diff
@@ -48,6 +49,14 @@
  */
 abstract class Document
 {
+    /**
+     * This holds the logger
+     *
+     * @var LogManager
```
albig: Good idea, but would you draw the changes to LogManager to a separate pull request? This is not related to the ocr highlighting.

Classes/Common/Document.php
```diff
@@ -667,54 +676,45 @@ protected function getRawTextFromXml($id)
...
+            $fullText = $fileContent;
         } else {
-            Helper::devLog('Unsupported text format "' . $textFormat . '" in physical node with @ID "' . $id . '"', DEVLOG_SEVERITY_WARNING);
+            $this->logger->warning('Unsupported text format "' . $textFormat . '" in physical node with @ID "' . $id . '"');
         }
-        return $rawText;
+        return $fullText;
+    }
+
+    private function getTextFormat($fileContent)
```
albig, line 730: Don't forget the method documentation.

albig (changes requested):
On my setup, the current state of the pull request does index all pages. Inside the fulltext field, the complete ALTO XML is saved. Not sure, that this is what we want.

But anyway:

* there is no toplevel document indexed
* that's why the collection plugin doesn't work
* that's why the search plugin doesn't work
* that's why the searchInDocument plugin doesn't work

So, maybe my setup is broken and/or I am missing something. We should look together on this work.

Solr Admin

beatrycze-volk:
>  Inside the fulltext field, the complete ALTO XML is saved. Not sure, that this is what we want.
What do you mean? OCR is determined out of this content.

> there is no toplevel document indexed

That's not surprise, I have commented out indexing of documents without OCR as they were causing error mentioned in #502.

> that's why the searchInDocument plugin doesn't work

On my local environment with 3 indexed documents it works. By searching and finding snippets I have tested if part with indexing documents containing OCR works.

beatrycze-volk:
The last two Codacy errors can't be fixed: _Identifier 'tx_dlf_viewer' is not in camel case._

Classes/Common/Document.php
```diff
@@ -650,81 +648,72 @@ public function getPhysicalPage($logicalPage)
...
     {
-        $rawText = '';
+        $fullText = '';
         // Load available text formats, ...
         $this->loadFormats();
         // ... physical structure ...
         $this->_getPhysicalStructure();
         // ... and extension configuration.
-        $extConf = GeneralUtility::makeInstance(ExtensionConfiguration::class)->get(self::$extKey);
+        $extConf = unserialize($GLOBALS['TYPO3_CONF_VARS']['EXT']['extConf'][self::$extKey]);
```
albig: Why do you revert to the deprecated syntax?

Classes/Common/Document.php
albig: Ping.

Classes/Plugin/Tools/SearchInDocumentTool.php
albig: Please add at least a sample of the documentIdUrlSchema setting. Or is this part of the documentation?

Documentation/Plugins/Index.rst
```diff
@@ -1087,7 +1087,11 @@ The fulltext is fetched and rendered by JavaScript into the `<div id="tx-dlf-ful
 
 
 Search in Document Tool
+<<<<<<< HEAD
```
albig: Please solve the merge conflict.

Resources/Public/Javascript/PageView/PageView.js
```diff
@@ -349,11 +358,18 @@ dlfViewer.prototype.displayHighlightWord = function() {
...
             fulltextData = dlfFullTextUtils.fetchFullTextDataFromServer(this.fulltexts[0].url, this.images[0]),
             fulltextDataImageTwo = undefined;
 
+        if(this.highlightWords != null) {
+            value = this.highlightWords;
+        } else {
+            value = urlParams[param];
+        }
+
+        var values = decodeURIComponent(value).split(';'),
```
albig:
```suggestion
        var values = decodeURIComponent(value).split(';');
```
This causes unfortunately an JavaScript exception which breaks the Pageview.

Resources/Public/Javascript/PageView/SearchInDocument.js
```diff
@@ -110,25 +318,22 @@ $(document).ready(function() {
...
-    $('#tx-dlf-search-in-document-clearing').click(function() {
+
+     // clearing button
+     $('#tx-dlf-search-in-document-clearing').click(function() {
         $('#tx-dlf-search-in-document-results ul').remove();
         $('.results-active-indicator').remove();
         $('#tx-dlf-search-in-document-query').val('');
     });
+
+    triggerSearchAfterHitLoad();
```
albig, line 328: This will be called on the listview, too and will end up in an JavaScript exception

```
Uncaught TypeError: Cannot read property 'split' of undefined
    at getBaseUrl (merged-0478f68d822d724a9d8e6d53edf60223.js?1625082852:4207)
    at triggerSearchAfterHitLoad (merged-0478f68d822d724a9d8e6d53edf60223.js?1625082852:4385)
    at HTMLDocument.<anonymous> (merged-0478f68d822d724a9d8e6d53edf60223.js?1625082852:4466)
```

albig (changes requested):
It works better but is still a little surprising:

* documents with emtpy titles won't index. Most newspaper issues have no title field set.
* search-in-document is quite funny
   * opening the search form inserts some number (page numer?) and executes a search --> why?
   * the search itself seems to work
   * clicking on the search result doesn't change the page and has IMHO incorrect URLs like `https://ddev-kitodo-presentation.ddev.site/workview?=4&=4&=4&hl=14_23_1577_1754%3B14_23_688_1196&tx_dlf%5Bid%5D=3&tx_dlf%5Bpage%5D=3`
Digitale Sammlungen  Allgemeine Zeitschrift für Textil-Industrie

Classes/Common/Document.php
beatrycze-volk: My changes were written before new syntax was introduced. I will use new syntax :)

Resources/Public/Javascript/PageView/SearchInDocument.js
beatrycze-volk: Is SearchInDocument used also in list view?

albig: Yes. All JavaScript ist loaded. That's what we changed recently: https://github.com/kitodo/kitodo-presentation/blob/master/Configuration/TypoScript/setup.txt

beatrycze-volk: Ok, so I have added check for undefined before trying to split base url.

Documentation/Plugins/Index.rst
beatrycze-volk: Fixed

beatrycze-volk:
> opening the search form inserts some number (page numer?) and executes a search --> why?

One of the requirements was that after moving from hit list of found documents to the certain document, it should execute this search again inside the document to get all found snippets in this document. The inserted value comes from URL parameter defined in configuration`plugin.tx_dlf_searchindocumenttool.queryInputName`. Default value for this configuration is  `tx_dlf[query]`. Could you check if you have something set up for this config?

> clicking on the search result doesn't change the page and has IMHO incorrect URLs like https://ddev-kitodo-presentation.ddev.site/workview?=4&=4&=4&hl=14_23_1577_1754%3B14_23_688_1196&tx_dlf%5Bid%5D=3&tx_dlf%5Bpage%5D=3

I would suspect that configuration`plugin.tx_dlf_searchindocumenttool.queryInputName` has empty value and it causes this weird inserted parameters.

I will add check for empty values to avoid such a situation.

Classes/Common/Document.php
albig: Ping again.

Classes/Plugin/Tools/SearchInDocumentTool.php
albig: This should be already part of #630. But I still need some (inline) documentation about what is going on here, why we are doing this and what string you expect in documentIdUrlSchema.

Classes/Plugin/Tools/SearchInDocumentTool.php
```diff
@@ -95,7 +95,7 @@ public function main($content, $conf)
 
         // Fill markers.
         $markerArray = [
-            '###ACTION_URL###' => $this->getActionUrl(),
+            '###ACTION_URL###' => getActionUrl(),
```
albig:
```suggestion
            '###ACTION_URL###' => $this->getActionUrl(),
```

Documentation/Plugins/Index.rst
```diff
@@ -1087,7 +1087,7 @@ The fulltext is fetched and rendered by JavaScript into the `<div id="tx-dlf-ful
 
 
 Search in Document Tool
-^^^^^^^^^^^^^^^^^^^^^^^
+
```
albig: Why did you remove the heading format?

Resources/Public/Javascript/PageView/SearchInDocument.js
```diff
@@ -43,17 +43,233 @@ function resetStart() {
...
         event.preventDefault();
+
         $('#tx-dlf-search-in-document-loading').show();
         $('#tx-dlf-search-in-document-clearing').hide();
         $('#tx-dlf-search-in-document-button-next').hide();
         $('#tx-dlf-search-in-document-button-previous').hide();
         // Send the data using post
         $.post(
-            "/",
+            "https://sdvtypo3ddbzeitungsportaldev.slub-dresden.de/",
```
albig: No, not really ;-)

albig (changes requested):
Please have a look at the comments. This is not yet working.

Resources/Public/Javascript/PageView/SearchInDocument.js
beatrycze-volk: Yeah, it was specific for DDB, I will revert it ;)

Documentation/Plugins/Index.rst
beatrycze-volk: Restored, it looks that it was lost in merge conflict.

Classes/Common/Document.php
beatrycze-volk: I'm not sure what is wrong with displaying of this part of the code in the Files changed view. There comment is missing, but in fact it is already added.

albig (changes requested):
We're getting closer.

What works for me
* indexing documents with and without full text
* fulltext search in documents with searchInDocumentPlugin

Using the fulltext search of the search plugin fails on showing the text snippets. Instead of the found snippet, the listview plugin marker `###SUBPREVIEW###` is filled with `"HPOS="1492"/><String WC="0.743529439" CONTENT="<em>Textil</em>-Industrie." HEIGHT="27" WIDTH="302" VPOS="388""`. 

fulltextkitodo

This seems to be caused in DocumentList.php: https://github.com/beatrycze-volk/kitodo-presentation/blob/use-solr-highlighting/Classes/Common/DocumentList.php#L278

I think, the adaption to the searchInDocumentTool you made for the oce highlighting plugin is necessary at this point, too. To make the general fulltext search work again with text snippets.

Could you please have a look at it?

beatrycze-volk:
> We're get!ting closer.

Great!

> Using the fulltext search of the search plugin fails on showing the text snippets. Instead of the found snippet, the listview plugin marker `###SUBPREVIEW###` is filled with `"HPOS="1492"/><String WC="0.743529439" CONTENT="<em>Textil</em>-Industrie." HEIGHT="27" WIDTH="302" VPOS="388""`.
> 
> fulltextkitodo
> 
> This seems to be caused in DocumentList.php: https://github.com/beatrycze-volk/kitodo-presentation/blob/use-solr-highlighting/Classes/Common/DocumentList.php#L278

Could you write down raw SOLR query which is created inside this method during search shown above?

albig:
> > This seems to be caused in DocumentList.php: https://github.com/beatrycze-volk/kitodo-presentation/blob/use-solr-highlighting/Classes/Common/DocumentList.php#L278
> 
> Could you write down raw SOLR query which is created inside this method during search shown above?

That's what Solr logs on my search for "Textil".

<pre>
2021-07-30 08:31:03.467 INFO  (qtp443721024-24) [   x:dlfCore0] o.a.s.c.S.Request [dlfCore0]  webapp=/solr path=/select params={q=fulltext:(textil)&json.nl=flat&omitHeader=true&fl=*,score&start=0&sort=score+desc&rows=0&wt=json} hits=294 status=0 QTime=0
2021-07-30 08:31:03.470 INFO  (qtp443721024-23) [   x:dlfCore0] o.a.s.c.S.Request [dlfCore0]  webapp=/solr path=/select params={q={!join+from%3Duid+to%3Duid}fulltext:(textil)&json.nl=flat&omitHeader=true&fl=uid,id&start=0&sort=score+desc&fq=toplevel:true&rows=25&wt=json} hits=6 status=0 QTime=1
2021-07-30 08:31:03.886 INFO  (qtp443721024-21) [   x:dlfCore0] o.a.s.c.S.Request [dlfCore0]  webapp=/solr path=/select params={json.nl=flat&hl=true&fl=uid,id,toplevel,thumbnail,page,type,title_tsi,volume_usu,author_tsi,place_tsi,year_usi&start=0&sort=score+desc&fq=uid:4&rows=10000&q=fulltext:(textil)+OR+toplevel:true&hl.useFastVectorHighlighter=true&omitHeader=true&hl.q=textil&hl.fl=fulltext&wt=json} hits=202 status=0 QTime=402
2021-07-30 08:31:04.305 INFO  (qtp443721024-22) [   x:dlfCore0] o.a.s.c.S.Request [dlfCore0]  webapp=/solr path=/select params={json.nl=flat&hl=true&fl=uid,id,toplevel,thumbnail,page,type,title_tsi,volume_usu,author_tsi,place_tsi,year_usi&start=0&sort=score+desc&fq=uid:3&rows=10000&q=fulltext:(textil)+OR+toplevel:true&hl.useFastVectorHighlighter=true&omitHeader=true&hl.q=textil&hl.fl=fulltext&wt=json} hits=25 status=0 QTime=50
2021-07-30 08:31:04.340 INFO  (qtp443721024-25) [   x:dlfCore0] o.a.s.c.S.Request [dlfCore0]  webapp=/solr path=/select params={json.nl=flat&hl=true&fl=uid,id,toplevel,thumbnail,page,type,title_tsi,volume_usu,author_tsi,place_tsi,year_usi&start=0&sort=score+desc&fq=uid:27&rows=10000&q=fulltext:(textil)+OR+toplevel:true&hl.useFastVectorHighlighter=true&omitHeader=true&hl.q=textil&hl.fl=fulltext&wt=json} hits=2 status=0 QTime=4
2021-07-30 08:31:04.438 INFO  (qtp443721024-24) [   x:dlfCore0] o.a.s.c.S.Request [dlfCore0]  webapp=/solr path=/select params={json.nl=flat&hl=true&fl=uid,id,toplevel,thumbnail,page,type,title_tsi,volume_usu,author_tsi,place_tsi,year_usi&start=0&sort=score+desc&fq=uid:5&rows=10000&q=fulltext:(textil)+OR+toplevel:true&hl.useFastVectorHighlighter=true&omitHeader=true&hl.q=textil&hl.fl=fulltext&wt=json} hits=57 status=0 QTime=92
2021-07-30 08:31:04.558 INFO  (qtp443721024-23) [   x:dlfCore0] o.a.s.c.S.Request [dlfCore0]  webapp=/solr path=/select params={json.nl=flat&hl=true&fl=uid,id,toplevel,thumbnail,page,type,title_tsi,volume_usu,author_tsi,place_tsi,year_usi&start=0&sort=score+desc&fq=uid:50&rows=10000&q=fulltext:(textil)+OR+toplevel:true&hl.useFastVectorHighlighter=true&omitHeader=true&hl.q=textil&hl.fl=fulltext&wt=json} hits=2 status=0 QTime=4
2021-07-30 08:31:04.584 INFO  (qtp443721024-21) [   x:dlfCore0] o.a.s.c.S.Request [dlfCore0]  webapp=/solr path=/select params={json.nl=flat&hl=true&fl=uid,id,toplevel,thumbnail,page,type,title_tsi,volume_usu,author_tsi,place_tsi,year_usi&start=0&sort=score+desc&fq=uid:60&rows=10000&q=fulltext:(textil)+OR+toplevel:true&hl.useFastVectorHighlighter=true&omitHeader=true&hl.q=textil&hl.fl=fulltext&wt=json} hits=12 status=0 QTime=20

</pre>

You can easily reproduce this with
* a page containing the search plugin
* a page containing the listview plugin

beatrycze-volk:
> That's what Solr logs on my search for "Textil".

Thanks!

> You can easily reproduce this with
> 
> * a page containing the search plugin
> * a page containing the listview plugin

Ok, I will need to get me page with this plugins on Monday.

 For now I have pushed the changes which should theoretically fix the problem, but it's not yet tested.

Classes/Common/DocumentList.php
```diff
@@ -316,6 +250,100 @@ protected function getRecord($element)
...
+        // Set additional query parameters.
+        // Set reasonable limit for safety reasons.
+        // We don't expect to get more than 10.000 hits per UID.
+        $query->setStart(0)->setRows(1000);
+        // Take over existing filter queries.
+        if (isset($this->metadata['options']['params']['filterquery'])) {
+            $query->addFilterQuery($this->metadata['options']['params']['filterquery']);
+        }
+        // Extend filter query to get all documents with the same UID.
+        foreach ($params['filterquery'] as $key => $value) {
```
albig: That's true what Scrutinizer tells us. $params is uninitialized and the foreach ends in an exception.

Classes/Common/DocumentList.php
```diff
@@ -316,6 +250,100 @@ protected function getRecord($element)
         return $record;
     }
 
+    private function getSolrResult($record) {
```
albig, line 264: Please add the PHPDocs.

albig (changes requested):
Unfortunately, your fast fix doesn't work yet. Please have a try after setting up your development environment.

Classes/Common/DocumentList.php
```diff
@@ -316,6 +252,122 @@ protected function getRecord($element)
...
+        // Add sorting.
+        $query->addSort('score', $this->metadata['options']['params']['sort']['score']);
+        // Set query.
+        $query->setQuery($this->metadata['options']['select'] . ' OR ' . $fields['toplevel'] . ':true');
+
+        // If it is a fulltext search, enable highlighting.
+        if ($this->metadata['fulltextSearch']) {
+            $query->setQuery(
+                $fields['fulltext'] . ':' . Solr::escapeQuery($this->metadata['searchString']));
+                // TODO:toplevel documents throw error in SOLR PLugin:
```
albig: I'm not sure, if this is the reason for the empty toplevel element in listview. Could you please have a look at this:

Listview

That's how it should look like:

Listview (1)

Below the toplevel element, the pages are shown. In the current state, the text snippets are shown again. That's perfect.

albig (changes requested):
Except for the mentioned issue in the listview, I stumbled about the word highlight.

I used the following GET-Parameter with your branch and with master. In your branch, no highlighting is active. Is this another bug or is this a missing configuration in my setup.

`tx_dlf[highlight_word]=Pieschen&tx_dlf[page]=22`

Digitale Sammlungen  Auf der Suche nach Zukunft  Das Beispiel Pieschen (1)

Digitale Sammlungen  Auf der Suche nach Zukunft  Das Beispiel Pieschen

Classes/Common/DocumentList.php
beatrycze-volk: Yes there is a problem with top element. It does't return any result if fulltext is empty as it throws error in SOLR Plugin.

So now basically plugin allows to index empty full text but return error when you search for document with such an empty full text field.

beatrycze-volk:
> Except for the mentioned issue in the listview, I stumbled about the word highlight.
> 
> I used the following GET-Parameter with your branch and with master. In your branch, no highlighting is active. Is this another bug or is this a missing configuration in my setup.
> 
> `tx_dlf[highlight_word]=Pieschen&tx_dlf[page]=22`
> 
In my branch highlight happens by coordinates given in result document in element 'highlight'.

albig:
> > `tx_dlf[highlight_word]=Pieschen&tx_dlf[page]=22`
> 
> In my branch highlight happens by coordinates given in result document in element 'highlight'.

I don't understand how to make this work:

1. Search -->
2. Listview (search results) -->
3. PageView with word highlighting

beatrycze-volk:
> I don't understand how to make this work:
> 
> 1. Search -->
> 2. Listview (search results) -->
> 3. PageView with word highlighting

After you click one of the hits in Listview and you will get redirected to certain document, the method `triggerSearchAfterHitLoad()` should be fired. This method will check if URL contains `tx_dlf[highlight_word]` parameter. If it exists and has value, search for this value will be executed. The returned result document will contain coordinates for highlighting,

The method which is drectly responsible for displaying highlights in image is called `addImageHighlight(data)`. This one will check if there is anything to highlight and if yes it will notify viewer. Then viewer will highlight words matching to coordinates stored in `element['highlight'].`

albig:
> > I don't understand how to make this work:
> > 
> > 1. Search -->
> > 2. Listview (search results) -->
> > 3. PageView with word highlighting
> 
> After you click one of the hits in Listview and you will get redirected to certain document, the method `triggerSearchAfterHitLoad()` should be fired. This method will check if URL contains `tx_dlf[highlight_word]` parameter. If it exists and has value, search for this value will be executed. The returned result document will contain coordinates for highlighting,
> 
> The method which is drectly responsible for displaying highlights in image is called `addImageHighlight(data)`. This one will check if there is anything to highlight and if yes it will notify viewer. Then viewer will highlight words matching to coordinates stored in `element['highlight'].`

Thank you for your explanation.

So, the text highlighting only works with the SearchInDocument plugin used together with the pageview plugin AND the TypoScript configuration for the "Toolbox" must be included.

Nevertheless, it's still not working. I'll add some remarks in the following review on the right lines.

Resources/Public/Javascript/PageView/SearchInDocument.js
```diff
@@ -43,10 +43,226 @@ function resetStart() {
...
+    }
+
+    return page;
+}
+
+function addImageHighlight(data) {
+    var page = getCurrentPage();
+
+    data['documents'].forEach(function (element, i) {
+        if(element['page'] === page) {
```
albig, line 233:
```suggestion
        if(element['page'] == page) {
```
In my Chromium console, the `page` is of type string. But the `element['page'] ` is of type integer. The comparison always fails.

Not sure, if it is the best way to change the comparison or fix the type of the `page` variable.

beatrycze-volk:
> So, the text highlighting only works with the SearchInDocument plugin used together with the pageview plugin AND the TypoScript configuration for the "Toolbox" must be included.

Currently yes. There is also possibility to implement it in the way that coordinates will be already passed in URL parameter.


> Nevertheless, it's still not working. I'll add some remarks in the following review on the right lines.

Thanks!

Resources/Public/Javascript/PageView/SearchInDocument.js
```diff
@@ -43,10 +43,226 @@ function resetStart() {
...
+}
+
+function getCurrentPage() {
+    var page = 1;
+    var queryParams = getCurrentQueryParams(getBaseUrl(" "));
+
+    for(var i = 0; i < queryParams.length; i++) {
+        var queryParam = queryParams[i].split('=');
+
+        if(queryParam[0] === $("input[id='tx-dlf-search-in-document-page']").attr('name')) {
```
albig: This comparison doesn't respect the possible encoding of the parameters. This is wrong on other places, too.

If you click next/last page, the link parameter get's encoded:
<code>
/workview?tx_dlf%5Bdouble%5D=0&tx_dlf%5Bhighlight_word%5D=pieschen&tx_dlf%5Bid%5D=1&tx_dlf%5Bpage%5D=3&cHash=04b8a699974161fd87b83c9b36d65483
</code>

But `'tx_dlf%5Bpage%5D' === 'tx_dlf[page]'` is not true so the highlighting fails.

albig (changes requested):
Please have look at my comments.

Resources/Public/Javascript/PageView/SearchInDocument.js
beatrycze-volk: I would convert string to number and compare them.

Resources/Public/Javascript/PageView/SearchInDocument.js
albig: Highlighting seems to work now. Thank you!

Classes/Common/DocumentList.php
```diff
@@ -316,6 +252,122 @@ protected function getRecord($element)
...
+                // java.lang.RuntimeException: Could not determine OCR format for sample ''
+                //. ' OR ' . $fields['toplevel'] . ':true');
```
albig:
```suggestion
```
Just remove these lines. If you do a fulltext search, it is done already done by `$this->metadata['options']['select']` in line 292. Here you only overwrite this query again.

Classes/Common/DocumentList.php
```diff
@@ -316,6 +252,122 @@ protected function getRecord($element)
...
+        // Extend filter query to get all documents with the same UID.
+        foreach ($filterQueries as $key => $value) {
+            if (isset($value['query'])) {
+                $filterQuery[$key] = $value['query'] . ' OR ' . $fields['toplevel'] . ':true';
+                $query->addFilterQuery($filterQuery);
+            }
+        }
+        // Add filter query to get all documents with the required uid.
+        $filterQuery['key'] = $fields['uid'] . ':' . Solr::escapeQuery($record['uid']);
+        $query->addFilterQuery($filterQuery);
```
albig:
```suggestion
        $query->createFilterQuery('uid')->setQuery($fields['uid'] . ':' . Solr::escapeQuery($record['uid']));
```
This doesn't work and will be ignored. As result, you always search fulltext:(something) OR toplevel:true and receive ALL toplevel documents. But you only want the toplevel element for the given uid.

`addFilterQuery()` expects an array like this:
`$filterQuery = [ 'key' => 'fq1', 'query' => 'uid:1' ];`

So I'd suggest to use `createFilterQuery()` like in the [documentation of Solarium](https://solarium.readthedocs.io/en/4.2.0/queries/select-query/building-a-select-query/adding-filterqueries/). Maybe you find another solution.

Classes/Common/DocumentList.php
```diff
@@ -316,6 +252,122 @@ protected function getRecord($element)
...
+        // Set reasonable limit for safety reasons.
+        // We don't expect to get more than 10.000 hits per UID.
+        $query->setStart(0)->setRows(1000);
+        // Take over existing filter queries.
+        $filterQueries = isset($this->metadata['options']['params']['filterquery']) ? $this->metadata['options']['params']['filterquery'] : [];
+        // Extend filter query to get all documents with the same UID.
+        foreach ($filterQueries as $key => $value) {
+            if (isset($value['query'])) {
+                $filterQuery[$key] = $value['query'] . ' OR ' . $fields['toplevel'] . ':true';
+                $query->addFilterQuery($filterQuery);
```
albig, line 287: This won't work either. But I found no way to test this. I think, there is another issue in `Search:makeFacetsMenuArray()`. So I can't get the available facets from Solr and show the facets menu.

Classes/Common/DocumentList.php
```diff
@@ -316,6 +252,122 @@ protected function getRecord($element)
...
+        $query = $this->solr->service->createSelect();
+        // Restrict the fields to the required ones
+        $query->setFields($fields['uid'] .',' . $fields['id'] .',' . $fields['toplevel'] .',' . $fields['thumbnail'] .',' . $fields['page']);
+        foreach ($this->solrConfig as $solr_name) {
+            $query->addField($solr_name);
+        }
+        // Set additional query parameters.
+        // Set reasonable limit for safety reasons.
+        // We don't expect to get more than 10.000 hits per UID.
+        $query->setStart(0)->setRows(1000);
```
albig:
```suggestion
        $query->setStart(0)->setRows(10000);
```
Fix the limit.

albig (changes requested):
With the patched ocr-highlighting plugin, I see the toplevel element. But there are some issues with the facets and the filter queries. Please see my comments.

Classes/Common/DocumentList.php
beatrycze-volk: I have taken look to this method. Do you think that it needs to be adjusted now or it can be done later?

albig: The addFilterQuery() above has to be fixed. But before doing this, I'm looking for working facets. Currently, the Solr has no facets available and I still don't know, why.

A query like 

`/solr/dlfCore0/select?facet=on&q=*%3A*&rows=0`

Should return something like

<pre>
{
  "responseHeader":{
    "status":0,
    "QTime":2,
    "params":{
      "q":"*:*",
      "rows":"0",
      "facet":"on"}},
  "response":{"numFound":4517,"start":0,"numFoundExact":true,"docs":[]
  },
  "facet_counts":{
    "facet_queries":{},
    "facet_fields":{},
    "facet_ranges":{},
    "facet_intervals":{},
    "facet_heatmaps":{}}}
</pre>

In my case - with this PR working - I get

<pre>
{
  "responseHeader":{
    "status":0,
    "QTime":0,
    "params":{
      "q":"*:*",
      "rows":"0",
      "facet":"on"}},
  "response":{"numFound":4376,"start":0,"numFoundExact":true,"docs":[]
  }}
</pre>

That's why no facetting is shown.

Classes/Common/Indexer.php
```diff
@@ -443,10 +448,10 @@ protected static function processLogical(Document &$doc, array $logicalUnit)
      */
     protected static function processPhysical(Document &$doc, $page, array $physicalUnit)
     {
-        if (
-            $doc->hasFulltext
-            && $fulltext = $doc->getRawText($physicalUnit['id'])
-        ) {
+        $logger = GeneralUtility::makeInstance('TYPO3\CMS\Core\Log\LogManager')->getLogger(__CLASS__);
+
+        $logger->error($doc->hasFulltext && $fullText = $doc->getFullText($physicalUnit['id']));
```
albig: This line is not needed, isn't it?
```suggestion
```

Configuration/ApacheSolr/configsets/dlf/conf/solrconfig.xml
```diff
@@ -700,6 +708,12 @@
          <str name="wt">xml</str> 
         -->
     </lst>
+
+    <arr name="components">
+          <str>query</str>
+          <str>ocrHighlight</str>
```
albig, line 715:
```suggestion
          <str>facet</str>
          <str>ocrHighlight</str>
```
We use the facetComponent for search facetting.

Classes/Common/DocumentList.php
albig: The faceting is solved (see other commit suggest).

With selecting a facet, we get the (expected) exception:

"A filterquery must have a key value" in Line 283.

The used GET-parameter are:

`tx_dlf[fq][0]=collection_faceting:("Projekt: Illustrierte Magazine der Klassischen Moderne")&tx_dlf[fq][1]=type_faceting:("volume")&tx_dlf[query]=*`

So please rewrite Line 283.

Classes/Plugin/Eid/SearchInDocument.php
```diff
@@ -58,21 +59,37 @@ public function main(ServerRequestInterface $request)
...
+            // is required if you want to have OCR highlighting
+            $solrRequest->addParam('hl.ocr.fl', $fields['fulltext']);
+            // return the coordinates of highlighted search as absolute coordinates
+            $solrRequest->addParam('hl.ocr.absoluteHighlights', 'on');
+            // max amount of snippets for a single page
+            $solrRequest->addParam('hl.snippets', 20);
+
+            $response = $solr->service->executeRequest($solrRequest);
+            $result = $solr->service->createResult($query, $response);
+            $output['numFound'] = $result->getNumFound();
```
albig, line 81:
```suggestion
            /** @scrutinizer ignore-call */
            $output['numFound'] = $result->getNumFound();
```
This is a false-positiv and should be ignored.

albig (changes requested):
Maybe the last change request for now. Please have a look at the comments.

beatrycze-volk:
> With selecting a facet, we get the (expected) exception:
> 
> "A filterquery must have a key value" in Line 283.

Is `$key` not a key value?

Classes/Common/DocumentList.php
```diff
@@ -316,6 +252,116 @@ protected function getRecord($element)
...
+        // Set additional query parameters.
+        // Set reasonable limit for safety reasons.
+        // We don't expect to get more than 10.000 hits per UID.
+        $query->setStart(0)->setRows(10000);
+        // Take over existing filter queries.
+        $filterQueries = isset($this->metadata['options']['params']['filterquery']) ? $this->metadata['options']['params']['filterquery'] : [];
+        // Extend filter query to get all documents with the same UID.
+        foreach ($filterQueries as $key => $value) {
+            if (isset($value['query'])) {
+                $filterQuery[$key] = $value['query'] . ' OR ' . $fields['toplevel'] . ':true';
```
albig, line 282:
```suggestion
                $filterQuery[$key] = $value['query'] . ' OR ' . $fields['toplevel'] . ':true';
                $filterQuery = [
                    'key' => $key,
                    'query' => $value['query'] . ' OR ' . $fields['toplevel'] . ':true'
                ];
```

Solarium expects an array like this at this point. Important is the key "key". The key "query" seems not to matter.

The other approach would be in the above foreach loop:

`$query->createFilterQuery('kitodo' . $key)->setQuery($value['query'] . ' OR ' . $fields['toplevel'] . ':true');`

This calls addFilterQuery internally.

beatrycze-volk: Oh, now I understand. Thanks!

albig:
I think we finally reached the state, all features seems to work and I can have a look at the performance of the new ocr highlighting plugin and about the "migration" path from last release to next release containing this feature. So maybe, there is coming another change request - or the final merge in new future.

albig:
Just to give an update.

Indexing with the same set of documents seems to work with this PR. But there are missing some pages. Obviously the ocrhighlight plugin doesn't like pages with the `<Illustration>` tag.

Example:
https://digital.slub-dresden.de/data/kitodo/lebe_364572701-19320800/lebe_364572701-19320800_ocr/00000068.xml
<pre>
at java.base/java.lang.Thread.run(Unknown Source)
Caused by: java.lang.RuntimeException: Failed to parse the OCR markup, make sure your files are well-formed and your regions start/end on complete tags! (Source was: [unknown])
	at de.digitalcollections.solrocr.formats.OcrParser.<init>(OcrParser.java:109)
	at de.digitalcollections.solrocr.formats.alto.AltoParser.<init>(AltoParser.java:27)
	at de.digitalcollections.solrocr.formats.alto.AltoFormat.getParser(AltoFormat.java:38)
	at de.digitalcollections.solrocr.model.OcrFormat.filter(OcrFormat.java:90)
	at de.digitalcollections.solrocr.lucene.filters.OcrCharFilterFactory.create(OcrCharFilterFactory.java:51)
</pre>

The other point is, that the index size is 3 times (!) bigger than before.

albig:
> Just to give an update.

I'm again on this PR. Just a new update.

> Indexing with the same set of documents seems to work with this PR. But there are missing some pages. Obviously the ocrhighlight plugin doesn't like pages with the `<Illustration>` tag.

It has been to very long `<Polygon>` element. See https://github.com/dbmdz/solr-ocrhighlighting/issues/210. This has been fixed.

> The other point is, that the index size is 3 times (!) bigger than before.

Storing the whole ALTO is not that nice. This PR ignores additionally the ALTO namespace. Not sure, if we should go this way.

If we store only the suggested miniOcr-format, the index has "only" double size than before.

beatrycze-volk:
> > Just to give an update.
> 
> I'm again on this PR. Just a new update.
> 
> > Indexing with the same set of documents seems to work with this PR. But there are missing some pages. Obviously the ocrhighlight plugin doesn't like pages with the `<Illustration>` tag.
> 
> It has been to very long `<Polygon>` element. See [dbmdz/solr-ocrhighlighting#210](https://github.com/dbmdz/solr-ocrhighlighting/issues/210). This has been fixed.

I was actually also on this PR. Right after push my 'fix' commit I have seen this comment. Probably I will need to delete it.

> > The other point is, that the index size is 3 times (!) bigger than before.
> 
> Storing the whole ALTO is not that nice. This PR ignores additionally the ALTO namespace. Not sure, if we should go this way.
> 
> If we store only the suggested miniOcr-format, the index has "only" double size than before.

Ok. Would it be our approach now to convert files to miniOCR? Should it be done in our code or by some external tool?

albig:
> I was actually also on this PR. Right after push my 'fix' commit I have seen this comment. Probably I will need to delete it.

Yes, I think so. This should'nt be fixed on our side.

> Ok. Would it be our approach now to convert files to miniOCR? Should it be done in our code or by some external tool?

I've done this in our code because this miniOCR is no standard format.

I'd suggest the following procedure to continue:

1. I merge #587 and #639 into a new feature branch `add-ocrhighlighting-plugin`
2. I'll add my prototype code for miniOCR to this branch
3. We can discuss the best solution in a new PR `add-ocrhighlighting-plugin` -> `master`

beatrycze-volk:
> 1. I merge [Use Solr OCR Highlighting Plugin in Search in Document Plugin #587](https://github.com/kitodo/kitodo-presentation/pull/587) and [Remove obsolete hl parameter use in image highlighting #639](https://github.com/kitodo/kitodo-presentation/pull/639) into a new feature branch `add-ocrhighlighting-plugin`
> 2. I'll add my prototype code for miniOCR to this branch
> 3. We can discuss the best solution in a new PR `add-ocrhighlighting-plugin` -> `master`

Ok!

PR #589, opened 2021-02-16
Remove unused scrollTo method from full text control

beatrycze-volk:
Is it still necessary for some special cases? Scrolling works for me without it.


sebastian-meyer (dismissed):

sebastian-meyer:
You are right, the method isn't necessary anymore. It was a fallback for using old versions of jQuery.

But although your pull request is fine and appreciated, I'd prefer to dismiss it, because it may cause conflicts with #584 (where this file is heavily changed and moved). I've removed the function from my working branch for #584.

sebastian-meyer:
Changes included in #584.

PR #592, opened 2021-02-17
Allow multiple file groups of any type

sebastian-meyer:
This pull request adds the possibility to specify more than one file group for fulltext, thumbs, audio and downloads. The comma-separated values are checked from left to right and the first fileGrp available in the METS file is used.

This is used in the Zeitungsportal in order to provide a locally hosted fallback for files (`fileGrpFulltext = DDB_FULLTEXT,FULLTEXT`).


Classes/Plugin/AudioPlayer.php
```diff
@@ -123,18 +129,24 @@ public function main($content, $conf)
...
-            // Get audio data.
-            $this->audio['url'] = $this->doc->getFileLocation($this->doc->physicalStructureInfo[$this->doc->physicalStructure[$this->piVars['page']]]['files'][$this->conf['fileGrpAudio']]);
-            $this->audio['label'] = $this->doc->physicalStructureInfo[$this->doc->physicalStructure[$this->piVars['page']]]['label'];
-            $this->audio['mimetype'] = $this->doc->getFileMimeType($this->doc->physicalStructureInfo[$this->doc->physicalStructure[$this->piVars['page']]]['files'][$this->conf['fileGrpAudio']]);
+        $fileGrpAudios = GeneralUtility::trimExplode(',', $this->conf['fileGrpAudio']);
+        while ($fileGrpAudio = array_shift($fileGrpAudios)) {
+            if (!empty($this->doc->physicalStructureInfo[$this->doc->physicalStructure[$this->piVars['page']]]['files'][$fileGrpAudio])) {
+                // Get audio data.
+                $this->audio['url'] = $this->doc->getFileLocation($this->doc->physicalStructureInfo[$this->doc->physicalStructure[$this->piVars['page']]]['files'][$fileGrpAudio);
```
albig: I think Scrutinizer is right here. The brackets are not correct.

albig (changes requested):
Good idea. Haven't tried it yet. Just stumbled about the Scrutinizer warning.

Classes/Plugin/AudioPlayer.php
sebastian-meyer: Fixed with 40c738f.

beatrycze-volk (commented):
I wanted to test it but get error:

```
undefined:1 GET https://sdvtypo3ddbzeitungsportal.slub-dresden.de/undefined 404

  | Image (async) |   |  
-- | -- | -- | --
  | dlfUtils.fetchStaticImageData | @ | Utility.js:288
  | (anonymous) | @ | Utility.js:248
  | dlfUtils.fetchImageData | @ | Utility.js:227
  | dlfViewer.initLayer | @ | PageView.js:482
  | dlfViewer.init | @ | PageView.js:376
  | dlfViewer | @ | PageView.js:14
```


I have checked it locally and on test system. In both cases the same error.

sebastian-meyer:
Since `fileGrps` was renamed to `fileGrpImages`, the configuration for the image file groups has to be reloaded. Just open the extension configuration, switch to the "Files" tab, make sure that `file.fileGrpImages` is set correctly and save the configuration.

sebastian-meyer:
I'll try to add a fallback for existing installations to ease updating...

beatrycze-volk (approved):

sebastian-meyer:
@albig This is ready for merging as soon as you can confirm that the changes you requested (regarding the Scrutinizer warnings) were made.

albig (approved):
Works as expected.

PR #588, opened 2021-02-16
Add configuration for fields which need to be highlighted in the image

beatrycze-volk:
This PR allows to use more than one field for highlighting words in the image. It gives also more flexibility in the name of parameters. 
Example: use 'query' param to highlight word after clicking the found result in the result list, right after chosen document is loaded it highlights word for which user was searching and then you have param 'highlight_word' which highlights word searched already inside the document itself. The first defined param on the configuration has highest priority, if not found it checks next ones. Parameters are separated by commas.

- add TypoScript configuration
- add documentation for configurable field 'searchHlElements'
- pass field value to full text control by dictionary
- use it in page view

PR #599, opened 2021-02-22
Highlight exact word

beatrycze-volk:
DDB has reported that highlight of the found word includes words which should not be there. 

Example:
`the` will highlight also '**the**ir', '**the**y' or '**The**odore'.

Depends on #588

### DIFF
--- a/Classes/Plugin/Tools/FulltextTool.php
+++ b/Classes/Plugin/Tools/FulltextTool.php
@@ -77,7 +77,13 @@ class FulltextTool extends \Kitodo\Dlf\Common\AbstractPlugin
         $this->getTemplate();
         $fullTextFile = $this->doc->physicalStructureInfo[$this->doc->physicalStructure[$this->piVars['page']]]['files'][$this->conf['fileGrpFulltext']];
         if (!empty($fullTextFile)) {
-            $markerArray['###FULLTEXT_SELECT###'] = '<a class="select switchoff" id="tx-dlf-tools-fulltext" title="" data-dic="fulltext:' . htmlspecialchars($this->pi_getLL('fulltext', '')) . ';fulltext-on:' . htmlspecialchars($this->pi_getLL('fulltext-on', '')) . ';fulltext-off:' . htmlspecialchars($this->pi_getLL('fulltext-off', '')) . ';activate-full-text-initially:' . MathUtility::forceIntegerInRange($this->conf['activateFullTextInitially'], 0, 1, 0) . ';full-text-scroll-element:' . $this->conf['fullTextScrollElement'] . '">&nbsp;</a>';
+            $markerArray['###FULLTEXT_SELECT###'] = '<a class="select switchoff" id="tx-dlf-tools-fulltext" title="" data-dic="'
+            . 'fulltext:' . htmlspecialchars($this->pi_getLL('fulltext', ''))
+            . ';fulltext-on:' . htmlspecialchars($this->pi_getLL('fulltext-on', ''))
+            . ';fulltext-off:' . htmlspecialchars($this->pi_getLL('fulltext-off', ''))
+            . ';activate-full-text-initially:' . MathUtility::forceIntegerInRange($this->conf['activateFullTextInitially'], 0, 1, 0)
+            . ';full-text-scroll-element:' . $this->conf['fullTextScrollElement']
+            . ';search-hl-elements:' . $this->conf['searchHlElements'] . '">&nbsp;</a>';
         } else {
             $markerArray['###FULLTEXT_SELECT###'] = '<span class="no-fulltext">' . htmlspecialchars($this->pi_getLL('fulltext-not-available', '')) . '</span>';
         }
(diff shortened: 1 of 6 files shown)


albig (changes requested):
Actually the code seems works like before. That's very good.

I used
* pageview 
* fulltext active
* searchInDocument

--> The word is highlighted.

If I change the TypoScript

```
plugin.tx_dlf_fulltexttool.searchHlElements = tx_dlf[highlight_word],tx_dlf[query],tx_dlf[another]
```

It still works for the default `tx_dlf[highlight_word]` parameter. If I add `tx_dlf[other]`, this doesn't  highlight. Only if `tx_dlf[highlight_word]` is missing. That's what you wrote in #588.

But is there any usage of another highlighting search element in the current Kitodo.Presentation? I do not get the point. Sorry ;-)

And if so, shouldn't we write some sentence to explain this setting / feature?

beatrycze-volk:
_It still works for the default tx_dlf[highlight_word] parameter. If I add tx_dlf[other], this doesn't highlight. Only if tx_dlf[highlight_word] is missing. That's what you wrote in #588._

 The first defined param on the configuration has highest priority, if not found it checks next ones.
 So in this case the param from search in the document has highest priority and search word from hit list has the second priority.

_But is there any usage of another highlighting search element in the current Kitodo.Presentation?_

No idea ;) IN DDB they use it in the way that if the newspaper is open from list of found results it highlights the search term which uses param 'query' and search in document uses different param to distinguish those two searches.

albig:
> _But is there any usage of another highlighting search element in the current Kitodo.Presentation?_
> 
> No idea ;) IN DDB they use it in the way that if the newspaper is open from list of found results it highlights the search term which uses param 'query' and search in document uses different param to distinguish those two searches.

With the configuration
```
plugin.tx_dlf_fulltexttool.searchHlElements = query,tx_dlf[highlight_word]
```

The flow would be

1. launch the pageview (from some listing outside) with e.g. `&query=Dresden` --> `Dresden` will be highlighted.
2. search in document inside the pageview --> `&tx_dlf[highlight_word]=Leipzig` --> `Leipzig` will be highlighted. But `Dresden` won't anymore.

You cannot highlight both and there is no use case for this, right?

If I understand this right, this feature doesn't hurt Kitodo.Presentation but also does not bring any advantage for itself. So we could merge it, if you just confirm my assumptions and add the explanation in your comment to the documentation.

beatrycze-volk:
_You cannot highlight both and there is no use case for this, right?_

Exactly so.

_If I understand this right, this feature doesn't hurt Kitodo.Presentation but also does not bring any advantage for itself. So we could merge it, if you just confirm my assumptions and add the explanation in your comment to the documentation._

True, it almost doesn't bring anything new and also doesn't hurt. Default behavior is preserved.

sebastian-meyer:
This pull request's title is _Highlight exact word_ and its description suggests that it's about highlighting whole words only instead of highlighting matching parts of a word as well. But all you are talking about is using different URL parameters for highlighting. I am confused...

What is this pull request really about? Does this pull request contain both changes? If yes, does it supersede #588 then?

Regarding the original intentions of highlighting only whole words: I don't think this is a good idea. Especially in the German language there are a lot of composed words and grammatical forms which wouldn't be highlighted anymore. Someone searching for e. g. "Dresden" would certainly want to find "Dresdens" or "Dresden-Klotzsche" as well. This should at least be configurable while preserving the current behaviour as default.

beatrycze-volk:
_This pull request's title is Highlight exact word and its description suggests that it's about highlighting whole words only instead of highlighting matching parts of a word as well. But all you are talking about is using different URL parameters for highlighting. I am confused..._

_What is this pull request really about? Does this pull request contain both changes? If yes, does it supersede #588 then?_

It contains also changes from previous PR, this one is only about highlighting exact words.

_Regarding the original intentions of highlighting only whole words: I don't think this is a good idea. Especially in the German language there are a lot of composed words and grammatical forms which wouldn't be highlighted anymore. Someone searching for e. g. "Dresden" would certainly want to find "Dresdens" or "Dresden-Klotzsche" as well. This should at least be configurable while preserving the current behaviour as default._

It was exactly what was expectation from Zeitungsportal. They expect that only exact word will be higlighted, not part of the words. What should be doing in that case? Should this change be only for them and not introduced here or maybe should be used parameter which lets decide about type of higlight (e.g. exactHighlight true or false)?

sebastian-meyer:
> It contains also changes from previous PR, this one is only about highlighting exact words.

OK, so #588 should be merged before this one. 👍🏼 

> It was exactly what was expectation from Zeitungsportal. They expect that only exact word will be higlighted, not part of the words. What should be doing in that case? Should this change be only for them and not introduced here or maybe should be used parameter which lets decide about type of higlight (e.g. exactHighlight true or false)?

I'd prefer the former: Since this change makes highlighting behave different then searching, I think it's a special case for the Zeitungsportal and shouldn't be introduced here. IMHO users would be confused if they get a hit searching for e. g. "Dresden", but then don't see anything highlighted because of the strict matching. On the other hand: maintaining a separate fork just for the DDB is a pain in the a**, too. @albig: What do you think?

beatrycze-volk:
@sebastian-meyer we have special extenstion for them, all changes which are not matching here, are introduced there.

sebastian-meyer:
> @sebastian-meyer we have special extenstion for them, all changes which are not matching here, are introduced there.

Yes, but is there a way to introduce this change there without forking the whold `PageView` plugin?

beatrycze-volk:
@sebastian-meyer we are replacing there JS files and pass them as an one file.

sebastian-meyer:
OK, so when it's fine with you, I'd suggest to keep this out of Kitodo.Presentation.

PR #602, opened 2021-03-02
Fix interaction initialization for OL map

beatrycze-volk:
It looks that missing brackets were causing problems and sometimes method zoomTo on this.map was throwing error.


albig (changes requested):
I sea no reason, why these brackets are missing. But we should write it in `Resources/Public/Javascript/PageView/ImageManipulationControl.js` the same way. Could you please add this?

albig (approved):

PR #603, opened 2021-03-22
Show all subparts in collection view

### LINKED ISSUE #559: Collection plugin does not list all volumes with same publication date
## Description

Two volumes overwrite each other in the listview of a single collection if the `volume_sorting` field is equal.

## Reproduction

Index two volumes with same `volume_sorting` value.

## Expected Behavior

Volumes of multivolume works get the `volume_sorting` field set in the database to the publication date.

### Document 1
```
<mods:originInfo eventType="publication">
  <mods:dateIssued keyDate="yes">[1887-1900]</mods:dateIssued>
  <mods:dateOther encoding="w3cdtf" type="order">1900</mods:dateOther>
</mods:originInfo>
```
### Document 2
```
<mods:originInfo eventType="publication">
  <mods:dateIssued keyDate="yes">[circa 1900]</mods:dateIssued>
  <mods:dateOther encoding="w3cdtf" type="order">1900</mods:dateOther>
</mods:originInfo>
```
Both documents have `volume_sorting=1900` after indexing.

The code in `Collection::showSingleCollection()` [Line 402](https://github.com/kitodo/kitodo-presentation/blob/master/Classes/Plugin/Collection.php#L402) assumes the `volume_sorting` field to be unique which is not the case.

## Question

Does this make sense? Why is the ORDER-Label in the structMap type="LOGICAL" ignored for this purpose? Shouldn't this be changed in the document class?

## Screenshots and Examples

Example files:

https://digital.slub-dresden.de/data/kitodo/Dien_1683659783_0008/Dien_1683659783_0008_mets.xml
https://digital.slub-dresden.de/data/kitodo/Dien_1683658167_0006/Dien_1683658167_0006_mets.xml

## Environment

 - Kitodo.Presentation 3.1.2

albig:
This will fix #559.

### Case 1 - year_sorting is not unique

Examples:

* https://digital.slub-dresden.de/data/kitodo/Dien_1683658167_0006/Dien_1683658167_0006_mets.xml
* https://digital.slub-dresden.de/data/kitodo/Dien_1683659783_0008/Dien_1683659783_0008_mets.xml

Both have `year_sorting == 1900`

### Case 2 - Newspaper/Ephemera structure year

Examples:

* https://digital.slub-dresden.de/data/kitodo/Brsfded_39946221X-19450106/Brsfded_39946221X-19450106_year.xml
* https://digital.slub-dresden.de/data/kitodo/Brsfded_39946221X-19441202/Brsfded_39946221X-19441202_year.xml

Both have no `title` or `title_sorting` set.

### Case 3 - other reasons

For possible other reasons why the volume_sorting is missing, the collection plugin adds the uid to the array key of the subparts to list. This makes sure to list all subparts even if the order may be wrong.

### DIFF
--- a/Classes/Plugin/Collection.php
+++ b/Classes/Plugin/Collection.php
@@ -404,7 +404,7 @@ class Collection extends \Kitodo\Dlf\Common\AbstractPlugin
                 ];
             } else {
                 // volume_sorting should be always set - but it's not a required field. We append the uid to the array key to make it always unique.
-                $subparts[$resArray['partof']][$resArray['volume_sorting'] . $resArray['uid']] = [
+                $subparts[$resArray['partof']][$resArray['volume_sorting'] . str_pad($resArray['uid'], 9, '0', STR_PAD_LEFT)] = [
                     'u' => $resArray['uid'],
                     'h' => '',
                     's' => $sorting,
--- a/Classes/Plugin/Collection.php
+++ b/Classes/Plugin/Collection.php
@@ -403,7 +403,8 @@ class Collection extends \Kitodo\Dlf\Common\AbstractPlugin
                     'p' => []
                 ];
             } else {
-                $subparts[$resArray['partof']][$resArray['volume_sorting']] = [
+                // volume_sorting should be always set - but it's not a required field. We append the uid to the array key to make it always unique.
+                $subparts[$resArray['partof']][$resArray['volume_sorting'] . $resArray['uid']] = [
                     'u' => $resArray['uid'],
                     'h' => '',
                     's' => $sorting,
--- a/Classes/Common/Document.php
+++ b/Classes/Common/Document.php
@@ -1169,7 +1169,10 @@ abstract class Document
                 $metadata['volume'] = $metadata['year'];
             }
             if (empty($metadata['volume_sorting'][0])) {
-                if (!empty($metadata['year_sorting'][0])) {
+                // If mets ORDER label is given it is prefered over year_sorting and year.
+                if (!empty($metadata['mets_order'][0])) {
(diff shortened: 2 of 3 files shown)


Classes/Plugin/Collection.php
```diff
@@ -403,7 +403,8 @@ protected function showSingleCollection($id)
                     'p' => []
                 ];
             } else {
-                $subparts[$resArray['partof']][$resArray['volume_sorting']] = [
+                // volume_sorting should be always set - but it's not a required field. We append the uid to the array key to make it always unique.
+                $subparts[$resArray['partof']][$resArray['volume_sorting'] . $resArray['uid']] = [
```
sebastian-meyer: Won't this lead to wrong sorting if UIDs differ in number of digits? E. g. one volume has `volume_sorting = 1900` and `uid = 10`. Another volume has `volume_sorting = 1901` and `uid = 9`. The intended sorting order obviously would be "1900 before 1901", but because of the appending they actually would be sorted "**1901**9 before **1900**10".

sebastian-meyer (changes requested):

Classes/Plugin/Collection.php
sebastian-meyer:
```suggestion
                $subparts[$resArray['partof']][$resArray['volume_sorting'] . str_pad($resArray['uid'], 9, '0', STR_PAD_LEFT)] = [
```
This should do the job. ;o)

albig: You're right. This can break the sorting depending on the amount of decimals. Your suggested patch is working. Thank you!

sebastian-meyer (approved):

PR #605, opened 2021-03-29
Move include of SearchInDocument.js to PageView

beatrycze-volk:
That's necessary to really get all pageview related JavaScript files out of the footer into the ###JAVASCRITP### marker of the pageview plugin.


albig (approved):
Works as expected. The searchInDocument plugin is intended to be used with the pageview plugin.

PR #607, opened 2021-03-30
Move JS includes from source code to TS configuration

### LINKED ISSUE #606: Move JavaScript from plugin code to TypoScript
## Description

Currently, three plugins add JavaScript to the website:
* Pageview
* AudioPlayer
* Search

Usually this is done by adding the files to the footer with Pagerenderer::addJsFooterFile().

In case of the Pageview and AudioPlayer plugins, it is possible to add all JavaScript to a marker inside the plugin section of the website body.

### Problem

This behavior makes it impossible to overwrite or exclude some JavaScript files or to influence the position (head or footer) of the included JavaScript files.

### Proposed Solution

The JavaScript files should be added by [TypoScript](https://docs.typo3.org/m/typo3/reference-typoscript/master/en-us/Setup/Page/Index.html#includejs-array). A basis TypoScript-Template is already part of Kitodo.Presentation and provides jQuery. This could be extended or a new one can be added.

The integrator now has the possibility to take this configuration or to discard it and uses the flexibility of TypoScript.

In case of the Pageview plugin, the dynamic JavaScript inline code must stay in the plugin. For this, the feature of the template marker may be preserved.

## Target Version

master

beatrycze-volk:
Solution for #606

### DIFF
--- a/Classes/Plugin/AudioPlayer.php
+++ b/Classes/Plugin/AudioPlayer.php
@@ -43,26 +43,15 @@ class AudioPlayer extends \Kitodo\Dlf\Common\AbstractPlugin
      *
      * @access protected
      *
-     * @return string The output string for the ###JAVASCRIPT### template marker
+     * @return void
      */
     protected function addPlayerJS()
     {
-        $markerArray = '';
-        // CSS files.
-        $cssFiles = [
-            'Resources/Public/Javascript/jPlayer/blue.monday/css/jplayer.blue.monday.min.css'
-        ];
         // Inline CSS.
         $inlineCSS = '#tx-dlf-audio { width: 100px; height: 100px; }';
-        //Javascript files.
-        $jsFiles = [
-            // jPlayer
-            'Resources/Public/Javascript/jPlayer/jquery.jplayer.min.js',
-            // AudioPlayer
-            'Resources/Public/Javascript/AudioPlayer/AudioPlayer.js'
-        ];
+
         // AudioPlayer configuration.
-        $audioplayerConfiguration = '
+        $audioPlayerConfiguration = '
             $(document).ready(function() {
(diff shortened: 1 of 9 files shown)


albig:
Looks quite complete complete. I'll have a try later.

sebastian-meyer (approved):
This should work, but it has one major drawback: All the javascript files are now included on every page, even if the corresponding plugins aren't even used. This not only slows page rendering down, but also causes unnecessary javascript processing.

albig:
> This should work, but it has one major drawback: All the javascript files are now included on every page, even if the corresponding plugins aren't even used. This not only slows page rendering down, but also causes unnecessary javascript processing.

That's of course true but I think you cannot measure any big differences in loading and rendering speed today. If optimization is really a task, it could be done by excluding certain files by TypoScript (e.g. OpenLayers is not necessary on pure listview). So I rate the pros higher than the cons.

But we loose with this pull request the feature to render the HTML markup of only one plugin. Especially the pageview has it's inline JavaScript with a necessary array (link to image, fulltext, etc.). This will be removed with `config.disableAllHeaderCode = 1`. And you cannot add it by TypoScript from outside. I found no way to get this working. All the other JavaScript could be added.

So maybe, we should keep this feature for the inline JavaScript? Now it could be written on the marker `###JAVASCRIPT###`. Later it could be rendered in the plugin fluid template.

So we need this functionality?

beatrycze-volk:
This PR still keeps functionality for adding inline JS. It is included in the footer.

albig:
> This PR still keeps functionality for adding inline JS. It is included in the footer.

Yes. In normal operation this is true. 

But in described "special case", the footer will be cleaned by `config.disableAllHeaderCode = 1` (the option name may be missleading).

PR #609, opened 2021-04-13
Add model and repository classes for upgrade to Extbase/Fluid

beatrycze-volk:
It is first step for upgrade to Extbase/Fluid:

- [create-the-domain-model](https://docs.typo3.org/m/typo3/book-extbasefluid/master/en-us/4-FirstExtension/3-create-the-domain-model.html)
- [make-products-persistent](https://docs.typo3.org/m/typo3/book-extbasefluid/master/en-us/4-FirstExtension/4-make-products-persistent.html)

It is necessary to run database migration (Admin Tools -> Extensions -> Update)


Classes/Command/IndexCommand.php
```diff
@@ -171,20 +170,8 @@ protected function execute(InputInterface $input, OutputInterface $output)
...
-            ->select('uid', 'index_name')
-            ->from('tx_dlf_solrcores')
-            ->where(
-                $queryBuilder->expr()->eq(
-                    'pid',
-                    $queryBuilder->createNamedParameter((int) $pageId, Connection::PARAM_INT)
-                )
-            )
-            ->execute();
+        $result = SolrCoreRepository::findByPid($pageId);
```
albig, line 174: That's not the way you access the repositories in Extbase usually. You know this guide? https://docs.typo3.org/m/typo3/book-extbasefluid/9.5/en-us/6-Persistence/2a-creating-the-repositories.html

Classes/Domain/Table.php
```diff
@@ -0,0 +1,151 @@
+<?php
```
albig, line 1: Never seen something like this.

ext_tables.sql
```diff
@@ -1,7 +1,7 @@
 --
--- Table structure for table 'tx_dlf_documents'
+-- Table structure for table 'tx_dlf_domain_model_document'
 --
-CREATE TABLE tx_dlf_documents (
+CREATE TABLE tx_dlf_domain_model_document (
```
albig, line 4: That's not really necessary. You can keep the "old" table names.

albig (changes requested):
I assume this is work in progress. Or did you test this with real data?

Did you have a look at the domain model and repositories in https://github.com/slub/slub_digitalcollections/tree/master/Classes? This is of course a special case and implements only a small part of Kitodo.Presentation. It's not perfect, but especially concerning the repositories, this is my expected implementation.

I would start with small steps to be able to test the result.

Classes/Command/IndexCommand.php
beatrycze-volk: That's a temporar solution. I have read about it :) It is a first step in which I wanted to separate all places where actual access to database is performed and move it to repositories.

beatrycze-volk:
I have tested this changes on my local dev system. Migration was performed correctly, newspaper was displayed, view of all issues for this newspaper was also displayed correctly.

ext_tables.sql
beatrycze-volk: That's good to know, becasue after reading few tutorials I have understood that the tables names should respond to the package in which the entity classes are placed. I was trying to find a way to inform the entity that table has actually different name.

How does Document entity knows that it accesses table tx_dlf_documents? 

In general I would be for unifing the table names to be only singular or only plural.

Classes/Domain/Table.php
beatrycze-volk: Actually I wanted to use enum not class but it turns out that PHP 7.x doesn't have yet access to it (https://stitcher.io/blog/php-enums).

The main idea here is to write string names of tables only once and use later only this variables instead of writing strings over and over again. I find it to be a bit less error prone.

beatrycze-volk:
> Did you have a look at the domain model and repositories in https://github.com/slub/slub_digitalcollections/tree/master/Classes? This is of course a special case and implements only a small part of Kitodo.Presentation. It's not perfect, but especially concerning the repositories, this is my expected implementation.

Rewriting queries to the way way as it is there is going to be my next step. I wanted first separate queries without actuall chaning them.

One question: Why are used there AbstractValueObject instead of AbstractEntity?

After that stuff is working, I'm planning to change plugins to controllers.

ext_tables.sql
albig: This could be done by TypoScript mapping

e.g. like this in ext_typoscript_setup.txt:

```
config.tx_extbase.persistence.classes {
    Kitodo\Dlf\Domain\Model\Document {
        mapping {
            tableName = tx_dlf_documents
        }
    }
}
```

Ok, this has been dropped and replaced by a PHP class in TYPO3 10:

https://docs.typo3.org/c/typo3/cms-core/master/en-us/Changelog/10.0/Breaking-87623-ReplaceConfigpersistenceclassesTyposcriptConfiguration.html

(Good question, if we would need both, if we support TYPO3 9+10 in the  next release).

This could be a first step to start with the domain model while keeping all other code untouched.

albig:
> One question: Why are used there AbstractValueObject instead of AbstractEntity?

In slub_digitalcollections? I only use read access to the real Kitodo table. That's maybe why. But I actually, I didn't thought about this. In Kitodo.Presentation models it should be mostly AbstractEntity, I think.

Classes/Domain/Table.php
albig: Ah, ok.

If "everything" (I doubt, that this is possible) is converted to Extbase, you do not enter the table name anywhere.

By-the-way: PHP 8 will not come to TYPO3 before version 12! Have a look at the roadmap and the PHP compatibility chart: https://typo3.org/cms/roadmap This would be Kitodo.Presentation 6.x (or more) ;-)

PR #615, opened 2021-05-10
Replace devlog with TYPO3 logger

beatrycze-volk:


Classes/Common/Helper.php
```diff
@@ -205,35 +206,26 @@ public static function decrypt($encrypted)
...
-                }
-            }
-            $arguments = '(' . implode(', ', $args) . ')';
-            $additionalData = (empty($data) ? false : $data);
-            /** @scrutinizer ignore-deprecated */ \TYPO3\CMS\Core\Utility\GeneralUtility::devLog('[' . $caller . $arguments . '] ' . $message, self::$extKey, $severity, $additionalData);
+        $logger = GeneralUtility::makeInstance(LogManager::class)->getLogger(__CLASS__);
+
+        switch ($severity) {
+            case 0:
+                $logger->info(message);
```
albig: This should be probably `$message`. Below the same error.

albig (changes requested):
Thank you for changing to the logging framework. I found one bug.

Did you notice this change from TYPO3 9.5? This would reduce the again and again MakeInstance for new logger instances:

https://docs.typo3.org/m/typo3/reference-coreapi/master/en-us/ApiOverview/Logging/Logger/Index.html

Could you please add some words about logging to the documentation? There is no "developer" section. But maybe below "System Setup" or "Administrator Manual".

albig (changes requested):
Using the indexer, I get the following exception:

```
typo3 kitodo:reindex -a -p 2 -s dlfCore0

Reindex a collection into database and Solr.
============================================
Uncaught TYPO3 Exception Call to a member function warning() on null
thrown in file /var/www/extensions/dlf/Classes/Common/Document.php
in line 720
```
Could you please have a look at it?

beatrycze-volk:
I have checked and it looks that LoggerAwareTrait doesn't function well with Document class as it has getInstance method. Logger must be initialized here directly in this method after instance is created. I have changed t to previous version and it works.

albig:
> I have checked and it looks that LoggerAwareTrait doesn't function well with Document class as it has getInstance method. Logger must be initialized here directly in this method after instance is created. I have changed t to previous version and it works.

Ok. This is of course stupid but more or less understandable. I'll have another try later.

Classes/Common/DocumentList.php
```diff
@@ -728,7 +733,10 @@ public function __construct(array $elements = [], array $metadata = [])
             // Add initial set of elements to the list.
             $this->elements = $elements;
         }
+
         $this->count = count($this->elements);
+
+        $this->logger = GeneralUtility::makeInstance(LogManager::class)->getLogger(__CLASS__);
```
albig: This will fail as LogManager is unknow in this class.

Is this still necessary with LoggerAwareInterface?

albig (changes requested):
Hopefully my last change request ;-)

Classes/Common/DocumentList.php
beatrycze-volk: No, I have removed it.

albig (approved):

PR #617, opened 2021-05-15
Remove TYPO3 8.7 specific code

albig:


Classes/Hooks/UserFunc.php
```diff
@@ -62,81 +62,4 @@ public function displayThumbnail(&$params)
         }
         return $output;
     }
-
```
beatrycze-volk, line 65: Is this part related to TYPO3 version?

albig: This getDocumentType()-function is not used in TYPO3 9.5 anymore. The protected loadDocument() was only needed by getDocumentType(). That's why I removed it.

beatrycze-volk (approved):

ByteParty (approved):

PR #623, opened 2021-05-31
Add classes for highlight of SOLR search results

beatrycze-volk:
It is preparation PR for introducing OCR SOLR Highlighting Plugin. It contains classes which are going be used for parsing and displaying the results of SOLR search. They are responding to the JSON response returned from the index.

Related to #587

### DIFF
--- a/Classes/Common/SolrSearchResult/Highlight.php
+++ b/Classes/Common/SolrSearchResult/Highlight.php
@@ -1,0 +1,185 @@
+<?php
+
+/**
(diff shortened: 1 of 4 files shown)


albig (changes requested):
Would you please drop some words in the PR description, what this PR will do, if it depends on something etc. E.g. that it is releated to the big PR of the Solr ocrhighlighting plugin. In short a motivation and a little documentation.

albig (approved):

PR #624, opened 2021-05-31
Add possibility of Index mapping

beatrycze-volk:
This is preparation PR for #587.


albig (changes requested):
My same question: What is the pull request about? Ok, the title helps. But does it break the current state? Do I have to keep something in mind, if I test this? Does it require #623 to be merged before?

beatrycze-volk:
> My same question: What is the pull request about? Ok, the title helps. But does it break the current state? Do I have to keep something in mind, if I test this? Does it require #623 to be merged before?

It doesn't break anything and doesn't need anything merged before - it is totally independent and should work without any adjustments :)

albig (approved):
Seems to work. The only thing is to save the extension configuration. Otherwise the new settings are not present.

PR #635, opened 2021-06-30
Fix error prone issues pointed out by Codacy in JS files

beatrycze-volk:
Small quality improvements

### DIFF
--- a/Resources/Public/Javascript/PageView/Utility.js
+++ b/Resources/Public/Javascript/PageView/Utility.js
@@ -760,7 +760,7 @@ dlfUtils.isCorsEnabled = function (imageObjs) {
  * @return {boolean}
  */
 dlfUtils.isWebGLEnabled = function () {
-    if (!!window.WebGLRenderingContext) {
+    if (!window.WebGLRenderingContext) {
         var canvas = document.createElement("canvas"),
             rendererNames = ["webgl", "experimental-webgl", "moz-webgl", "webkit-3d"],
             context = false;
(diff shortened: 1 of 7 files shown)


Resources/Public/Javascript/PageView/Utility.js
```diff
@@ -760,7 +760,7 @@ dlfUtils.isCorsEnabled = function (imageObjs) {
  * @return {boolean}
  */
 dlfUtils.isWebGLEnabled = function () {
-    if (!!window.WebGLRenderingContext) {
```
albig: Actually this changes logic of this function. Please don't ask me, why this is double-negated. Probably the explanation is found here: https://stackoverflow.com/questions/10467475/double-negation-in-javascript-what-is-the-purpose

This isWebGLEnabled function is used with the imagemanipulation tool. If WebGL is not avaiable, the tool is hidden.

albig (changes requested):

Resources/Public/Javascript/PageView/Utility.js
beatrycze-volk: Answer in this question is not related to this exact case, as there double negation is not inside the `if` condition. Codacy suggest that inside `if` block it is not necessary to make double negation: _In contexts such as an if statement's test where the result of the expression will already be coerced to a Boolean, casting to a Boolean via double negation (!!) or a Boolean call is unnecessary._

Should I anyway reverse this change?

albig: Yes, please revert. The suggestion from Codacy seems to be not true in this case. Have a look at this older question:

https://stackoverflow.com/questions/11871077/proper-way-to-detect-webgl-support

And have a try with this jsfiddle:

https://jsfiddle.net/albig/r58yuxmt/7/

If we keep your change, the function does not work as expected.

beatrycze-volk: Reversed

albig (approved):

PR #645, opened 2021-08-18
Change Search Forms from POST to GET

albig:
Search with HTTP POST may fail with TYPO3 9.5.

This is always the case, if the target page is cached by mistake. This
can easily happen, if a search engine or anyone else calls the target
page with some GET parameter but of course without a valid
cHash-parameter.

In this case, TYPO3 9.5 safes the page to cache including the wrong
parameters.

A search sending the parameters by POST will receive the cached page
though the parameter 'tx_dlf[id]' is passed which should disable
caching.

This is a know bug (comparable with [1]) and can be worked around by
sending the search parameters by HTTP GET.

[1] https://forge.typo3.org/issues/88631

### DIFF
--- a/Documentation/Configuration/Index.rst
+++ b/Documentation/Configuration/Index.rst
@@ -79,7 +79,7 @@ You may notice from time to time, the viewer page stays empty even though you
 pass the :code:`tx_dlf[id]` parameter.
 
 This happens, if someone called the viewer page without any parameters or with parameters
-without a valid cHash. In this case, TYPO3 saves the page to it's cache. If you call the
+without a valid cHash. In this case, TYPO3 saves the page to its cache. If you call the
 viewer page again with any parameter and without a cHash, the cached page is
 delivered.
 
--- a/Documentation/Configuration/Index.rst
+++ b/Documentation/Configuration/Index.rst
@@ -75,7 +75,7 @@ This configuration is written to *typo3conf/LocalConfiguration.php*::
 Avoid empty Workview
 ~~~~~~~~~~~~~~~~~~~~
 
-You may notice from time to time, the viewer page keeps empty even though you
+You may notice from time to time, the viewer page stays empty even though you
 pass the :code:`tx_dlf[id]` parameter.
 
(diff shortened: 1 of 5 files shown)


beatrycze-volk (dismissed):
I have applied this changes on SOLR Plugin branch and it works with GET.

Documentation/Configuration/Index.rst
```diff
@@ -36,37 +36,71 @@ After this step, the require tx_dlf_formats records are created on the root page
...
     'FE' => [
             'pageNotFoundOnCHashError' => '0',
-            'pageNotFound_handling' => '',
         ],
 
 
+Avoid empty Workview
+~~~~~~~~~~~~~~~~~~~~
+
+You may notice from time to time, the viewer page keeps empty even though you
```
sebastian-meyer:
```suggestion
You may notice from time to time, the viewer page stays empty even though you
```

Documentation/Configuration/Index.rst
```diff
@@ -36,37 +36,71 @@ After this step, the require tx_dlf_formats records are created on the root page
...
+pass the :code:`tx_dlf[id]` parameter.
+
+This happens, if someone called the viewer page without any parameters or with parameters
+without a valid cHash. In this case, TYPO3 saves the page to it's cache. If you call the
```
sebastian-meyer:
```suggestion
without a valid cHash. In this case, TYPO3 saves the page to its cache. If you call the
```

sebastian-meyer (dismissed):
Although this works for me, I consider it bad practice to circumvent bugs in TYPO3 by patching the symptoms. But I guess, we don't have much of a choice...
At least we should remember to reverse the changes as soon as the bug in TYPO3 gets fixed.

albig:
> Although this works for me, I consider it bad practice to circumvent bugs in TYPO3 by patching the symptoms. But I guess, we don't have much of a choice...
> At least we should remember to reverse the changes as soon as the bug in TYPO3 gets fixed.

That's true. This has to be verified after the switch to Extbase anyhow.

sebastian-meyer (approved):

PR #646, opened 2021-08-18
Allow new solr records even without Solr connected.

albig:
On creating a new tx_dlf_solrcores record, the corresponding core is
created on the connected Solr automatically. This is nice for the first
setup.

If you change the Solr server, move Kitodo.Presentation to a new server
or do some fancy stuff like using an external readonly Solr for search
only, you have to manipulate the tx_dlf_solrcores records 'by hand'.

This patch gives the full freedom to TYPO3 admins to create records
independantly to an established Solr connection.

As long, as there is no connection, a red banner informs the user. If
the connection is etablished later and the corresponding core is
present, the core stats is shown. This is no new feature.

### DIFF
--- a/Configuration/TCA/tx_dlf_solrcores.php
+++ b/Configuration/TCA/tx_dlf_solrcores.php
@@ -49,7 +49,6 @@ return [
                 'max' => 255,
                 'eval' => 'alphanum,nospace,required,unique',
                 'default' => '',
-                'readOnly' => true,
                 'fieldInformation' => [
                     'solrCoreStatus' => [
                         'renderType' => 'solrCoreStatus',
--- a/Classes/Common/Solr.php
+++ b/Classes/Common/Solr.php
@@ -132,6 +132,8 @@ class Solr implements LoggerAwareInterface
      */
     public static function createCore($core = '')
     {
+        $logger = GeneralUtility::makeInstance(LogManager::class)->getLogger(__CLASS__);
+
         // Get next available core name if none given.
         if (empty($core)) {
             $core = 'dlfCore' . self::getNextCoreNumber();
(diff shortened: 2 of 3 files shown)


Configuration/TCA/tx_dlf_solrcores.php
```diff
@@ -49,7 +49,6 @@
                 'max' => 255,
                 'eval' => 'alphanum,nospace,required,unique',
                 'default' => '',
-                'readOnly' => true,
```
sebastian-meyer, line 52: This allows changing `index_name` (i.e. the core name) in production, which is quite dangerous. Is this really necessary for the scenario you are trying to achieve? As far as I understand you can get all your described scenarios to work by just allowing to add cores without a working Solr connection (i. e. the changes in `DataHandler.php` above). There is no need to allow changes of already configured core names. Or am I missing something?

I mean, this works for me. I am just concerned about the removed protection against potentially dangerous changes.

sebastian-meyer (approved):
I have some concerns (see below), but generally your changes are working for me. So I'll approve this.

Configuration/TCA/tx_dlf_solrcores.php
albig: Of course this is dangerous. Or let's better say "stupid", if you change the index_name and no corresponding Solr core exists. But it doesn't unload/remove the Solr core or something else happens. You only get a red warning, that the current configuration doesn't reach a Solr core. So the risk is low.

When you need to change the `index_name`? For example 

- if you move to a new TYPO3-server, export your tree with the impexp module. All records are created on import. So you create a Solr core, too. And cannot "switch" to the right core without SQL-commands.
- if you have a read-only Solr core like the zeitungsportal one, you need to set the given `index_name` 

This is how it looked like before:

New TYPO3 site  TYPO3 CMS 9 5 30

And this happens, if someone enters a wrong `index_name`:

New TYPO3 site  TYPO3 CMS 9 5 30  (1)

sebastian-meyer: You're right! I didn't think this through all the way. 😄

PR #650, opened 2021-08-24
Extend Document class

beatrycze-volk:
This PR prepares adding of 3D model class which is going to extend Document class. 3D models don't have full text at all, so all methods related to full text are not needed for it.

- move document classes to the own package
- add abstract FullTextDocument class which extends Document class
- MetsDocument and IiifManifest classes extend now FullTextDocument
- reorder classes in Document class (first public abstract, then public / static public, then protected abstract, then protected and at the end magic methods)

For easier review is good to check first commit (moving code) and then second (reordering methods).


beatrycze-volk:

PR #669, opened 2021-09-01
Move getting of registered Document instance to own method

beatrycze-volk:
Making getInstance easier to read before adding stuff related to 3D documents


beatrycze-volk:

Issue #488, opened 2020-03-04
Support for multiple namespaces


sebastian-meyer:
Many formats used in the context of Kitodo reflect their versioning in the namespace (i. e. MODS, ALTO, IIIF, TEI). We therefore need to add default support for more than one version (although multiple versions could use the same parser).

This issue was raised by https://github.com/tesseract-ocr/tesseract/pull/2815

bertsky:
> although multiple versions could use the same parser

To that: I completely agree, and it seems at least the ALTO parser currently [does already tolerate](https://github.com/OCR-D/core/issues/544#issuecomment-868233760) multiple namespace versions. It would be great if the planned changes would hold that up as a principle. 

(And strategically, I think it is also the best choice for data providers and workflows. We wouldn't want to discourage anyone from ingesting the best possible version of their fulltext – even if the current Presentation cannot make full use of it yet. Features like polygonal regions, angle/orientation, layout tags to differentiate different text region types and image region types, baseline curves, glyphs could all prove very valuable to future versions of Presentation. Moreover, they can be immediately useful for downstream applications of researches that simply download the fulltext.)

bertsky:
@sebastian-meyer notified me (via other communication) that one necessary ingredient is v4 support in the Solr indexer.

So as long as this is not integrated/configured/tested, we still have to ingest v2, otherwise there will be full text highlighting, but no search.

sebastian-meyer:
To be precise: the problem here is not the indexer itself, because that just takes whatever text it gets. But we have parsers for every supported fulltext format (for ALTO it's https://github.com/kitodo/kitodo-presentation/blob/master/Classes/Format/Alto.php) and those are currently using hard-coded namespace URIs instead of the actual URIs configured in the format table.

Also, we have to take into account the [solr_ocrhighlighting](https://dbmdz.github.io/solr-ocrhighlighting/0.8.3/) plugin we are using, because that interprets ALTO as well in order to get the word coordinates. I am not sure which ALTO versions are supported by this plugin.

bertsky:
> Also, we have to take into account the [solr_ocrhighlighting](https://dbmdz.github.io/solr-ocrhighlighting/0.8.3/) plugin we are using, because that interprets ALTO as well in order to get the word coordinates. I am not sure which ALTO versions are supported by this plugin.

[...]

PR #673, opened 2021-09-21
Use ocr-highlighting plugin with MiniOCR

albig:
Just a PR to discuss the current state.

### DIFF
--- a/Classes/Format/Alto.php
+++ b/Classes/Format/Alto.php
@@ -47,8 +47,6 @@ class Alto implements \Kitodo\Dlf\Common\FulltextInterface
         return $rawText;
     }
 
-
-
     /**
      * This extracts the fulltext data from ALTO XML and returns it in MiniOCR format
      *
@@ -60,7 +58,6 @@ class Alto implements \Kitodo\Dlf\Common\FulltextInterface
      */
     public function getTextAsMiniOcr(\SimpleXMLElement $xml)
     {
-        $rawText = '';
         $xml->registerXPathNamespace('alto', 'http://www.loc.gov/standards/alto/ns-v2#');
 
         // get all text blocks
@@ -80,8 +77,7 @@ class Alto implements \Kitodo\Dlf\Common\FulltextInterface
                     foreach ($value->children() as $lkey => $word) {
                         if ($lkey == "String") {
                             $attributes = $word->attributes();
-                            $la =  Solr::escapeQuery((string) $attributes['CONTENT']);
-                            $newWord = $newLine->addChild('w',  htmlspecialchars(Solr::escapeQuery((string) $attributes['CONTENT'])) . ' ');
+                            $newWord = $newLine->addChild('w', htmlspecialchars(Solr::escapeQuery((string) $attributes['CONTENT'])) . ' ');
                             $newWord->addAttribute('x', (string) $attributes['HPOS'] . ' ' . (string) $attributes['VPOS'] . ' ' . (string) $attributes['WIDTH'] . ' ' . (string) $attributes['HEIGHT']);
                         }
                     }
--- a/Classes/Format/Alto.php
+++ b/Classes/Format/Alto.php
@@ -102,7 +102,7 @@ class Alto implements \Kitodo\Dlf\Common\FulltextInterface
      * @return string The parsed word extracted from attribute
      */
     private function getWord($attributes) {
-        return htmlspecialchars(Solr::escapeQuery((string) $attributes['CONTENT'])) . ' ';
+        return htmlspecialchars((string) $attributes['CONTENT']) . ' ';
     }
 
     /**
--- a/Classes/Format/Alto.php
+++ b/Classes/Format/Alto.php
@@ -12,8 +12,12 @@
 
 namespace Kitodo\Dlf\Format;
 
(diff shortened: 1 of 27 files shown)


Classes/Format/Alto.php
```diff
@@ -42,4 +46,49 @@ public function getRawText(\SimpleXMLElement $xml)
...
+     *
+     * @access public
+     *
+     * @param \SimpleXMLElement $xml: The XML to extract the raw text from
+     *
+     * @return string The unformatted fulltext in MiniOCR format
+     */
+    public function getTextAsMiniOcr(\SimpleXMLElement $xml)
+    {
+        $rawText = '';
```
beatrycze-volk: There is a complain from Scrutinizer about not used `$rawText` variable. It needs to to be removed or used in return in case blocks are empty.

Classes/Format/Alto.php
```diff
@@ -42,4 +46,49 @@ public function getRawText(\SimpleXMLElement $xml)
...
+        $xml->registerXPathNamespace('alto', 'http://www.loc.gov/standards/alto/ns-v2#');
```
beatrycze-volk, line 59: There are already documents with version 3 and 4. If I'm not mistaken Zeitungsportal has mostly them. How are we going to handle those documents?

Classes/Format/Alto.php
```diff
@@ -42,4 +46,49 @@ public function getRawText(\SimpleXMLElement $xml)
...
+
+        foreach ($blocks as $block) {
+            $newBlock = $miniOcr->addChild('b');
+            foreach ($block->children() as $key => $value) {
+                if ($key === "TextLine") {
+                    $newLine = $newBlock->addChild('l');
+                    foreach ($value->children() as $lkey => $word) {
+                        if ($lkey == "String") {
+                            $attributes = $word->attributes();
+                            $la =  Solr::escapeQuery((string) $attributes['CONTENT']);
```
beatrycze-volk: What is the purpose of this variable?

Classes/Format/Alto.php
```diff
@@ -42,4 +46,49 @@ public function getRawText(\SimpleXMLElement $xml)
...
+                            $newWord = $newLine->addChild('w',  htmlspecialchars(Solr::escapeQuery((string) $attributes['CONTENT'])) . ' ');
+                            $newWord->addAttribute('x', (string) $attributes['HPOS'] . ' ' . (string) $attributes['VPOS'] . ' ' . (string) $attributes['WIDTH'] . ' ' . (string) $attributes['HEIGHT']);
+                        }
+                    }
+                }
+            }
+        }
+
+        return $miniOcr->asXml();
```
beatrycze-volk: Shouldn't we note that in the case of error of the document creation this method can return false instead of string?

Classes/Format/Alto.php
```diff
@@ -42,4 +46,49 @@ public function getRawText(\SimpleXMLElement $xml)
         }
         return $rawText;
     }
+
```
beatrycze-volk, line 47: Probably only one empty line is needed here.

Classes/Format/Alto.php
albig: Thank you. That's fixed now. This was taken from the `getRawText()` method, I've modified before.

Classes/Format/Alto.php
albig: Of course, there is ALTO v3 and v4. But this is a another issue (#488) which has to be fixed. Maybe by different Alto classes with different "getTextAsMiniOcr()" methods.

For the DDB-Zeitungsportal is no problem here as we do not index into their Solr.

Classes/Format/Alto.php
albig: Nothing. This was a remaining debugging helper. I removed it.

Classes/Format/Alto.php
albig: Not sure, what the ocr highlighting plugin will accept. I think, we must send an empty string to it. 

Could be improved for sure.

albig: But you're right. We shouldn't return the output of `asXml()` directly.

Classes/Format/Alto.php
beatrycze-volk: Good point! Thanks! Then there is no problem.

Classes/Format/Alto.php
```diff
@@ -42,4 +46,75 @@ public function getRawText(\SimpleXMLElement $xml)
...
+                            $newWord = $newLine->addChild('w', $this->getWord($attributes));
+                            $newWord->addAttribute('x', $this->getCoordinates($attributes));
+                        }
+                    }
+                }
+            }
+        }
+
+        $miniOcrXml = $miniOcr->asXml();
+        if(\is_string($miniOcrXml)) {
```
albig:
```suggestion
        if (\is_string($miniOcrXml)) {
```

beatrycze-volk (dismissed):
I have looked over the code and tested changes. It seems to be cleaned out and working properly (indexing of the documents and searching in the document).

beatrycze-volk (dismissed):

beatrycze-volk (approved):

PR #680, opened 2021-10-06
Split getInstance method of Document class

beatrycze-volk:
Take out parts for getting document format depending on the source (database or some remote source).


beatrycze-volk:

PR #685, opened 2021-10-11
Add XmlDocument trait for Document class 

beatrycze-volk:
Depends on #650


beatrycze-volk:

PR #693, opened 2021-10-14
Fix check whether fulltext is present

dvoracek-slub:
This PR fixes a condition that checks whether fulltext is present, and refactors the condition into a utility function to avoid code duplication.

## Issue

When opening a page that does not have fulltext using the v3.3 of Kitodo.Presentation, I get two console errors:

    XML Parsing Error: mismatched tag. Expected: </link>.
    [...]

    Uncaught Error: Invalid XML: [...]

## Reproduce

For example, install the kitodo/ddev-kitodo-presentation environment and open the following link:

> https://ddev-kitodo-presentation.ddev.site/workview?tx_dlf[id]=https%3A%2F%2Fdigital.slub-dresden.de%2Foai%3Fverb%3DGetRecord%26metadataPrefix%3Dmets%26identifier%3Doai%3Ade%3Aslub-dresden%3Adb%3Aid-1665101628

## Trace

- Currently, when there is no fulltext, Kitodo tries to load fulltext from URL `undefined`.
- `$.ajax` defaults to loading the current page.
- The response text is fed to the ALTO parser.
- Choke.

### DIFF
--- a/Resources/Public/Javascript/PageView/PageView.js
+++ b/Resources/Public/Javascript/PageView/PageView.js
@@ -164,7 +164,7 @@ dlfViewer.prototype.addCustomControls = function() {
 
     // Adds fulltext behavior and download only if there is fulltext available and no double page
     // behavior is active
-    if (this.fulltexts[0] !== undefined && this.fulltexts[0].length !== 0 && this.fulltexts[0].url !== '' && this.images.length === 1) {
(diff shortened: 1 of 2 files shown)


albig:
Thank you for your fix. I'll have try soon. I remember this behavior which is present for a long time.

albig (approved):
Works like charm. Thank you!

PR #699, opened 2021-10-15
Quick Fix: Wait for `tx_dlf_viewer` when highlighting word

dvoracek-slub:
When clicking on a result in the in-document search (`SearchInDocumentTool`), I get the following following console error:

    Uncaught ReferenceError: tx_dlf_viewer is not defined
        addImageHighlight https://ddev-kitodo-presentation.ddev.site/typo3temp/assets/compressed/merged-eb6368222cc28ec61f1f1ba34ea295cc.js?1634203496:4401
        [...]

Link to reproduce: https://ddev-kitodo-presentation.ddev.site/workview?tx_dlf[id]=1&tx_dlf[highlight_word]=Zukunft

This happens because in order to highlight the search term (via `tx_dlf[highlight_word]` parameter), `addImageHighlight()` tries to access `tx_dlf_viewer`, but at that point `tx_dlf_viewer` isn't initialized yet (see `PageView::addViewerJS()`).

Since the code already has some mechanism to wait for the OL map being initialized, I just extend that to also check whether `tx_dlf_viewer` has already been set.

### DIFF
--- a/Resources/Public/Javascript/PageView/SearchInDocument.js
+++ b/Resources/Public/Javascript/PageView/SearchInDocument.js
@@ -259,7 +259,7 @@ function addImageHighlight(data) {
     data['documents'].forEach(function (element, i) {
         if(element['page'] === page) {
             if (element['highlight'].length > 0) {
-                if(tx_dlf_viewer.map != null) {
+                if (typeof tx_dlf_viewer !== 'undefined' && tx_dlf_viewer.map != null) { // eslint-disable-line camelcase
                     tx_dlf_viewer.displayHighlightWord(encodeURIComponent(getHighlights(element['highlight'])));
                 } else {
                     setTimeout(addImageHighlight, 500, data);


albig:
Interesting. I cannot reproduce this in Chromium and Firefox.

But I see in Firefox 93.0 that the SearchInDocument seems to be not stable. Only every second load the highlighting and search seems to work. Your fix doesn't help in this case. Do you observe the same strange behavior?

dvoracek-slub:
> But I see in Firefox 93.0 that the SearchInDocument seems to be not stable. Only every second load the highlighting and search seems to work. Your fix doesn't help in this case. Do you observe the same strange behavior?

Oh. So highlighting just doesn't work, without any console errors? I do have a similar-sounding behavior in Firefox 93, but I'd get the abovementioned error message whenever highlighting doesn't work.

> Interesting. I cannot reproduce this in Chromium and Firefox.

I played around a bit, and indeed I can only reproduce this reliably in Firefox with "Disable Cache" (in the Devtools Network tab) turned on. Could you try doing that and see if you get the error?

---

To explain, there seems to be a race condition:

1. `SearchInDocument.js` installs a `$(document).ready()` handler.
    1. The callback does a network request (`$.post`).
        1. On success, `addImageHighlight` is called, which relies on `tx_dlf_viewer` being set.
2. `PageView.php` injects code to install another `$(document).ready()` handler.
    1. The callback initializes `tx_dlf_viewer`.

This requires that the network response handler 1.i.a is called *after* 2.i.

dvoracek-slub:
Slightly off-topic, I had added `// eslint-disable-line camelcase` to please Codacy/ESLint, which was complaining that `tx_dlf_viewer` is not camel-cased. Could/should we configure Codacy/ESLint to allow that symbol?

beatrycze-volk (approved):
I didn't encounter this error but some check itself makes sense.

albig (approved):

PR #705, opened 2021-10-19
Simplify methods for displaying metadata in Metadata plugin

beatrycze-volk:


albig:
Thank you. I won't have a look at this before the Extbase merge is finished. So it may take a while.

albig (approved):

Issue #590, opened 2021-02-16
Add structure type "journal" and handle it like newspaper or ephemera in calendar plugin


albig:
## Description

The calendar plugin currently only works with structure type `newspaper` or `ephemera`.

To use the same structure for journals as requested in https://github.com/slub/dfg-viewer/issues/175, Kitodo.Presentation must accept `journal` in the documenType switch.

The calendar should always be used for:

newspaper | ephemera | **journal** -> year -> month -> issue

PR #710, opened 2021-10-20
Add case for journal in Calendar plugin

beatrycze-volk:
#590


albig:
I don't think, we should do this without use-case.

PR #736, opened 2021-11-12
Fix logger in MetsDocument and IiifManifest 

albig:
$this->logger is currently `null` in classes extending `Document::class`.

### DIFF
--- a/Classes/Common/Document.php
+++ b/Classes/Common/Document.php
@@ -1612,6 +1612,7 @@ abstract class Document
      */
     protected function __construct($uid, $pid, $preloadedDocument)
     {
+        $this->logger = GeneralUtility::makeInstance(LogManager::class)->getLogger();
         $queryBuilder = GeneralUtility::makeInstance(ConnectionPool::class)
             ->getQueryBuilderForTable('tx_dlf_documents');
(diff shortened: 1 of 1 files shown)


ByteParty (approved):

beatrycze-volk:
Should we in that case not remove this line
`$instance->logger = GeneralUtility::makeInstance(LogManager::class)->getLogger(get_class($instance));`
from `getInstace` method?

albig:
> Should we in that case not remove this line `$instance->logger = GeneralUtility::makeInstance(LogManager::class)->getLogger(get_class($instance));` from `getInstace` method?

If this still works. I haven't thought about it. My issue was the missing instance in MetsDocument class.

Have a try!

beatrycze-volk:
> > Should we in that case not remove this line `$instance->logger = GeneralUtility::makeInstance(LogManager::class)->getLogger(get_class($instance));` from `getInstace` method?
> 
> If this still works. I haven't thought about it. My issue was the missing instance in MetsDocument class.

It looks that this line cannot be removed. After I have deleted it, logger doesn't work anymore.

beatrycze-volk (approved):

PR #744, opened 2021-11-23
Fix deprecation code for Typo3 v10

christopher-timm:


albig (changes requested):
Indexing and reindexing works with my patch. See PR https://github.com/chrizzor/kitodo-presentation/pull/8.

Opening the Settings->Extension Configuration menu fails with the following message in 10.4.:

New TYPO3 site  TYPO3 CMS 10 4 21  (1)

christopher-timm:
The extension configuration menu shouldn't fail anymore.
It seems to me, that the manual bootstrapping is not needed in version 10.

albig (changes requested):
New TYPO3 site  TYPO3 CMS 10 4 21  (3)

albig (approved):

Issue #747, opened 2021-12-21
Search in Document tool creates invalid links if using TYPO3-Slugs


albig:
## Description

With release 3.3 the links inside the textsnippets of the search in document tools are not valid if you have configured TYPO3 Slugs for the `tx_dlf[page]` parameter.

The error message is 

```
(1/1) #1518472189 TYPO3\CMS\Core\Error\Http\PageNotFoundException
The requested URL is not distinct
```

## Reproduction

Slug configuration in SiteConfiguration

```
routeEnhancers:
  KitodoWorkview:
    type: Plugin
    namespace: tx_dlf
    limitToPages:
      - 123 # (uid of workview)
    routePath: '/dlf/{id}/{page}'
    requirements:
      id: '(\d+)|(http.*xml)'
      page: \d+
```

Page of Workview with document uid=72043:

https://digital.slub-dresden.de/werkansicht/dlf/72043/1

Search inside the document, the hits are linked like this:

https://digital.slub-dresden.de/werkansicht/dlf/72043/1?tx_dlf[highlight_word]=Pieschen&tx_dlf[page]=3

TYPO3 now receives the tx_dlf[page] parameter twice: once by routing configuration (current page=1) and once by GET parameter (page=3). This is not accepted and an the PageNotFoundException exception is thrown.

See: https://github.com/TYPO3/typo3/blob/9.5/typo3/sysext/frontend/Classes/Middleware/PageResolver.php#L134

## Expected Behavior

The `baseUrl` for the text snippets in common Kitodo.Presentation usage must be taken from the action URL of the search form. This was the way [Kitodo.Presentation 3.2](https://github.com/kitodo/kitodo-presentation/blob/v3.2.2/Resources/Public/Javascript/PageView/SearchInDocument.js#L74) has been working.

In the example above, the `baseUrl` is https://digital.slub-dresden.de/werkansicht. All parameters have to be appended like the navigation plugin does. The resulting link must be

https://digital.slub-dresden.de/werkansicht?tx_dlf[id]=72043&tx_dlf[highlight_word]=Pieschen&tx_dlf[page]=3

I'm not sure about the impact to the DDB-Zeitungsportal.

## Environment

 - TYPO3 version: [e.g. 31]
 - Kitodo.Presentation 3.3.1
## Additional Context

This behaviour of TYPO3 won't change in TYPO3 10.4.

beatrycze-volk:
[...]

PR #748, opened 2021-12-22
Fix link creation in search in document tool if using TYPO3 speaking urls (slug).

albig:
This would fix probably #747

### DIFF
--- a/Resources/Public/Javascript/PageView/SearchInDocument.js
+++ b/Resources/Public/Javascript/PageView/SearchInDocument.js
@@ -255,7 +255,8 @@ function getNavigationButtons(start, numFound) {
  */
 function getCurrentPage() {
     var page = 1;
-    var queryParams = getCurrentQueryParams(getBaseUrl(" "));
+    var baseUrl = getBaseUrl(" ");
+    var queryParams = getCurrentQueryParams(baseUrl);
     var pageFound = false;
 
     for(var i = 0; i < queryParams.length; i++) {
@@ -271,7 +272,7 @@ function getCurrentPage() {
         var url = baseUrl.split('/');
         page = url.pop();
     }
-
+    console.log('Page for highlight: ' + page);
     return page;
 }
 
--- a/Resources/Public/Javascript/PageView/SearchInDocument.js
+++ b/Resources/Public/Javascript/PageView/SearchInDocument.js
@@ -272,7 +272,6 @@ function getCurrentPage() {
         var url = baseUrl.split('/');
         page = parseInt(url.pop(), 10);
     }
-    console.log('Page for highlight: ' + page);
     return page;
 }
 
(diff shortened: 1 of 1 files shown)


beatrycze-volk (dismissed):

beatrycze-volk (dismissed):

Resources/Public/Javascript/PageView/SearchInDocument.js
```diff
@@ -234,15 +256,22 @@ function getNavigationButtons(start, numFound) {
...
         var queryParam = queryParams[i].split('=');
 
         if(decodeURIComponent(queryParam[0]) === $("input[id='tx-dlf-search-in-document-page']").attr('name')) {
             page = parseInt(queryParam[1], 10);
+            pageFound = true;
         }
     }
 
+    if (!pageFound) {
+        var url = baseUrl.split('/');
```
albig, line 272: Unfortunately, this code fails at this point: 

```
SearchInDocument.js?1640205096:271 Uncaught ReferenceError: baseUrl is not defined
    at getCurrentPage (SearchInDocument.js?1640205096:271)
    at addImageHighlight (SearchInDocument.js?1640205096:286)
    at Object.success (SearchInDocument.js?1640205096:366)
    at c (jquery-3.5.1.min.js?1605185398:2)
    at Object.fireWith [as resolveWith] (jquery-3.5.1.min.js?1605185398:2)
    at l (jquery-3.5.1.min.js?1605185398:2)
    at XMLHttpRequest.<anonymous> (jquery-3.5.1.min.js?1605185398:2)
```

albig: If I add  `var baseUrl = getBaseUrl(" ");` above, it doesn't failt. But the highlighting isn't shown.

albig (commented):
Please have a look at my comments. It doesn't work yet for me :-(

Resources/Public/Javascript/PageView/SearchInDocument.js
beatrycze-volk: I have added logger for page. Could you tell me for which page highlighting is performed? (I have no test system with configured slugs, so I'm testing with use of fake links and there it looks correct) I was assuming that problem with highlight after link correction comes from incorrect page number taken by highlight method. This page is checked out by getCurrentPage() method.

Resources/Public/Javascript/PageView/SearchInDocument.js
```diff
@@ -233,16 +255,24 @@ function getNavigationButtons(start, numFound) {
...
+        page = parseInt(url.pop(), 10);
+    }
+    console.log('Page for highlight: ' + page);
```
albig:
```suggestion
```

albig (commented):
This works for me with and without Slug configuration.

Please just remove the console.log().

beatrycze-volk (approved):

PR #750, opened 2022-01-07
Fix some issues with default owner, metadataformat, logger

albig:


Classes/Common/Doc.php
```diff
@@ -1190,8 +1190,6 @@ protected function __construct($location, $pid, $preloadedDocument)
         $this->setPreloadedDocument($preloadedDocument);
         $this->init();
         $this->establishRecordId($pid);
-        $this->logger = GeneralUtility::makeInstance(LogManager::class)->getLogger();
-
```
dvoracek-slub, line 1194: As long as we can't inject the logger, shouldn't we rather set it at the top of the constructor and __wakeup? Then we won't need to set it in init().

Classes/Common/MetsDocument.php
```diff
@@ -709,7 +647,7 @@ protected function init()
             // Register namespaces.
             $this->registerNamespaces($this->mets);
         } else {
-            $this->logger->error('No METS part found in document with UID ' . $this->uid);
+            $this->logger->error('No METS part found in document with location "' . $this->location . '"');
```
dvoracek-slub: As far as I see, currently `$this->location` isn't being set. Probably it could be set in the constructor from the `$location` argument.

Classes/Controller/TableOfContentsController.php
```diff
@@ -44,7 +44,9 @@ class TableOfContentsController extends AbstractController
     public function __construct()
     {
         // Read plugin TS configuration.
+        // TODO: This is more or less obsolete - the table of document plugin does not use the TypoScript anymore
         $this->pluginConf = $GLOBALS['TSFE']->tmpl->setup['plugin.']['tx_dlf_tableofcontents.'];
+        $this->initialize();
```
dvoracek-slub: Since `initialize()` is called in `AbstractController::__construct`, it seems to make sense to just call `parent::__construct()` at the top of this constructor.

Classes/Common/MetsDocument.php
```diff
@@ -589,70 +590,6 @@ class_exists($class)
...
-                    ->where(
-                        $queryBuilder->expr()->eq('tx_dlf_documents.pid', intval($cPid)),
-                        $queryBuilder->expr()->eq('tx_dlf_documents.uid', intval($this->uid))
-                    )
-                    ->execute();
-
-                $resArray = $result->fetch();
-
-                $metadata['owner'][0] = $resArray['owner'];
-            }
```
dvoracek-slub, line 655: Why remove this instead of adding checks as in #752? (Though I guess the DB queries should be extracted to DocumentRepository.)

Classes/Controller/TableOfContentsController.php
albig: Good point. I'll do like this.

Classes/Common/MetsDocument.php
albig: Actually, we have no uid at this point anymore. The `Doc::class` should only keep the information from the METS-file. Everything else should come from `Document::class` which has access to the `DocumentRepository`. 

So we can't execute the queries at this point anyhow and we can't even use the documentRepository. And I think it's not necessary at all.

Classes/Common/MetsDocument.php
albig: Your're right.

Classes/Common/Doc.php
albig: The problem with this logger is:
* it should be already available (in Extbase context implmenting the `LoggerAwareInterface`) but this does not work here.
* the `__wakeup` is only called, if the XML comes from Cache (means get's deserialized).
This is for sure not the finaly beauty. But for now...

dvoracek-slub: Yes. I wondered if it would still be slightly cleaner to logger instantiation to the outermost position that works.

albig: I would prefer this, too. But I didn't find a better place to make it work in all situations... I wouldn't search any further because I hope to find a better solution later on.

dvoracek-slub (approved):

PR #753, opened 2022-01-08
It may happen, that fulltext files are empty. In this case, indexing should not abort with exception.

albig:
Example with empty full text files (page >10): https://digital.slub-dresden.de/data/kitodo/vollgesc_393504948/vollgesc_393504948_mets.xml

@andre-hohmann: Could you fix this example anyway?


albig:
This fix is not yet sufficient. It should be checked, if valid XML is returned.

Classes/Common/Document.php
```diff
@@ -668,7 +668,7 @@ protected function getFullTextFromXml($id)
                 if (!empty($this->physicalStructureInfo[$id]['files'][$fileGrpFulltext])) {
                     // Get full text file.
                     $fileContent = GeneralUtility::getUrl($this->getFileLocation($this->physicalStructureInfo[$id]['files'][$fileGrpFulltext]));
-                    if ($fileContent !== false) {
+                    if ($fileContent !== false && !empty($fileContent)) {
```
ByteParty: false is also empty
if (!empty($fileContent)) {

ByteParty (approved):

PR #756, opened 2022-01-10
Bootstrap unit testing

dvoracek-slub:
This PR bootstraps unit and functional testing, adds demo tests and integrates with GitHub workflows. Generally speaking, I took https://docs.typo3.org/m/typo3/reference-coreapi/9.5/en-us/Testing/ExtensionTesting.html and the TYPO3 core as a basis.

Some questions:
- The manual suggests to add a `.Build` folder for `vendor`, `public` etc. I didn't do this for now, mostly to stay more in line with existing conventions. On the other hand, this way the root directory becomes more cluttered, and I can't use the ExtensionTestEnvironment from testing-framework. Any opinion on this?
- Is it okay to integrate with GitHub CI, and does the config look sane?
- Should METS files for testing be hosted locally, or is it okay to access digital.slub-dresden.de?
- In TYPO3, they also include linting, ReST doc validation etc. within the setup. Add any of those?


Classes/Common/Helper.php
```diff
@@ -208,6 +208,12 @@ public static function decrypt($encrypted)
      */
     public static function getXmlFileAsString($content)
     {
+        // Don't make simplexml_load_string throw (when $content is an array
+        // or object)
+        if (!is_string($content)) {
```
beatrycze-volk, line 218: Here could be fixed the PHPDoc, curently it states: _Get content of XML file as string or false if there is nothing._ In fact it is _XML file as SimpleXMLElement_ and _@return mixed_ could be changed to _@return SimpleXMLElement|false_.
 
In general I really like your idea to introduce automatic testing! It is really cool!

dvoracek-slub: Yes, absolutely! I updated the PHPDoc.

albig:
> This PR bootstraps unit testing, adds a demo test and integrate with GitHub workflows. Generally speaking, I took https://docs.typo3.org/m/typo3/reference-coreapi/9.5/en-us/Testing/ExtensionTesting.html and the TYPO3 core as a basis.

I think testing should be implemented for 10.4. Bye the way a cool idea. But I haven't the time yet to try it out.

I assume, the 10.4 documentation is not quite different, isn't it?

https://docs.typo3.org/m/typo3/reference-coreapi/10.4/en-us/Testing/ExtensionTesting.html

> * The manual suggests to add a `.Build` folder for `vendor`, `public` etc. I didn't do this for now, mostly to stay more in line with existing conventions. On the other hand, this way the root directory becomes more cluttered, and I can't use the ExtensionTestEnvironment from testing-framework. Any opinion on this?

I have no opinion because I have no experience at all.

> * Why does `.gitiginore` currently only exclude `public/typo3/` instead of the whole `public/` folder?

Probably from former times (TYPO3 6.2/7.6). If you didn't used a composer installation, you had to run composer in the extension directory, which created `public/typo3`. So `public/` should be correct in `.gitignore`.

> * Is it okay to integrate with GitHub CI, and does the config look sane?

I'lll have a look later on. Maybe you could present / explain this work. Could be faster than trying by myself.

dvoracek-slub:
For anyone willing to try out, I added two kinds of tests:
- Functional tests, which can access a database hosted in Docker. The sample test involves reading a METS document.
- Acceptance tests, which run on a full installation of TYPO3. In the sample test, I use this to access the PageViewProxy API.

dvoracek-slub:
Latest changes:

- Extended usage information
- Added composer test commands (e.g. for functional tests)
- Added watch mode to automatically re-run tests when test files change
- Added test and data for having document in collection
- Added Solr docker service and first tests for Solr and indexing

dvoracek-slub:
Latest changes:
- Added Solr to acceptance test container,
  added tests for OAI-PMH interface
- Some cleanup

dvoracek-slub:
- Rebase to latest `dev-extbase-fluid`, cleanup commit history.
- Simplify test setup by merging API tests (PageViewProxy, OAI-PMH) into the "Functional" category, remove "Acceptance" for now.
- Add some more tests to PageViewProxy.
- Extend GitHub CI to run all tests, and to run them against both v9.5 and v10.4.
- Various minor changes.

albig (commented):
I have some trouble to run the tests on my system. Maybe you could clarify, how-to do it.

Where is my working directory? In the extension directory, I suppose. In my case it's `/home/ab/extensions/dlf`.

Now, I fail with the dependancies. My local PHP is far (8.1.3) away from the target PHP (7.4).

```
[ab@alibo dlf]$ composer update --with=typo3/cms-core:^10.4
Loading composer repositories with package information
Info from https://repo.packagist.org: #StandWithUkraine
Updating dependencies
Your requirements could not be resolved to an installable set of packages.

  Problem 1
    - Root composer.json requires php ^7.3|^7.4 but your php version (8.1.3) does not satisfy that requirement.
  Problem 2
    - typo3/cms-core[v10.4.0, ..., v10.4.26] require php ^7.2 -> your php version (8.1.3) does not satisfy that requirement.
    - Root composer.json requires typo3/cms-core ^9.5|^10.4, ^10.4 -> satisfiable by typo3/cms-core[v10.4.0, ..., v10.4.26].
```

Then I went into the DDEV container in `/var/www/html/public/typo3conf/ext/dlf`. Installing dependancies with `composer update --with=typo3/cms-core:^10.4` is no problem here. But running the tests fails with the following message:

```
ab@ddev-kitodo-presentation-web:/var/www/html/public/typo3conf/ext/dlf$ composer test
> phpunit -c Build/Test/UnitTests.xml
PHPUnit 9.5.19 #StandWithUkraine

..                                                                  2 / 2 (100%)

Time: 00:00.006, Memory: 8.00 MB

OK (2 tests, 16 assertions)
> Build/Test/runTests.sh -s functional
Build/Test/runTests.sh: line 157: type: docker-compose: not found
This script relies on docker and docker-compose. Please install
Script Build/Test/runTests.sh -s functional handling the test:func event returned with error code 1
Script @test:func was called via test
```

Going back to my host environment, it's different but not better:

```
[ab@alibo dlf]$ composer test
> phpunit -c Build/Test/UnitTests.xml
PHPUnit 9.5.19 #StandWithUkraine

..                                                                  2 / 2 (100%)

Time: 00:00.005, Memory: 10.00 MB

OK (2 tests, 16 assertions)
> Build/Test/runTests.sh -s functional
docker-compose 2.3.3 is currently broken and not supported by runTests.sh.
If you are running Docker Desktop for MacOS/Windows disable 'Use Docker Compose V2 release candidate' (Settings > Experimental Features)
Script Build/Test/runTests.sh -s functional handling the test:func event returned with error code 1
Script @test:func was called via test
```

dvoracek-slub:
Thanks for the report. Would you have a look again?

Right now, some tests fail because digital.slub-dresden.de returns 503 due to maintenance. Do you think I should rather copy the METS files to the repository and not rely on an external server?

---

- When using PHP 8, please install dependencies using the new "install-via-docker" script:
  ```
  composer install-via-docker -- -t 9.5
  ```
  This runs "composer update" inside Docker, which previously I only used for GitHub CI. (I also tried `--ignore-platform-req=php`, but then it installs a version of "psr/cache" that isn't compatible with PHP 7.4, so e.g. the Dockerized tests would fail.)

- I removed the version check for docker-compose. v2 seems to work.

albig (approved):
I could run all the documented test commands on my local linux system. So this seems to be perfect for me and really fancy.

Usurally, you can rely on METS files from digital.slub-dresden.de. It's really rare case, that it is not available.

PR #757, opened 2022-01-12
Fixes to Extbase/Fluid transition

dvoracek-slub:
Work-in-progress of various fixes related to the dev-extbase-fluid branch.


Classes/Command/BaseCommand.php
```diff
@@ -81,9 +82,12 @@ protected function initializeRepositories($storagePid)
             $frameworkConfiguration['persistence']['storagePid'] = MathUtility::forceIntegerInRange((int) $storagePid, 0);
             $configurationManager->setConfiguration($frameworkConfiguration);
 
-            $this->collectionRepository = GeneralUtility::makeInstance(CollectionRepository::class);
-            $this->documentRepository = GeneralUtility::makeInstance(DocumentRepository::class);
-            $this->libraryRepository = GeneralUtility::makeInstance(LibraryRepository::class);
+            // TODO: When we drop support for TYPO3@9, we needn't/shouldn't use ObjectManager anymore
+            $objectManager = GeneralUtility::makeInstance(ObjectManager::class);
+
```
albig, line 87: Why do you switch to use the objectManager here? I don't see any advantage right now.

Classes/Common/Helper.php
```diff
@@ -1068,5 +1070,39 @@ public static function getLanguageService(): LanguageService
...
+     * Make classname configuration from `Classes.php` available in contexts
+     * where it normally isn't, and where the classical way via TypoScript won't
+     * work either.
+     *
+     * @deprecated Remove once we drop support for TYPO3@9
+     *
+     * @access public
+     */
+    public static function polyfillExtbaseClassesForTYPO3v9()
+    {
```
albig, line 1086: Looks complicate. What is this good for? Where do we need it?

albig: The table mapping is done in TYPO3 9 in `ext_typoscript_setup.txt`. In TYPO3 10 it is done in `Configuration/Extbase/Persistence/Classes.php`. Both should coexist without problems - I thought.

Classes/Hooks/DataHandler.php
```diff
@@ -38,20 +39,11 @@ class DataHandler implements LoggerAwareInterface
     use LoggerAwareTrait;
 
     /**
+     * @Inject
```
albig: I wonder about the @Inject annotation. Does this really work at this place?

albig (commented):
Thank you for your work. I haven't try it out but some - maybe stupid - questions.

Classes/Command/BaseCommand.php
dvoracek-slub: AFAIK, that's needed in TYPO3 v9 (in my v9 setup, it didn't work before). The reason is that the repositories take an `ObjectManager` in their constructor, which is automatically injected via `makeInstance` only from v10 on. A variant of this might be to pass `objectManager` to `makeInstance`.

For reference, v10 has some DI code here:
https://github.com/TYPO3/typo3/blob/10.4/typo3/sysext/core/Classes/Utility/GeneralUtility.php#L3471-L3477

It's not present in the v9 version:
https://github.com/TYPO3/typo3/blob/9.5/typo3/sysext/core/Classes/Utility/GeneralUtility.php#L3690

Classes/Common/Helper.php
dvoracek-slub: They probably can coexist, but in v9 neither works for use in expression language functions (`DocumentTypeFunctionProvider`). In testing I got an error that the table `tx_dlf_domain_model_documents` does not exist, i.e., it took the default tablename instead of reading the configuration.

I don't know enough about TYPO3 to tell exactly why, but it's probably related to expression language working within TypoScript as well. I guess that `ext_typoscript_setup.txt` is loaded relatively early, but perhaps it hasn't yet been processed by Extbase when the `getDocumentType()` is evaluated within another TypoScript.

Classes/Hooks/DataHandler.php
dvoracek-slub: No it doesn't. :-/

Classes/Command/BaseCommand.php
albig: Ah, ok. I'll must check with TYPO3 9 again. Currently my development system is TYPO3 10.

Classes/Common/Helper.php
albig: I remember this behaviour. And I think, there was in the meantime in `DocumentTypeFunctionProvider` a patch for the mapping which I might have removed again. I'll will check again.

Classes/Hooks/DataHandler.php
albig: :-)

Classes/Common/Helper.php
dvoracek-slub: In 5e6110fb you added some code to `Document::getInstance()` for that (and removed it later). I basically generalized this to take all classnames from `Classes.php`.

albig: Exactly. That's what I meant. Ok, I understand your code now.

We could drop all this, if we rename the tables to extbase default naming scheme (tx_dlf_domain_model_document for tx_dlf_document). Maybe you could add a comment about this. We can discuss this before 4.0 release.

dvoracek-slub: I wasn't sure what the best place would be. For now, I put some remarks into a new file called "DEVELOPMENT.md".

Classes/ViewHelpers/ParseMetadataWrapViewHelper.php
```diff
@@ -0,0 +1,34 @@
...
+use TYPO3Fluid\Fluid\Core\Rendering\RenderingContextInterface;
+use TYPO3Fluid\Fluid\Core\ViewHelper\AbstractViewHelper;
+use TYPO3\CMS\Core\TypoScript\Parser\TypoScriptParser;
+use TYPO3\CMS\Core\Utility\GeneralUtility;
+use TYPO3\CMS\Extbase\Configuration\ConfigurationManager;
+
+class ParseMetadataWrapViewHelper extends AbstractViewHelper
+{
+    protected $escapeOutput = false;
+
```
albig: Please add the PHPdocs like e.g. in slub_digitalcollections ViewHelpers. Or have a look at TYPO3 core for more examples.

Classes/ViewHelpers/ParseMetadataWrapViewHelper.php
```diff
@@ -0,0 +1,34 @@
...
+    public static function renderStatic(array $arguments, \Closure $renderChildrenClosure, RenderingContextInterface $renderingContext)
+    {
+        $parser = GeneralUtility::makeInstance(TypoScriptParser::class);
+        $parser->parse($renderChildrenClosure());
+        return [
```
albig: Not sure how to fix this return value. Usually it's string. Please think again about it.

albig (changes requested):
Accept my comments, I can confirm it is working and I would merge it in a next step.

Classes/ViewHelpers/ParseMetadataWrapViewHelper.php
dvoracek-slub: Oh I see. Please have a look at my fix, where I combine TypoScript parsing and variable creation into a single viewhelper.

Classes/Common/Helper.php
albig: Unfortunately, this change has a side effect with TYPO3 9.

* go to the backend module and edit a document of type "year" (e.g. uid=15 if you are using the same data sets)
* save the record

This will cause the following error:

New TYPO3 site  TYPO3 CMS 9 5 31  (1)

The "missing" table looks like a mapping fake. It's true, that this table does not exists.

This does not happen without your PR and TYPO3 9.
This does not happen with your PR and TYPO3 10.

Hope, you can reproduce this behavior.

albig: Very strange. I can't reproduce the behavior from last friday anymore :-(

albig (approved):

PR #758, opened 2022-01-13
Speedup PageViewProxy

dvoracek-slub:
This PR optimizes the PageViewProxy. While we may still want to switch to a proper image proxy eventually (I suppose), this could at least improve the current situation.

The speedup will be roughly to the amount that it takes for the proxy to connect to and download the data from the target. So this may help whenever the target is slow to respond, or the connection between proxy and target is slow. 

Sample URLs:
- Original image (on my setup, this has a TTFB of ~650ms)
  https://pic.sub.uni-hamburg.de/PPN1012341925_18871104/2000/0/00000001.tif (from [Zeitungsportal](https://www.deutsche-digitale-bibliothek.de/newspaper/item/3K75DNS4VPZ56Y6YLVNLJJWITTHCEWLP?fromDay=1&fromMonth=1&fromYear=1887&toDay=31&toMonth=12&toYear=1887&hit=1&issuepage=1))
- Proxied URL
  https://ddev-kitodo-presentation.ddev.site/?eID=tx_dlf_pageview_proxy&url=https%3A%2F%2Fpic.sub.uni-hamburg.de%2FPPN1012341925_18871104%2F2000%2F0%2F00000001.tif

There are two main changes:
- The previous version called `GeneralUtility::getUrl()` twice, once for headers and once for data. This is replaced with just once request (though I also use GuzzleHttp directly). This should help for cases with high TTFB, where even a HEAD request may be slow.
- Previously, the proxy would first download all content and only then start to send it to its client. This is replaced with a streaming approach, where data is fetched/sent in chunks. This should help for cases with low throughput between target and proxy.

### DIFF
--- a/Classes/Plugin/Eid/PageViewProxy.php
+++ b/Classes/Plugin/Eid/PageViewProxy.php
@@ -15,6 +15,7 @@ namespace Kitodo\Dlf\Plugin\Eid;
 use Kitodo\Dlf\Common\Helper;
 use Psr\Http\Message\ResponseInterface;
 use Psr\Http\Message\ServerRequestInterface;
+use TYPO3\CMS\Core\Configuration\ExtensionConfiguration;
 use TYPO3\CMS\Core\Http\JsonResponse;
 use TYPO3\CMS\Core\Http\RequestFactory;
 use TYPO3\CMS\Core\Http\Response;
@@ -35,9 +36,15 @@ class PageViewProxy
      */
     protected $requestFactory;
 
+    /**
+     * @var mixed
+     */
+    protected $extConf;
+
     public function __construct()
     {
         $this->requestFactory = GeneralUtility::makeInstance(RequestFactory::class);
+        $this->extConf = GeneralUtility::makeInstance(ExtensionConfiguration::class)->get('dlf');
     }
 
     protected function overwriteHeaders(array $headers)
@@ -73,10 +80,15 @@ class PageViewProxy
 
         try {
             $response = $this->requestFactory->request($url, 'GET', [
+                'headers' => [
+                    'User-Agent' => $this->extConf['useragent'] ?? 'Kitodo.Presentation Proxy',
+                ],
+
                 // For performance, don't download content up-front. Rather, we'll
                 // download and upload simultaneously.
                 // https://docs.guzzlephp.org/en/6.5/request-options.html#stream
                 'stream' => true,
+
                 // Don't throw exceptions when a non-success status code is
                 // received. We handle these manually.
                 'http_errors' => false,
--- a/Classes/Plugin/Eid/PageViewProxy.php
+++ b/Classes/Plugin/Eid/PageViewProxy.php
@@ -15,8 +15,11 @@ namespace Kitodo\Dlf\Plugin\Eid;
 use Kitodo\Dlf\Common\Helper;
 use Psr\Http\Message\ResponseInterface;
 use Psr\Http\Message\ServerRequestInterface;
+use TYPO3\CMS\Core\Http\JsonResponse;
+use TYPO3\CMS\Core\Http\RequestFactory;
 use TYPO3\CMS\Core\Http\Response;
 use TYPO3\CMS\Core\Utility\GeneralUtility;
+
 /**
  * eID image proxy for plugin 'Page View' of the 'dlf' extension
  *
@@ -28,6 +31,26 @@ use TYPO3\CMS\Core\Utility\GeneralUtility;
 class PageViewProxy
 {
     /**
+     * @var RequestFactory
+     */
+    protected $requestFactory;
+
+    public function __construct()
+    {
+        $this->requestFactory = GeneralUtility::makeInstance(RequestFactory::class);
+    }
+
+    protected function overwriteHeaders(array $headers)
+    {
+        header_remove();
+        foreach ($headers as $key => $value) {
+            // The `header()` function should already make sure this is safe
+            // (Won't let $value containing newline pass.)
+            @header($key . ': ' . $value);
+        }
+    }
+
+    /**
      * The main method of the eID script
      *
      * @access public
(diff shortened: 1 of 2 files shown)


albig (commented):
I like your idea very much, but didn't try it yet.

Could you please have a look at https://docs.typo3.org/m/typo3/reference-coreapi/9.5/en-us/ApiOverview/Http/Index.html and the usage e.g. in [slub_digitalcollections](https://github.com/slub/slub_digitalcollections/blob/master/Classes/Domain/Repository/KitodoDocumentRepository.php#L62) and the TYPO3 core?

I would prefer, not to use Guzzle directly. Maybe we need it, because of the streaming feature. But please verify your solution.

What I see as ToDo is in fact to replace _all_ `getUrl()` by a proper middleware usage. In this case we can bring back the lost feature to set the user agent (we have a Kitodo.Presentation the constant setting `useragent` wich has currently no effect).

dvoracek-slub:
> I like your idea very much, but didn't try it yet.
> 
> Could you please have a look at https://docs.typo3.org/m/typo3/reference-coreapi/9.5/en-us/ApiOverview/Http/Index.html and the usage e.g. in [slub_digitalcollections](https://github.com/slub/slub_digitalcollections/blob/master/Classes/Domain/Repository/KitodoDocumentRepository.php#L62) and the TYPO3 core?
> 
> I would prefer, not to use Guzzle directly. Maybe we need it, because of the streaming feature. But please verify your solution.

I removed the reference to Guzzle by using a `RequestFactory`. The options still are Guzzle-specific, but I agree it's better. (And, it turns out, the call to Guzzle's `stream_for` utility has been superfluous.)

> What I see as ToDo is in fact to replace _all_ `getUrl()` by a proper middleware usage.

Yes, replacing calls to `getUrl()` is something I have on my list. I think I briefly mentioned this to you – perhaps, when replacing `getUrl()`, we can use the opportunity to parallelize network requests when indexing.

> In this case we can bring back the lost feature to set the user agent (we have a Kitodo.Presentation the constant setting `useragent` wich has currently no effect).

I now added the User-Agent to the proxy requests.

Classes/Plugin/Eid/PageViewProxy.php
```diff
@@ -26,6 +31,48 @@
...
+        $this->extConf = GeneralUtility::makeInstance(ExtensionConfiguration::class)->get('dlf');
+    }
+
+    protected function overwriteHeaders(array $headers)
+    {
+        header_remove();
+        foreach ($headers as $key => $value) {
+            // The `header()` function should already make sure this is safe
+            // (Won't let $value containing newline pass.)
+            @header("$key: $value");
```
albig:
```suggestion
            header($key . ': ' . $value);
```
Personally, I don't like variables inside string. Not sure, if this is somewhere written in the coding guidelines. At least, it is not common in TYPO3.

Classes/Plugin/Eid/PageViewProxy.php
```diff
@@ -26,6 +31,48 @@
  */
 class PageViewProxy
 {
+    /**
+     * @var RequestFactoryInterface
```
albig:
```suggestion
     * @var RequestFactory
```

Classes/Plugin/Eid/PageViewProxy.php
```diff
@@ -37,46 +84,63 @@ class PageViewProxy
...
+
+            // Stream proxied content in chunks of 8KB
+            $outStream = $response->getBody();
+            while (!$outStream->eof()) {
+                echo $outStream->read(8 * 1024);
+            }
         }
-        return $response;
+
+        exit;
```
albig: Better return? the PHPdocs for the return value must be fixed, too.

albig (commented):
Find some suggestions.

Classes/Plugin/Eid/PageViewProxy.php
dvoracek-slub: Thanks for probing into this. It turns out that by implementing `SelfEmittableStreamInterface`, we can implement streaming without breaking the encapsulation.

Classes/Plugin/Eid/PageViewProxy.php
dvoracek-slub: Alright, [the TYPO3 guideline is to use single quotes only](https://docs.typo3.org/m/typo3/reference-coreapi/main/en-us/CodingGuidelines/CglPhp/PhpFileFormatting/PhpSyntaxFormatting.html#strings).

Classes/Plugin/Eid/PageViewProxy.php
dvoracek-slub: Why would you prefer this?

dvoracek-slub:
I added handling of OPTIONS request and sending a Content-Length header if possible, which may later be useful for a loading indicator. If you think they are out scope, I can of course revert.

Classes/Plugin/Eid/PageViewProxy.php
albig: I tried to get some scrutinizer hints away. You assign an instance of `RequestFactory::class ` to this variable. That's why.

albig (commented):
Looks fancy to me. If the OPTIONS / pre-flight feature really works, this would be pretty cool.

Do you have any test cases for this?

dvoracek-slub:
> Looks fancy to me. If the OPTIONS / pre-flight feature really works, this would be pretty cool.
> 
> Do you have any test cases for this?

I have only tested this manually yet (curl + browser), and it seems to work. I could try to add test cases to the testing PR, though.

dvoracek-slub:
To get started, I added two test cases in #756. These are currently based on the dev-extbase-fluid branch, but also pass when using the code of this PR. (Of course, this could/should be extended to cover more cases.)

albig:
Would you please rebase this PR? We dropped the usage of the header parameter in recent commit https://github.com/kitodo/kitodo-presentation/commit/04df24c2943b2ec763e2ff7a6ad3ba774ac7bbf3.

dvoracek-slub:
I rebased to current master and cleaned up commit history a little. (It should be possible to merge into `dev-extbase-fluid` without problems, though.)

albig (approved):
Looks good to me. Loading of images is faster on certain conditions with this approach.

PR #761, opened 2022-01-26
Fix adding default formats for TYPO3 v10

dvoracek-slub:
In TYPO3 v10, I had the following issues:
- On a fresh install where `tx_dlf_formats` is empty, in the extension configuration I get the error
  ```
  Call to a member function isAdmin() on null
  ```
  in Helper.php line 800:
  ```php
  && $GLOBALS['BE_USER']->isAdmin()
  ```
- Also, when entries *are* present, the `deleted` field has not been considered.

---

The origin of this is the `ConfigurationForm::checkMetadataFormats()` hook.

- Generally speaking, the reason seems to be that `Helper::processDBasAdmin()` requires various bootstrapping that may not have happened in the context (install tool), which is why there is some manual bootstrapping. Since just removing the version guard would have resulted in another error, and doing manual bootstrapping isn't all that clean anyways, in this PR I use query builders directly.
- I was wondering if there is a more TYPO3-canonical place for this code. For now, I moved it into an upgrade wizard.

---

One difference here is that `tstamp`, `crdate` and `cruser_id` are set to zero instead of their actual values. This didn't seem to be important to me, but of course I could add these fields to the insert statement if wanted.


dvoracek-slub:
Update: Rebased to latest `dev-extbase-fluid`.

Resources/Private/Data/FormatDefaults.php
```diff
@@ -0,0 +1,44 @@
+<?php
```
albig, line 1: This is a good location to put the format definitions.

Classes/Updates/AddDefaultFormats.php
```diff
@@ -0,0 +1,204 @@
+<?php
```
albig, line 1: I struggle about the location in the update wizard folder. An update wizard is usually used on updating an existent Installation. So Kitodo.Presentation in an older version is already on place and the `tx_dlf_formats` records, too.

Please check if it won't be possible to merge this functionality into the NewTenant module and even remove it from the ConfigurationForm hook at all. Creating records on calling the configuration form is an unexpected behavior.

Why? In the past the `tx_dlf_format` records on the root level always caused trouble if you export a tree and move to another system e.g.. All beginners will fail on this. And the experts hate it.

So why not store the `tx_dlf_format` records in the Kitodo.Presentation folder, too? Of course you get multiple `tx_dlf_format` records for always the same formats, if you use multiple tenants on different folders. But is this really a problem?

Even, if we keep the location on uid=0, we could create these records in the NewTenant module. The module could check, if `tx_dlf_format` records are present. If not, it will create them _before_ creating the `tx_dlf_metadata` records.

albig (changes requested):
I confirm, that this is a bug in 10.4. Please have a look at my comments / questions. Thank you.

Classes/Updates/AddDefaultFormats.php
dvoracek-slub: > I struggle about the location in the update wizard folder. [...]

Yes, I also don't find this ideal but didn't know a better place. I didn't put it into the NewTenant module because that is more geared towards creating records in the data folders, but I don't mind moving it there if you think it's better. (And I still wonder if TYPO3 really has no standard way for seeding the database, except for static tables.)

> Please check if it won't be possible to merge this functionality into the NewTenant module and even remove it from the ConfigurationForm hook at all. Creating records on calling the configuration form is an unexpected behavior.

Absolutely, and in my PR the entries aren't created in the configuration form. There's only a message that tells whether or not the records exist, and asks the user to use the upgrade wizard if they don't. (That's also meant for people who know the old behavior.)

> So why not store the tx_dlf_format records in the Kitodo.Presentation folder, too? Of course you get multiple tx_dlf_format records for always the same formats, if you use multiple tenants on different folders. But is this really a problem?

I don't see a problem in general, though we would probably need an upgrade wizard that copies ``tx_dlf_format`` records from the root page to each data folder and adjust ``tx_dlf_metadataformat``? So we can do it, it just seems like a more intrusive change.

albig: > Yes, I also don't find this ideal but didn't know a better place. I didn't put it into the NewTenant module because that is more geared towards creating records in the data folders, but I don't mind moving it there if you think it's better. (And I still wonder if TYPO3 really has no standard way for seeding the database, except for static tables.)

I only know the ancient `ext_tables_static+adt.sql` which we shouldn't use anymore and the [Distribution Package](https://docs.typo3.org/m/typo3/reference-coreapi/main/en-us/ExtensionArchitecture/CreateNewDistribution/Index.html) what is not applicable, either. But in the last, we could maybe have a look to get some good ideas?

> Absolutely, and in my PR the entries aren't created in the configuration form. There's only a message that tells whether or not the records exist, and asks the user to use the upgrade wizard if they don't. (That's also meant for people who know the old behavior.)

Oh, sorry. I didn't see this. This is of course better.

> I don't see a problem in general, though we would probably need an upgrade wizard that copies `tx_dlf_format` records from the root page to each data folder and adjust `tx_dlf_metadataformat`? So we can do it, it just seems like a more intrusive change.

It's for sure the more intrusive change. But for the 4.0 release we could so changes like this. The question is, would it work easily or not.

Yes, we need an update wizard for this which might cause some work. In former times, there has been an update job to move the solr-records to the tenant folder, too.

albig:
Close in favor of #777

PR #763, opened 2022-02-10
Upgrade OpenLayers to 6.12.0

dvoracek-slub:
This PR upgrades OpenLayers to the latest version (6.12.0). The base branch is `master` to make this easier to test for now, though eventually it's supposed to end up in the Extbase/Fluid branch.

My approach here is to do the upgrade incrementally – following along the upgrade notes, there is a separate commit for each version of OpenLayers (except for patch releases). Until v6, I also migrated the patches from the OL3 fork (original fork is https://www.github.com/slub/ol3, my repo is https://www.github.com/dvoracek-slub/ol3); this isn't necessary anymore since v6.

In comparison to @sebastian-meyer's PR (https://github.com/kitodo/kitodo-presentation/pull/584), which also does refactorization and other improvements, here I'm more focused on doing just the upgrade. Still, I took the other PR as a basis for some changes such as the replaced image filter (CSS instead of WebGL).

There's a ~pageview.md~ PageView.rst where I list the features (and how to test them) plus some sample URLs for all supported image formats.


Resources/Private/Docs/pageview.md
```diff
@@ -0,0 +1,108 @@
+# Page View
```
albig: This is a great work over all. 

Two remarks in this state:

1. We should not open another documentation path. Please move the pageview documentation to Documentation and use ReST as format.
2. Please rebase to the current master

I'm not yet sure how to bring together dev-extbase-fluid with this PR. But we will see.

dvoracek-slub: Thanks :-) I transcribed the documentation to .rst and rebased to master.

Documentation/Features/PageView.rst
```diff
@@ -0,0 +1,140 @@
...
+   plugin.tx_dlf_pageview {
+       features = OverviewMap,ZoomPanel
+   }
+
+These are created in ``dlfViewer::createControls_()``.
+
+Experimental Tools
+------------------
+
+Magnifier
```
albig, line 88: "Magnifier" and "Cropping Tool" are only used together with the basket feature or are they useful for something else?

Resources/Public/Javascript/OpenLayers/get-ol.sh
```diff
@@ -0,0 +1,8 @@
+#!/bin/bash
+
+VERSION="6.12.0"
+
+wget -O ol.zip "https://github.com/openlayers/openlayers/releases/download/v$VERSION/v$VERSION-dist.zip"
+unzip -p ol.zip "v$VERSION-dist/ol.js" > openlayers.js
+unzip -p ol.zip "v$VERSION-dist/ol.css" > openlayers.css
```
albig, line 7: Before, we made a custom build of the Openlayers library. Isn't this still useful or is this not possible anymore?
Now the full openlayers.js is about 1,3 MB, the custom build before 190 kB.

The transfered data to the browser in the workview is only about 1,7 MB (including image, libraries, HTML) vs. 1,5 MB before. This is due to compression on the transfer, I think. So it's usable. But of course, if we could reduce the size here, it could be worth maybe.

albig (commented):
The OpenLayers upgrade works for me in the workview on current master. There seems to be no functional difference for me at a first view.

I placed two comments below. Please have a look at it.

Documentation/Features/PageView.rst
dvoracek-slub: Oh. Stupidly, I didn't even realize that these are related to the basket and treated them in isolation.

Resources/Public/Javascript/OpenLayers/get-ol.sh
dvoracek-slub: I see, I had a remark on this but it's buried in the commit notes.

So I think it would still be useful to reduce the bundle size considering that parsing the decompressed JS also takes time. My idea was to postpone this until we start using a bundler (something like Webpack) for our JS.

As far as I see, OL doesn't anymore have a build tool as before where features could be selected in a configuration file. If you think it's urgent, it would probably be possible to do the build manually by writing a JS file that exports the necessary components. (That is, we could adopt the way how they build the full JS distribution.)

The approach in OL6 seems to be to use ES6 import together with a bundler. That is, we wouldn't rely on a global openlayers.js anymore, but only import what we need in the application. Then Webpack, via "tree-shaking", would only include the code that is actually used in the final bundle. That way we also wouldn't need to know and list what we use in a separate file.

Documentation/Features/PageView.rst
albig: That's nowhere written. I reverse engineered this some time ago from the initial PR of the basket feature. In the Extbase branch, I changed the setting of the flexform to `settings.basket.magnifier` to make it clear for the future.

Resources/Public/Javascript/OpenLayers/get-ol.sh
albig: Ok, thank you. I haven't read your remark about this.

Yes, you can postpone this point. I only wanted to be sure, you kept it in mind. 

Actually, the browser loads the JavaScript only once and loads it from cache afterwards. Of course the parsing time has to be done on every load (as long the browser doesn't cache it either). So I could live with this and maybe we get better later on with webpack or something similar.

albig (dismissed):
Is there anything you still want to do on this issue? Otherwise, I would say: let's got master and merge it to dev-extbase-fluid afterwards.

dvoracek-slub:
Thanks for your remarks. I don't have any functional changes left. I only though about squashing all or some of the commits to make commit history a little cleaner. For instance, from 6.4 onwards it's only upgrading OL without changes in our code.

For dev-extbase-fluid, I'll look into updating the remarks on magnifier/cropping.

dvoracek-slub:
So I squashed some of the commits and think this could go into master.

albig:
> So I squashed some of the commits and think this could go into master.

There is one complain by Codacy. Could you have a look on it? https://app.codacy.com/gh/kitodo/kitodo-presentation/pullRequest?prid=8867229

albig (approved):

PR #768, opened 2022-02-15
dev-extbase-fluid: Small fixes

dvoracek-slub:
This fixes a Codacy warning and two typehint-related issues.

- The signatures of `SolrCore::getPid/setPid` aren't compatible with those in the base class `AbstractDomainObject`, and there's no need to reimplement them anyways.
- By using the `Lazy` annotation on `Document::$owner`, it may end up as an instance of `LazyLoadingProxy`. This is converted to `Library` in `getOwner()`. (No fix is needed for `$collections`, because it's instance of `LazyObjectStorage`, which is a subclass of `ObjectStorage`.)
  - See https://forge.typo3.org/issues/60460 on the underlying issue


Classes/Common/Helper.php
```diff
@@ -199,6 +200,19 @@ public static function decrypt($encrypted)
         return $decrypted;
     }
 
+    /**
+     * Convert a domain object that is lazy-loaded to the actual instance, i.e.,
+     * unwrap LazyLoadingProxy if necessary.
+     *
+     * @access public
+     */
+    public static function fromLazy($object)
```
albig: I don't like to move this into the Helper class as long we do not use it multiple times in the code. If it's only the `Document::getOwner()`, please move it there.

Classes/Domain/Model/SolrCore.php
```diff
@@ -29,22 +29,6 @@ class SolrCore extends \TYPO3\CMS\Extbase\DomainObject\AbstractEntity
      */
     protected $indexName;
 
-    /**
```
albig, line 32: That's totally true. `@return string` is even wrong.

albig (changes requested):
Please have look at my comment.

albig (approved):
Looks good to me. Couldn't test it, because I destroyed my DDEV environment for testing #761 :-(

PR #770, opened 2022-02-17
dev-extbase-fluid: Some fixes to OAI-PMH interface

dvoracek-slub:
This fixes some issues that I noticed while writing tests (#756).

---

- Fix listing records and identifiers (set $resultSet['elements'])
- Fix XML error when embedding METS documents (there has been an excess angle bracket)
- Don't include <resumptionToken> tag when there's no token
- Fix value of cursor attribute in <resumptionToken>
- Don't show request parameters when there's a badVerb or badArgument
- Restore fallback "earliestDatestamp" when no document is found

### DIFF
--- a/Classes/Controller/OaiPmhController.php
+++ b/Classes/Controller/OaiPmhController.php
@@ -240,10 +240,7 @@ class OaiPmhController extends AbstractController
             // Get root element.
             $root = $xml->getElementsByTagNameNS($this->formats['mets']['namespace'], 'mets');
             if ($root->item(0) instanceof \DOMNode) {
-                // Import node into \DOMDocument.
-                $mets = $xml->saveXML();
-                // Remove leading line
-                $mets = substr($mets, strpos($mets, '>'));
+                $mets = $xml->saveXML($root->item(0));
             } else {
                 $this->logger->error('No METS part found in document with location "' . $record['location'] . '"');
             }
@@ -284,6 +281,9 @@ class OaiPmhController extends AbstractController
                 break;
             case 'ListSets':
                 $this->verbListSets();
+                break;
+            default:
+                $this->error = 'badVerb';
                 break;
         }
 
@@ -404,6 +404,9 @@ class OaiPmhController extends AbstractController
         if ($document) {
             $oaiIdentifyInfo['earliestDatestamp'] = gmdate('Y-m-d\TH:i:s\Z', $document->getTstamp()->getTimestamp());
         } else {
+            // Provide a fallback timestamp if no document is found
+            $oaiIdentifyInfo['earliestDatestamp'] = '0000-00-00T00:00:00Z';
+
             // access storagePid from TypoScript
             $pageSettings = $this->configurationManager->getConfiguration($this->configurationManager::CONFIGURATION_TYPE_FULL_TYPOSCRIPT);
             $storagePid = $pageSettings["plugin."]["tx_dlf."]["persistence."]["storagePid"];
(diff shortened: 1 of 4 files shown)


Classes/Controller/OaiPmhController.php
```diff
@@ -240,10 +240,7 @@ protected function getMetsData(array $record)
             // Get root element.
             $root = $xml->getElementsByTagNameNS($this->formats['mets']['namespace'], 'mets');
             if ($root->item(0) instanceof \DOMNode) {
-                // Import node into \DOMDocument.
-                $mets = $xml->saveXML();
-                // Remove leading line
-                $mets = substr($mets, strpos($mets, '>'));
+                $mets = $xml->saveXML($root->item(0));
```
albig, line 243: so easy. Thank you :-)

Classes/Controller/OaiPmhController.php
```diff
@@ -764,7 +776,7 @@ protected function generateResumptionTokenForDocumentListSet(array $documentList
 
         $resumptionTokenInfo = [];
         $resumptionTokenInfo['token'] = $resumptionToken;
-        $resumptionTokenInfo['cursor'] = $documentListSet['metadata']['completeListSize'] - count($documentListSet);
+        $resumptionTokenInfo['cursor'] = $currentCursor;
```
albig, line 779: This is correct for the `cursor` attribute following the linked docu. But we should change the XSLT file for the HTML-Output for this, don't we?

Kitodo Presentation  OAI 2 0 Request Results

Before your change, there was written "10" (for 10 documents per request).

Resources/Private/Templates/OaiPmh/Main.xml
```diff
@@ -122,7 +124,9 @@
                         </f:else>
                     </f:if>
                 </f:for>
-                <resumptionToken cursor="{resumptionToken.cursor}" completeListSize="{resumptionToken.completeListSize}" expirationDate="<f:format.date format="Y-m-d\TH:i:s\Z">{resumptionToken.expired}</f:format.date>">{resumptionToken.token}</resumptionToken>
+                <f:if condition="{resumptionToken.token}">
```
albig: The `resumptionToken` is missing now on the last "page".

In the linked "FlowControl" it's written:

> The repository responds with an incomplete list of 75 records. The repository marks this list as the final incomplete list by including in the response an empty resumptionToken element with two attributes: a completeListSize of 175, and a cursor of 100.

albig (changes requested):
Thank you for working on that. Please have a look at my comments.

Classes/Controller/OaiPmhController.php
dvoracek-slub: :-)

Resources/Private/Templates/OaiPmh/Main.xml
dvoracek-slub: Sorry for having missed that. Do you think the patch is right? As far as I understood, the `<resumptionToken>` tag should still be omitted if there is only one page, and there shouldn't be an `expirationDate` parameter on the last page. So I've extracted this to a separate section in the Fluid template.

Classes/Controller/OaiPmhController.php
dvoracek-slub: I'm not sure what's expected here, but if it's just a more user-friendly representation of the XML file, perhaps we could just clarify that it's the "*Previously* Submitted Records"?

albig: Unfortunately, we do not have the amount of items in the current list. I've just looked around, how other solved this.

albig (approved):

PR #771, opened 2022-02-17
Documentation

dvoracek-slub:
- For previewing documentation locally, I add some composer commands to make using the official `t3docs/render-documentation` a bit easier. Also, since that tool is somewhat slow, I added a custom Sphinx config that allows to spawn a server with auto-refresh. There's some instructions in DEVELOPMENT.md.
- For database documentation, I propose a CLI command that collects information from all sources I could think of (doc-comments on Extbase models; TCA labels; COMMENTs in SQL) and generates an .rst page that is meant as a table reference.
- There's now more explanations on the database fields, mostly in doc-comments.


dvoracek-slub:
@albig Any opinions on this? If you think it's useful, I could try to merge or rebase this onto your PR #779.

albig:
> @albig Any opinions on this? If you think it's useful, I could try to merge or rebase this onto your PR #779.

Not yet. Haven't looked at it. But of course, we have some conflicts now :-(

Build/Documentation/sphinx.sh
```diff
@@ -0,0 +1,55 @@
+#!/bin/bash
```
albig, line 1: This worked for me and is very useful.

Classes/Command/DatabaseDocCommand.php
```diff
@@ -0,0 +1,292 @@
+<?php
```
albig: This didn't work.

```
Uncaught TYPO3 Exception Class 'TYPO3\CMS\Extbase\Reflection\DocCommentParser' not found
thrown in file /var/www/extensions/dlf/Classes/Command/DatabaseDocCommand.php
in line 188
```

I don't like this approach anyway. The TYPO3 scheduler / CLI task should never write into `typo3conf/*`. During development there might be exceptions like we use lfeditor to edit the XLIFF files. But lfeditor is in this case an external tool. This CLI job is part of Kitodo.Presentation and is equal to all other CLI jobs. Every user could add a Scheduler task to render the Documents.rst file in a read-only directory.

Hope you get the point. All this should go to `Build/`.

Classes/Common/DbDocs/RstSection.php
```diff
@@ -0,0 +1,131 @@
+<?php
+
```
albig: Copyright and PHPdocs are missing in this file.

albig (changes requested):
great approach, but not happy about parts of the concept.

Classes/Command/DatabaseDocCommand.php
dvoracek-slub: Oh, apparently the `DocCommentParser` has been removed in TYPO3 v10. For now I integrated it into the command if you would like to try it out.

If it's just about the command being run by other users, we could check for `TYPO3_CONTEXT=Development`.

But yes, it would be nicer if this could go into `Build/` and be called directly from the extension. The issue was that I didn't know how to do this, because I'm relying on TYPO3, e.g. for getting the classmap. I'll think about it.

dvoracek-slub:
I updated the branch (rebase) to incorporate your suggestions and #779, and wrote a little more on metadata. Would you have a look?

For the CLI part, I couldn't find a way to extract the information without at least pretending that TYPO3 is installed. That's why I added a `public/` folder and some minimal configuration to `Build/Documentation/dbdocs`.

albig (approved):

PR #772, opened 2022-02-21
Add 3D objects list to Table of content plugin

beatrycze-volk:


beatrycze-volk:
Close in favor to #803

PR #773, opened 2022-02-22
Update 3D View plugin

beatrycze-volk:
Depends on #686 and #772


albig:
This is on hold until the Extbase branch is merged. I doubt that this will be mergeable afterwards. So keep in mind, that you will have some work afterwards. Or try to switch to the future code-base already now.

PR #777, opened 2022-03-07
Add the default formats by newTenant module

albig:
Creating all necessary records to make Kitodo.Presentation work, is done now in the "New Tenant" module.

The module creates the `tx_dlf_formats` records, too. The location is still globally on PID=0. This is done only once for every tenant, if used multiple folders.

This PR is a replacement for #761 and fixes some struggles with the flash messages on the module output, too. Now everything is done in Fluid.


Classes/Domain/Model/Metadata.php
```diff
@@ -30,7 +34,9 @@ class Metadata extends \TYPO3\CMS\Extbase\DomainObject\AbstractEntity
     protected $indexName;
 
     /**
-     * @var int
+     * @var \TYPO3\CMS\Extbase\Persistence\ObjectStorage<\Kitodo\Dlf\Domain\Model\MetadataFormat>
```
albig, line 37: This IRRE-relation was not correct. I'm not fully sure about the lazy-loading. Please @dvoracek-slub, have a look on it, if we need the LazyLoadingProxy here.

Classes/Controller/NewTenantController.php
```diff
@@ -155,29 +227,22 @@ public function addMetadataAction()
...
-                $this->getLanguageService()->getLL('flash.solrcoreNotAddedMsg'),
-                $this->getLanguageService()->getLL('flash.solrcoreNotAdded'),
-                \TYPO3\CMS\Core\Messaging\FlashMessage::ERROR
-            );
+        $doPersist = false;
+
+        if ($this->solrCoreRepository->findOneByPid($this->pid) === null) {
+            $newRecord = GeneralUtility::makeInstance(SolrCore::class);
+            $newRecord->setLabel($this->getLanguageService()->getLL('flexform.solrcore') . ' (PID ' . $this->pid . ')');
+            $newRecord->setIndexName('');
```
dvoracek-slub: This didn't work for me. The database record was created, but no Solr core because now the DataHandler hook isn't called. Perhaps we could just create the core here and pass its name to `setIndexName()`?

Classes/Domain/Model/Metadata.php
dvoracek-slub: IMO, this should be fine. The LazyLoadingProxy is only used for single objects (`_:1` relations); for collections, their `LazyObjectStorage` inherits from `ObjectStorage`, so a conversion isn't necessary.

Configuration/TCA/tx_dlf_metadataformat.php
```diff
@@ -25,9 +25,6 @@
         'searchFields' => 'encoded',
         'hideTable' => 1,
     ],
-    'feInterface' => [
-        'fe_admin_fieldList' => '',
-    ],
     'interface' => [
     ],
     'columns' => [
```
dvoracek-slub, line 30: When trying to add a metadata format, I got an error that l18n_diffsource does not have a default value. Should we add the `ctrl.transOrigDiffSourceField` and `columns.l18n_diffsource` fields like for the other tables?

Resources/Private/Data/MetadataDefaults.php
```diff
@@ -35,17 +35,18 @@
...
             [
-                'encoded' => 2,
+                'format_root' => 'teiHeader',
                 'xpath' => './teihdr:fileDesc/teihdr:sourceDesc/teihdr:msDesc/teihdr:head/teihdr:note[@type="caption"]',
                 'xpath_sorting' => '',
             ],
             [
-                'encoded' => 5,
+                'format_root' => 'IIIF2',
+                'format_root' => 'IIIF1',
```
dvoracek-slub: The second line should probably be removed.

Resources/Private/Language/de.locallang.xlf
```diff
@@ -461,6 +469,71 @@
...
+                <target><![CDATA[Die Standard-Namensräume für MODS, TEIHDR und/oder ALTO wurden erfolgreich definiert.]]></target>
+            </trans-unit>
+            <trans-unit id="newTenant.formatNotCreated" approved="yes">
+                <source><![CDATA[Default namespaces not found!]]></source>
+                <target><![CDATA[Standard-Namensräume nicht gefunden!]]></target>
+            </trans-unit>
+            <trans-unit id="newTenant.formatNotCreatedMsg" approved="yes">
+                <source><![CDATA[The namespace definitions for MODS, TEIHDR and/or ALTO do not exist and could not be created.]]></source>
+                <target><![CDATA[Die Standard-Namensräume für MODS, TEIHDR und/oder ALTO konnten nicht definiert werden.]]></target>
+            </trans-unit>
```
dvoracek-slub, line 495: These messages don't seem to be used.

dvoracek-slub (changes requested):

Classes/Domain/Model/Metadata.php
albig: Ok. Thank you for clarifying.

Classes/Controller/NewTenantController.php
albig: Good point. So my setup just recreated the deleted solr record. I'll think about it.

Configuration/TCA/tx_dlf_metadataformat.php
albig: How to you add an metadata format record? I add it as inline record inside the metadata record. I get no error.
Did you made a database compare? I added the language field for `tx_dlf_metadataformats` again, because I had similar problems before.

Resources/Private/Data/MetadataDefaults.php
albig: Thank you. That's correct.

Resources/Private/Language/de.locallang.xlf
albig: Correct. I removed all these flash messages. You land on the page with the New Tenant module and see, that nothing changed. I think, this is enough.

Configuration/TCA/tx_dlf_metadataformat.php
dvoracek-slub: Okay, the issue has been that my database still had the l18n_diffsource column even though it had been removed from ext_tables. Did you want to also bring back that column? Otherwise, the fix would be to remove the column via "Maintenance → Analyze Database Structure" after updating.

albig: No, this column is not needed and was (IMHO) introduced during development. And it's totally ok for a major release to require a full database compare.

Classes/Controller/NewTenantController.php
albig: I think, it's fixed now.
In the DataHandler we need this part, too because you can just create the Solr core record via the List module.

dvoracek-slub (approved):

PR #782, opened 2022-03-11
Cleanup to new tenant module

albig:
... and to metadata controller.

In my current setup, the "owner" is not rendered in metadata plugin. @dvoracek-slub do you have an idea, why?


dvoracek-slub:
Not sure. I instead got an error "Call to a member function getLabel() on null" because `libraryRepository->findOneByIndexName()` returns null. The owner in my case comes from MODS (`./mods:name[./mods:role/mods:roleTerm="own"]/mods:displayForm`), which rather corresponds to the library label. (Though presumably we cannot assume there is a record at all.)

Or did you mean the owner from the database? This reminds me that some code reading the owner has been removed in #750.

On a side note, I just noticed that when trying to create a library, I get the error that `image` doesn't have a default value. Probably the image column needs to be adjusted in `ext_tables.sql` as per https://docs.typo3.org/m/typo3/reference-coreapi/main/en-us/ApiOverview/Fal/UsingFal/Tca.html? (Which makes me wonder if that is somehow related to the `sys_file_reference` issues we had some while ago...)

albig:
> Not sure. I instead got an error "Call to a member function getLabel() on null" because `libraryRepository->findOneByIndexName()` returns null. The owner in my case comes from MODS (`./mods:name[./mods:role/mods:roleTerm="own"]/mods:displayForm`), which rather corresponds to the library label. (Though presumably we cannot assume there is a record at all.)

Could you please share this example with me? I've currently one examples without owner. In this case, the documents are associated to the default owner record.

The flow is (now) as follows:

1. On indexing the owner is found by xpath of the metadata record. If no "library" record exists, it will be created.
2. On pageview, the metadata is fetched again via xpath from the metadata record. If this is not found, the associated library-record of the document is shown.

Bevor step 2. was a little different. 

> Or did you mean the owner from the database? This reminds me that some code reading the owner has been removed in #750.

This was the behavior before and filled the `metadataArray['owner']` in MetadataController. Something we can drop due to the Extbase approach. I think the better way is to show the library record associated to the document.

(But still, if there is a working xpath, this will be overwritten).

> On a side note, I just noticed that when trying to create a library, I get the error that `image` doesn't have a default value. Probably the image column needs to be adjusted in `ext_tables.sql` as per https://docs.typo3.org/m/typo3/reference-coreapi/main/en-us/ApiOverview/Fal/UsingFal/Tca.html? (Which makes me wonder if that is somehow related to the `sys_file_reference` issues we had some while ago...)

That's correct. The database field for `image` and `thumbnails` should be the same and contain a default value. I fixed this now. Please do a database compare.

But... The translated record of the library is not found. This might be because of our lazy loading constrution of `document->library`. With translated type (structure) and collection, it's working. What do you think?

I'll have a look later, if we could even add the translated metadata and structures in the newTenant module. This would be really an improvement.

albig:
The new tenant module now adds the metadata and structure records in all configured siteLanguages. Usually it's German (default) -> English. But the other way round should work, too.

This points out another issue: The metadata wrap is not found of the default language. @dvoracek-slub can you confirm this behavior?

Digitale Sammlungen  Auf der Suche nach Zukunft  Das Beispiel Pieschen (6)

dvoracek-slub:
(Sorry, I misclicked while writing this and accidentally closed the PR.)

- Ad owner: Sorry, what I said was wrong as I had misinterpreted the result. The owner didn't come from MODS but from the `default_value` of the metadata record. So I don't have an example URL, but the result is the same when using a hand-written example.
- I also got an error when a collection or structure used in the METS file isn't found.
- Yes, it looks like the `getOwner()` doesn't return the translated record. Don't know why.
- Ad metadata wrap: Yes, I also have that behavior when using the newly created records. But how does Extbase handle translated records? Shouldn't all the non-translated fields be copied into the new record instead of setting wrap to an empty string in the translated record?
- My BE user has language set to English. When creating the records in NewTenant, I got the same (English) label in both records. It would use German and English labels only when I changed the BE user language.
- Two issues I had in 9.5:
  - `Could not find class definition for name "Metadata".`, and the same for structures. This seems to refer to the type annotations for `$l18nParent`. Apparently in v9 they didn't yet consider the current namespace, so we should use the fully qualified class name like `\Kitodo\Dlf\Domain\Model\Metadata`.
  - When opening NewTenant on the root page, there is an exception instead of simply "not a folder". Looks like a bug in `NullSite` when no language is given to constructor.
- Another issue was that when saving a record that has a translation in backend, it would append a zero to the fields marked as "uniqueInPid". This seems strange because https://forge.typo3.org/issues/83572 looks like this shouldn't happen since v8?

albig:
> * Yes, it looks like the `getOwner()` doesn't return the translated record. Don't know why.

Thanks. We should try to remove the LazyLoading.

> * Ad metadata wrap: Yes, I also have that behavior when using the newly created records. But how does Extbase handle translated records? Shouldn't all the non-translated fields be copied into the new record instead of setting wrap to an empty string in the translated record?

This is a TCA configuration. This should work with `l10n_mode=exclude` and take the field value of the default record. See https://docs.typo3.org/m/typo3/reference-tca/main/en-us/Columns/Properties/L10nMode.html

> * My BE user has language set to English. When creating the records in NewTenant, I got the same (English) label in both records. It would use German and English labels only when I changed the BE user language.

Puh, what a nightmare. You're right.  I played around with all this and currently, I only see the possibility to read the language files by hand somehow. Maybe you have another idea?

> * Two issues I had in 9.5:
>   
>   * `Could not find class definition for name "Metadata".`, and the same for structures. This seems to refer to the type annotations for `$l18nParent`. Apparently in v9 they didn't yet consider the current namespace, so we should use the fully qualified class name like `\Kitodo\Dlf\Domain\Model\Metadata`.

Seems to be like this. I added the full namespace.

>   * When opening NewTenant on the root page, there is an exception instead of simply "not a folder". Looks like a bug in `NullSite` when no language is given to constructor.

Possible. Havent' looked at this yet. 

> * Another issue was that when saving a record that has a translation in backend, it would append a zero to the fields marked as "uniqueInPid". This seems strange because https://forge.typo3.org/issues/83572 looks like this shouldn't happen since v8?

You mean the `index_name` field? But this is not new, isn't it?

PR #791, opened 2022-03-26
Add unique id to search form

albig:
* bring back lost CSS class
* add unique-id to form labels of search plugin

### DIFF
--- a/Classes/Controller/SearchController.php
+++ b/Classes/Controller/SearchController.php
@@ -176,7 +176,7 @@ class SearchController extends AbstractController
      */
     protected function addFacetsMenu()
     {
-        // Quit without doing anything if no facets are selected.
+        // Quit without doing anything if no facets are configured.
         if (empty($this->settings['facets']) && empty($this->settings['facetCollections'])) {
             return '';
         }
@@ -205,6 +205,7 @@ class SearchController extends AbstractController
         $menuArray = [];
         // Set default value for facet search.
         $search = [
+            'query' => '*:*',
             'params' => [
                 'component' => [
                     'facetset' => [
--- a/Classes/Controller/SearchController.php
+++ b/Classes/Controller/SearchController.php
@@ -153,16 +153,12 @@ class SearchController extends AbstractController
         // Get additional fields for extended search.
         $this->addExtendedSearch();
 
-        // Add the current document if present to fluid. This way, we can limit further searches to this document.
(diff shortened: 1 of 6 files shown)


Classes/Controller/SearchController.php
```diff
@@ -203,6 +205,7 @@ public function makeFacetsMenuArray($facets)
         $menuArray = [];
         // Set default value for facet search.
         $search = [
+            'query' => '*:*',
```
dvoracek-slub, line 208: When using "OR" in the extended search with no search query, this would result in a Solr query `*:* OR title_tsi:...`. So the extended search would basically be ignored. (Is this why in 0e23cb8 you removed it in the first place?)

Resources/Private/Templates/Search/Main.html
```diff
@@ -12,15 +12,15 @@
...
+<f:form action="search" controller="Search" name="searchParameter" method="get" class="tx-dlf-search-form">
+    <label for="tx-dlf-search-query-{viewData.uniqueId}">
         <f:if condition="{lastSearch.query}">
             <f:then>{lastSearch.query -> f:format.htmlspecialchars()}</f:then>
             <f:else><f:translate key="search.query"/></f:else>
         </f:if>
     </label>
 
-    <f:form.textfield id="tx-dlf-search-query" property="query" value="{lastSearch.query}" />
+    <f:form.textfield id="tx-dlf-search-query-{viewData.uniqueId}" property="query" value="{lastSearch.query}" />
```
dvoracek-slub: Wouldn't this break the search suggester, because it uses `$("#tx-dlf-search-query")`? (Also, there's some CSS in slub_digitalcollections and slub_web_sachsendigital that relies on the ID.)

albig: You're totally right. I reverted my change and brought back the former warning, not to change the id ;-)

Classes/Controller/SearchController.php
albig: You're right. This happens. The whole "extended search" is not well designed. E.g. it is unusable, if you make a full text search.
In our catalogue, the simple search is hidden, if you use the extended search. Maybe this should be done in Kitodo.Presentation, too. I reset any query in `makeFacetsMenuArray()` if using the extended search now. Hopefully this doesn't destroy anything else...

dvoracek-slub (approved):

PR #792, opened 2022-03-28
Simplify method for getting logical structure info from XML metadata

beatrycze-volk:


albig (commented):
Is this simpler? Is the getLogicalStructureDetails used anywhere else?

beatrycze-volk:
I had such an episode of usage it for 3DViewer but I think at the end I have solved it differently. This change is left from that times.

PR #795, opened 2022-03-31
PageView Performance

dvoracek-slub:
This PR improves PageView performance and fixes two small issues in word highlighting.

- Highlighting: Also highlight words on second page in double-page mode, and clear highlight on subsequent searches.
- Don't load static images twice (with the exception of non-CORS or mixed content)
- Don't load and parse fulltext files twice (previously both in FulltextConrol and in FulltextDownloadControl)
- Replace OL's `forEachFeatureAtPixel`, which can be very slow on some systems, with manual bounds checking
- Some more optimizations to fulltext loading (load async/non-blocking) and rendering

Some ways to test the PR:
- Check that the image is loaded correctly, also from HTTP and non-CORS URLs.
- In DevTools, check the "Finish" times when loading a document, for example:
  - https://ddev-kitodo-presentation.ddev.site/en/workview?tx_dlf[id]=https%3A%2F%2Fdigital.slub-dresden.de%2Fdata%2Fkitodo%2Faufdesun_351357262%2Faufdesun_351357262_mets.xml&tx_dlf[page]=4
  - https://ddev-kitodo-presentation.ddev.site/workview?tx_dlf[double]=0&tx_dlf[id]=https%3A%2F%2Fapi.deutsche-digitale-bibliothek.de%2Fitems%2F4UMJQFKXM2E2WZEAG6H5BH6JVHW4ZRW6%2Fsource%2Frecord
- In case you have a system where hovering over fulltext is very laggy, check if this PR improves it.

### DIFF
--- a/Documentation/Features/PageView.rst
+++ b/Documentation/Features/PageView.rst
@@ -22,6 +22,7 @@ The viewer supports various image formats. Each of them should work in single pa
    *  https://digital.slub-dresden.de/data/kitodo/aufdesun_351357262/aufdesun_351357262_mets.xml
    *  `Single page view <https://ddev-kitodo-presentation.ddev.site/workview?tx_dlf[id]=https%3A%2F%2Fdigital.slub-dresden.de%2Fdata%2Fkitodo%2Faufdesun_351357262%2Faufdesun_351357262_mets.xml&tx_dlf[page]=4>`__
    *  `Single page view with word highlighting <https://ddev-kitodo-presentation.ddev.site/workview?tx_dlf[id]=https%3A%2F%2Fdigital.slub-dresden.de%2Fdata%2Fkitodo%2Faufdesun_351357262%2Faufdesun_351357262_mets.xml&tx_dlf[page]=4&tx_dlf[highlight_word]=Dresden>`__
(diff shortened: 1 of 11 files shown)


albig (approved):
This seems to work pretty good.

One question. What does this "blob:https://.." mean in Chromium network monitor. Haven't seen this so far.

### current master with double loading
kitodo-pr795-master-2


### your PR with performance enhancement
kitodo-pr795-2

dvoracek-slub:
> One question. What does this "blob:https://.." mean in Chromium network monitor. Haven't seen this so far.

Generally, what I'm doing in `fetchStaticImageData` is to load the image into memory (into a [Blob](https://developer.mozilla.org/en-US/docs/Web/API/Blob)), then [create an object URL](https://developer.mozilla.org/en-US/docs/Web/API/URL/createObjectURL) that can be accessed like any network URL but really only references the object in memory. So OpenLayers can treat this as though I had passed the image URL directly. The DevTools entry indicates that OL is now accessing the image via the object URL.

albig:
Fancy stuff. Thank you for the explanation (and it's in the code, too - perfect).

PR #797, opened 2022-04-05
Small cleanups

albig:
* remove unused flexform settings (maybe there are more)
* remove unusable Hook --> use Fluid for your needs
* add CSS class
* remove unused TypoScript for TableOfContents


dvoracek-slub (approved):
Can't really tell about some changes (metadata hook, expAll setting), but generally looks good.

PR #801, opened 2022-04-12
Remove template variables pointing to not existing files

beatrycze-volk:


albig (approved):
Thank you. For sure not the last "dead code"...

PR #802, opened 2022-04-13
[BUGFIX] Some bug fixes and maintenance tasks

albig:
* fix listing of unstored metadata records in `ListView`
* remove unused code from `DocumentRepository`
* fix a bug for external files in `TableOfContents` (the TableOfContents plugin needs some more patient but this is not new to 4.0)
* fix adding of new collections on indexing (how could we miss this?)
* fix statistics count for volumes and in case it is limited to certain collections

### DIFF
--- a/Classes/Controller/TableOfContentsController.php
+++ b/Classes/Controller/TableOfContentsController.php
@@ -176,7 +176,7 @@ class TableOfContentsController extends AbstractController
         // Does the document have physical elements or is it an external file?
         if (
             !empty($this->document->getDoc()->physicalStructure)
-            || !MathUtility::canBeInterpretedAsInteger($this->document->getDoc()->uid)
+            || !MathUtility::canBeInterpretedAsInteger($this->requestData['id'])
         ) {
             // Get all logical units the current page or track is a part of.
             if (
--- a/Classes/Domain/Repository/DocumentRepository.php
+++ b/Classes/Domain/Repository/DocumentRepository.php
@@ -90,17 +90,6 @@ class DocumentRepository extends \TYPO3\CMS\Extbase\Persistence\Repository
         return $document;
     }
 
-
-    public function findByUidAndPartOf($uid, $partOf)
-    {
-        $query = $this->createQuery();
-
-        $query->matching($query->equals('uid', $uid));
-        $query->matching($query->equals('partof', $partOf));
-
-        return $query->execute();
-    }
-
     /**
      * Find the oldest document
      *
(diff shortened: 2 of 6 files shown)


Classes/Controller/TableOfContentsController.php
```diff
@@ -176,7 +176,7 @@ public function makeMenuArray()
         // Does the document have physical elements or is it an external file?
         if (
             !empty($this->document->getDoc()->physicalStructure)
-            || !MathUtility::canBeInterpretedAsInteger($this->document->getDoc()->uid)
+            || !MathUtility::canBeInterpretedAsInteger($this->requestData['id'])
```
dvoracek-slub, line 179: I was just wondering what the intended behavior is when a URL is given to a document that is stored in database. In that case (AFAICS), `$this->document` would still point to the actual database record, because in AbstractController it's matched by the recordId.

dvoracek-slub (dismissed):

Classes/Domain/Repository/DocumentRepository.php
```diff
@@ -215,177 +205,151 @@ public function findAllByCollectionsLimited($collections, $limit = 50)
...
+     *  b) "root" elements that are not referenced by other documents ("root" elements that have no descendants)
 
-    /**
-     * Count the volumes
-     *
-     * documents with partof != 0
-     *
      * @param array $settings
      *
      * @return int
```
beatrycze-volk: Everything fine, except applying Scrutinizer fixes.

beatrycze-volk (dismissed):

Classes/Controller/TableOfContentsController.php
beatrycze-volk: Interesting question.

I would assume that XML file content has in that case higher priority than this what is stored in the database.

Is it also not specified in some other place (creating document instance?) that if URL points into the database object it transforms from URL to numerical ID?

albig: That's a good question and we should find a solution for the TOC anyway. But this is buggy in 3.x (and even 2.x) I think.

Example:
1. call Börsenblatt anchor: http://digital.slub-dresden.de/id39946221X
2. in TOC: all years are listed with IDs --> expected behaviour
3. click "1834" --> all issues have IDs but the backlink to the anchor has an URL now
4. click the anchor link --> you are back in 1.) with the difference, that
   * the link to itself is via URL
   * you cannot search in the anchor (full title) because no ID is given 
   
Things get worse, if you have only e.g. a part of the years indexed locally. In this case in step 2.) all years are listed because they are part of the METS file and they are linked via URL.

We cannot fix this in this PR and should think about in a new issue.

Classes/Domain/Repository/DocumentRepository.php
albig: Right. Thank you.

beatrycze-volk (approved):

PR #803, opened 2022-04-20
[FEATURE] Add 3D objects list with filtering to Table of Contents plugin 

beatrycze-volk:
PR contains displaying list of 3D models stored inside the XML file. 3D models consist of title, image and description. There is also included a placeholder for filtering of the models (filtering itself will be implemented first when requirements are defined).


Classes/Controller/AbstractController.php
```diff
@@ -141,6 +141,11 @@ protected function loadDocument($requestData)
                         $this->document = GeneralUtility::makeInstance(Document::class);
                     }
 
+                    //it looks that for not loaded from database it is always 0, so needs to be set up on load
```
```suggestion
                    // Make sure configuration PID is set when applicable
```

Classes/Controller/TableOfContentsController.php
```diff
@@ -33,6 +33,167 @@ class TableOfContentsController extends AbstractController
...
+        // $this->requestData['page'] may be integer or string (physical structure @ID)
+        if (
+            (int) $this->requestData['page'] > 0
+            || empty($this->requestData['page'])
+        ) {
+            $this->requestData['page'] = MathUtility::forceIntegerInRange((int) $this->requestData['page'], 1, $this->document->getDoc()->numPages, 1);
+        } else {
+            $this->requestData['page'] = array_search($this->requestData['page'], $this->document->getDoc()->physicalStructure);
+        }
+        $this->requestData['double'] = MathUtility::forceIntegerInRange($this->requestData['double'], 0, 1, 0);
```
sebastian-meyer: Is this really necessary? When dealing with an 3D object list there are no physical pages...

Resources/Private/Partials/TableOfContents/Filter.html
```diff
@@ -0,0 +1,17 @@
...
+ * This file is part of the Kitodo and TYPO3 projects.
+ *
+ * @license GNU General Public License version 3 or later.
+ * For the full copyright and license information, please read the
+ * LICENSE.txt file that was distributed with this source code.
+-->
+</f:comment>
+<f:form action="filter" controller="TableOfContents" name="filterParameter" method="get" class="tx-dlf-filter-form">
+    <label for="tx-dlf-filter-title">
+        Title
```
sebastian-meyer: Please make this translatable.

sebastian-meyer (changes requested):

Classes/Controller/TableOfContentsController.php
beatrycze-volk: Removed

Resources/Private/Partials/TableOfContents/Filter.html
beatrycze-volk: Replaced with translation key

Classes/Format/Mods.php
```diff
@@ -129,5 +129,12 @@ public function extractMetadata(\SimpleXMLElement $xml, array &$metadata)
                 }
             }
         }
+        // Get "description"
+        $descriptions = $xml->xpath('./mods:recordInfo/mods:recordInfoNote/text()');
+        if (!empty($descriptions)) {
+            foreach ($descriptions as $description) {
+                $metadata['description'][0] = (string) $description;
+            }
+        }
```
sebastian-meyer: We should not hard-code this here, but instead add this to `https://github.com/kitodo/kitodo-presentation/blob/master/Resources/Private/Data/MetadataDefaults.php` as a new default configuration.
```suggestion
```

sebastian-meyer: The default entry in `https://github.com/kitodo/kitodo-presentation/blob/master/Resources/Private/Data/MetadataDefaults.php` would be like:

```
    'description' => [
        'format' => [
            [
                'format_root' => 'mods',
                'xpath' => './mods:recordInfo/mods:recordInfoNote',
                'xpath_sorting' => '',
            ],
        ],
        'default_value' => '',
        'wrap' => '',
        'index_tokenized' => 0,
        'index_stored' => 0,
        'index_indexed' => 0,
        'index_boost' => 0.00,
        'is_sortable' => 0,
        'is_facet' => 0,
        'is_listed' => 0,
        'index_autocomplete' => 0,
    ]
```

Also the translation strings for `description` have to be added to `https://github.com/kitodo/kitodo-presentation/blob/master/Resources/Private/Language/locallang_metadata.xlf` and `https://github.com/kitodo/kitodo-presentation/blob/master/Resources/Private/Language/de.locallang_metadata.xlf`.

Classes/Controller/TableOfContentsController.php
```diff
@@ -34,6 +34,156 @@ class TableOfContentsController extends AbstractController
...
+     * @param string $content: The PlugIn content
+     * @param array $conf: The PlugIn configuration
+     *
+     * @return array HMENU array
+     */
+    protected function makeMenuFor3DObjects()
+    {
+        $menuArray = [];
+        // Is the document an external file?
+        if (!MathUtility::canBeInterpretedAsInteger($this->requestData['id'])) {
```
sebastian-meyer: Why do you check for external files here? Isn't this supposed to work for indexed files as well?

sebastian-meyer (changes requested):
Please see my remarks below:

Classes/Controller/TableOfContentsController.php
beatrycze-volk: Currently it works only with external files. I don't think that currently indexed files would contain all needed information to display.

Classes/Format/Mods.php
beatrycze-volk: Ready

sebastian-meyer: Sorry for being so picky, but the translations go to `locallang_metadata.xlf` and `de.locallang_metadata.xlf`. You added them to the main translation files instead.

Classes/Controller/TableOfContentsController.php
sebastian-meyer: What's the difference here between an external file and an indexed file? Since the METS files don't get changed when indexing, this should make no difference. Or am I missing something?

Classes/Format/Mods.php
beatrycze-volk: Changed

Classes/Controller/TableOfContentsController.php
beatrycze-volk: I'm not sure. Te whole METS gets indexed or only defined fields?

sebastian-meyer: Only the configured fields get extracted from the METS and inserted into the Solr index and/or database. Thus, only those fields can be searched or displayed in the list view.

But as soon as you are working within a single document (i. e. page view, metadata view, table of contents, etc.) the original METS file is loaded anyway - regardless if the document is indexed or external.

Classes/Format/Mods.php
sebastian-meyer: Did you forget to push your changes? This branch seems unchanged...

Classes/Controller/TableOfContentsController.php
```diff
@@ -156,103 +304,56 @@ protected function resolveMenuEntry($entry)
...
+        $entryArray['title'] = !empty($entry['label']) ? $entry['label'] : $entry['orderlabel'];
+        $entryArray['orderlabel'] = $entry['orderlabel'];
+        $entryArray['type'] = Helper::translate($entry['type'], 'tx_dlf_structures', $this->settings['storagePid']);
+        $entryArray['pagination'] = htmlspecialchars($entry['pagination']);
+        $entryArray['doNotLinkIt'] = 1;
+        $entryArray['ITEM_STATE'] = 'HEADER';
+
+        if ($entry['children'] == NULL) {
+            $entryArray['description'] = $entry['description'];
+            $id = str_replace("LOG", "PHYS", $entry['id']);
```
sebastian-meyer: `$entry['id']` holds a XML ID, right? Then you can't rely on them always being `LOG...` and `PHYS...`. This might be the case for METS files produced by Kitodo.Production, but it's not standardized. Instead of assuming ID values here, we should use the actual values taken from the METS file.

But besides that, why is it even necessary to replace the ID value here? If the references in the METS file are pointing to the wrong target, we should fix the METS file rather then implementing a workaround in the code.

sebastian-meyer (changes requested):
Sorry for bothering you again, but I found another issue...

Classes/Controller/TableOfContentsController.php
beatrycze-volk: But the images are stored inside the physical structure and entry["id"] contains id of logical structure. Because of that, the id number is always the same but prefix needs to be changed.

sebastian-meyer: Yes, but nevertheless we can't just rely on the IDs always having the same numbering. You can get the physical structure ID related to a given logical structure ID via the array `$this->document->smLinks` (see in `Classes/Common/MetsDocument.php`.

beatrycze-volk: Thank you! That's exactly what I was needing here. Changed to read id from smLinks array.

sebastian-meyer (approved):

PR #812, opened 2022-05-11
[FEATURE] show divs for thumbnails and descriptions only when needed and add link

BFallert:
- only insert divs for thumbnails and description if the related fields have content.
- add a link to the collection page to the collection thumbnail image.

**This eliminates the need for customization in slub_digitalcollections:**
The link to the collection thumbnail replaces a short javascript code (marked with: "// Add click event to complete collections element on intro page") in the [DigitalcollectionsListScripts.js file](https://github.com/slub/slub_digitalcollections/blob/master/Resources/Private/JavaScript/DigitalcollectionsListScripts.js#L59) in slub_digitalcollections.

### DIFF
--- a/Resources/Private/Templates/Collection/List.html
+++ b/Resources/Private/Templates/Collection/List.html
@@ -36,16 +36,20 @@
                         </span>
                     </f:if>
                 </h2>
-                <f:if condition="{item.collection.thumbnail}"><div class="tx-dlf-collection-thumbnail">
-                    <f:link.action action="show"
-                                   arguments="{collection : item.collection}"
-                                   pageUid="{pageUid}">
-                        <f:image image="{item.collection.thumbnail}" alt="{f:translate(key: 'thumbnail')} {item.collection.label}" title="{item.collection.label}" maxWidth="250" />
-                    </f:link.action>
-                </div></f:if>
-                <f:if condition="{item.collection.description}"><div class="tx-dlf-collection-description">
-                    <f:format.html>{item.collection.description}</f:format.html>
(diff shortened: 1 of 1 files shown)


Resources/Private/Templates/Collection/List.html
```diff
@@ -36,14 +36,16 @@ <h2>
                         </span>
                     </f:if>
                 </h2>
-                <div class="tx-dlf-collection-thumbnail">
-                    <f:if condition="{item.collection.thumbnail}">
+                <f:if condition="{item.collection.thumbnail}"><div class="tx-dlf-collection-thumbnail">
+                    <f:link.action action="show"
+                                   arguments="{collection : item.collection}"
+                                   pageUid="{pageUid}">
```
albig: Yes, this makes sense and should work.

Could you please reformat the template tags? The source file was wrong, too. But you could fix it easily. E.g. like the following (should be examined in an IDE/Editor):
```suggestion
                <f:if condition="{item.collection.thumbnail}">
                    <div class="tx-dlf-collection-thumbnail">
                        <f:link.action action="show"
                                       arguments="{collection : item.collection}"
                                       pageUid="{pageUid}">
```

albig (changes requested):
Please 
* reformat the fluid tags
* make a PR afterwards to slub_digitalcollections to remove the not needed JavaScript code

albig (approved):

BFallert:
Done: make a PR afterwards to slub_digitalcollections to remove the not needed JavaScript code

PR #815, opened 2022-05-12
[BUGFIX] Fix all-year calendar title and restore previous markup

dvoracek-slub:
The adjustments to markup are mostly so that styles won't have to be adjusted.


albig (approved):
Thanks for fixing!

PR #810, opened 2022-05-10
[FEATURE] Integrate JS for 3D viewer

beatrycze-volk:
Depends on #686

PR #817, opened 2022-05-17
[FEATURE] Use proxy for loading 3D models

beatrycze-volk:
Depends on #810


Classes/Controller/View3DController.php
```diff
@@ -41,21 +41,23 @@ public function mainAction()
...
                 // Configure @action URL for form.
                 $uri = $this->uriBuilder->reset()
                     ->setTargetPageUid($GLOBALS['TSFE']->id)
                     ->setCreateAbsoluteUri(!empty($this->settings['forceAbsoluteUrl']) ? true : false)
                     ->setArguments([
                         'eID' => 'tx_dlf_pageview_proxy',
-                        'url' => urlencode($url),
-                        'uHash' => GeneralUtility::hmac($url, 'PageViewProxy')
+                        'url' => $url,
+                        'uHash' => 'View3DController'
```
sebastian-meyer: Why did you remove the uHash? This makes Kitodo.Presentation vulnerable to attacks. Please keep this the same as in the PageView plugin. (You can even keep the same message `PageViewProxy`.)
```suggestion
                        'uHash' => GeneralUtility::hmac($url, 'PageViewProxy')
```

Classes/Eid/PageViewProxy.php
```diff
@@ -130,7 +130,9 @@ protected function handleGet(ServerRequestInterface $request): ResponseInterface
 
         // get and verify the uHash
         $uHash = (string) ($queryParams['uHash'] ?? '');
-        if (!hash_equals(GeneralUtility::hmac($url, 'PageViewProxy'), $uHash)) {
+        if ($uHash == 'View3DController') {
+            // don't verify hash
+        } else if (!hash_equals(GeneralUtility::hmac($url, 'PageViewProxy'), $uHash)) {
```
sebastian-meyer: See above: this should stay unchanged for security reasons.
```suggestion
        if (!hash_equals(GeneralUtility::hmac($url, 'PageViewProxy'), $uHash)) {
```

Resources/Private/Templates/View3D/Main.html
```diff
@@ -13,7 +13,7 @@
       xmlns:f="http://typo3.org/ns/TYPO3/CMS/Fluid/ViewHelpers"
       data-namespace-typo3-fluid="true">
 
-    <p id="DFG_3DViewer" 3d="{url}" class="tx-dlf-view-3d"></p>
+    <p id="DFG_3DViewer" 3d="{url}" 3d="{proxy}" class="tx-dlf-view-3d"></p>
```
sebastian-meyer: Is it intentional, that both the plain URL and the Proxy-URL use the same attribute name?

sebastian-meyer (changes requested):

Resources/Private/Templates/View3D/Main.html
sebastian-meyer: Do we even need the plain URL when the proxy should be used?

Classes/Controller/View3DController.php
beatrycze-volk: Because if the file name of model is changed than proxy is failing. You know this problem with the incorrect names of the model and then changed suffix. After suffix is adjusted our URL needs to also change and then it all doesn't work anymore.

Resources/Private/Templates/View3D/Main.html
beatrycze-volk: > Is it intentional, that both the plain URL and the Proxy-URL use the same attribute name?

No, in later changes it is adjusted to _proxy_.

> Do we even need the plain URL when the proxy should be used?

The original one (_3d_) is used when proxy URL needs to be modified with the corrected 3D model.

Classes/Controller/View3DController.php
sebastian-meyer: OK, then @thedworak needs to find a fix for this issue even more urgently. We can't open up a security issue in Kitodo.Presentation just because Drupal isn't able to give the right name of an uploaded file. @thedworak has to make sure that the METS file contains a valid file URL - either by finding a way to get the right file name from Drupal in the first place or by implementing some kind of transparent redirecting on the Drupal side when calling a file without its suffix. We have to be able to rely on the path given in the METS file!

Resources/Private/Templates/View3D/Main.html
sebastian-meyer: OK, thanks for the explanation!

sebastian-meyer:
There are still Codacy issues in `main.js`. Maybe that is something, @thedworak can take a look at?

thedworak:
> There are still Codacy issues in `main.js`. Maybe that is something, @thedworak can take a look at?

I really need more details here. Don't even know what you really need here.

sebastian-meyer:
You can find the issues on Codacy: https://app.codacy.com/gh/kitodo/kitodo-presentation/pullRequest?prid=9443702

sebastian-meyer:
And for the file name issue see above: https://github.com/kitodo/kitodo-presentation/pull/817#discussion_r1000780270

Classes/Controller/View3DController.php
thedworak: Maybe from the beginning: Why you don't have a right filename for downloading?

beatrycze-volk: The file name can be modified inside the JS if XML contains the incorrect one, right? Drupal problem with the incorrect model file naming. If suffix is changing then proxy link is immediately invalidated.

thedworak:
> You can find the issues on Codacy: https://app.codacy.com/gh/kitodo/kitodo-presentation/pullRequest?prid=9443702

I think it should be solved right now. Moreover there is a content-length header in the curl request:
image

beatrycze-volk:
> > You can find the issues on Codacy: https://app.codacy.com/gh/kitodo/kitodo-presentation/pullRequest?prid=9443702
> 
> I think it should be solved right now. Moreover there is a content-length header in the curl request: image

And the file naming? I see that it is in the big part commented out.

thedworak:
> > > You can find the issues on Codacy: https://app.codacy.com/gh/kitodo/kitodo-presentation/pullRequest?prid=9443702
> > 
> > 
> > I think it should be solved right now. Moreover there is a content-length header in the curl request: image
> 
> And the file naming? I see that it is in the big part commented out.

https://github.com/kitodo/kitodo-presentation/pull/817#discussion_r1005283163
I see you have a problem but which part of the code you are thinking of? We should be more precise. I'm not sitting in your heads and it's really hard for me to understand your point of view :)

sebastian-meyer:
> [#817 (comment)](https://github.com/kitodo/kitodo-presentation/pull/817#discussion_r1005283163) I see you have a problem but which part of the code you are thinking of? We should be more precise. I'm not sitting in your heads and it's really hard for me to understand your point of view :)

I am sorry, I'll try to explain:

First of all, in order to load the models from the repository outside of the repository context (i. e. from another domain), we need permission from the repository's server. This is a modern browser's feature, so we can't get around this easily. Permission is granted by setting Cross Origin Resource Sharing (CORS) headers. So, setting these headers on the repository's side would be the easiest solution for us.

But since adding the headers on your side seems to be impossible, we are using a proxy script, which fetches the model from your server and then delivers it with added CORS headers to the viewer. So from a browser's perspective the model is no longer delivered by the repository but by the DFG-Viewer, and has valid CORS headers - so everything is fine.

But in order to make sure that users can't just fetch any file via our proxy (including local files from our server) by tampering with the proxy link, we must add a uHash. This hash is calculated by using the original link to the 3D model given in the METS file (which is the same link our proxy is supposed to fetch) and makes sure that only exactly this URL can be fetch via the proxy. But this obviously doesn't work any more as soon as the proxy link gets changed, which is the case in `main.js` (e. g. when fetching the metadata from WissKI in method `fetchSettings()` or in method `init()` when loading the models). That's why @beatrycze-volk has removed checking the uHash, which makes the viewer work but opens up a security hole. (In the comment above I am complaining about the removal and that's where she explained that this is needed for the viewer.)

I am not sure if we can fully address this by just making sure that the link in the METS file is the definitive file name in Drupal (with correct suffixes, etc.), because it seems to me that you need to fetch additional files from the repository inside the viewer JS anyway. Is this correct?

So, to conclude:
- The best way of handling this issues would be to have valid CORS headers on the WissKI/Drupal side. Then we could turn off our proxy and everything should work just fine.
- If that's still not possible, we would need to make sure, that every file you are trying to fetch from WissKI/Drupal inside the viewer JS is correctly referenced in the METS file in order to allow us to provide valid proxy links for those files to the viewer. You would then need to use those links without changing them.

And on a side note: There seems to be WissKI/Drupal specific code in the viewer JS, like e. g. assuming paths or file names to fetch additional metadata. I think this is fine for our current project phase, but in phase 2 we need to aim for a more independent approach in order to be able to display models from other repositories as well.

thedworak:
Ok, thank you for detailed explanation, now I can fully understand the background. I can whitelist your domain for CORS policy.
Would it be ok for you?
Header add Access-Control-Allow-Origin "3d.dfg-viewer.de"

sebastian-meyer:
> Ok, thank you for detailed explanation, now I can fully understand the background. I can whitelist your domain for CORS policy. Would it be ok for you? Header add Access-Control-Allow-Origin "3d.dfg-viewer.de"

As far as I understand, you would need to set `Access-Control-Allow-Origin: "*"` since the requests to fetch the files don't come from our domain, but from the users' browsers (because it's your JS that fetches them).
But I am not sure about that, so maybe let's just try it out!

beatrycze-volk:
I have disabled proxy and it looks that unfortunately @sebastian-meyer is right:
cors

sebastian-meyer:
This actually looks like we don't need `Access-Control-Allow-Origin: "*"`, but `Access-Control-Allow-Origin: "https://3d.dfg-viewer.de"` would work as well (the scheme needs to be included!).

**BUT:** This would only work for the instance under 3d.dfg-viewer.de. In order to also support our test and development system, we would again need to allow another domain. So, we either need `Access-Control-Allow-Origin: "*"` to support both with a single setting or the server would have to set the header dynamically depending on the remote host.

That's actually how we do it in Kitodo.Presentation: https://github.com/kitodo/kitodo-presentation/blob/89aec4cafaee11c5c38700740f8c1c6f1a68fe72/Classes/Eid/PageViewProxy.php#L63-L73
We check for the `Origin` header of the request and set the CORS header accordingly. We set it to "*" only if there is no `Origin` header given.

thedworak:
I set up  Header add Access-Control-Allow-Origin "*", please check it again and report if there are more issues.

sebastian-meyer (approved):

PR #821, opened 2022-06-09
[BUGFIX] Fix various issues in page view and search

dvoracek-slub:
See intranet page.

The main open issue is that currently, all search results (up to 10000) are fetched and processed in `findSolrByCollection`. An option would be to implement `QueryResultInterface`.


dvoracek-slub:
Update: For season splits, the commit "Fix calendar: Collect all years for rendering" only worked when `showEmptyMonths = 0`. This is now improved/fixed.

sebastian-meyer (approved):

PR #826, opened 2022-07-05
[FEATURE] Add transformation to METS if transformation file is passed in parameter

beatrycze-volk:
This PR provides possibility to pass document which is not METS document but should be transformed to METS with given XSLT transformation.

This change demands adjustment in slub_digitalcollections: https://github.com/slub/slub_digitalcollections/pull/44


sebastian-meyer:
I won't merge this for now, because I still hope, that the XSLT transformation will be done by the repository. Ideally, this should not be part of the DFG-Viewer or Kitodo.Presentation.

sebastian-meyer:
Closing this since the feature is no longer needed.

PR #828, opened 2022-07-06
[FEATURE] Add reading of metadata from MODS for 3D objects

beatrycze-volk:
Next step will be including reading of personal data from ORCID API and corporate from VIAF API.


Classes/Format/Mods.php
```diff
@@ -78,56 +119,216 @@ public function extractMetadata(\SimpleXMLElement $xml, array &$metadata)
...
+    /**
+     * Get "description" stored in recordInfoNote.
+     *
+     * @access private
+     *
+     * @return void
+     */
+    private function getDescription() {
+        $this->getSingleMetadata('description', './mods:recordInfo/mods:recordInfoNote/text()');
+    }
```
sebastian-meyer: An implementation in this class is only necessary for metadata which can't be queried by XPath alone (like for example author or holder, where multiple possible encodings have to be taken in account). Everything else should be configured in the backend (with defaults in `Resources/Private/Data/MetadataDefaults.php`).

Classes/Format/Mods.php
```diff
@@ -78,56 +119,216 @@ public function extractMetadata(\SimpleXMLElement $xml, array &$metadata)
...
+    /**
+     * Get "identifier" - example: hotels (built public accommodations), AAT ID: 300007166.
+     *
+     * @access private
+     *
+     * @return void
+     */
+    private function getIdentifier() {
+        $this->getSingleMetadata('identifier', './mods:identifier/text()');
+    }
```
sebastian-meyer: See above: please add this as default mapping to `Resources/Private/Data/MetadataDefaults.php`.

Classes/Format/Mods.php
```diff
@@ -78,56 +119,216 @@ public function extractMetadata(\SimpleXMLElement $xml, array &$metadata)
...
+    /**
+     * Get "license" - license on which object is published.
+     *
+     * @access private
+     *
+     * @return void
+     */
+    private function getLicense() {
+        $this->getSingleMetadata('license', './mods:accessCondition/text()');
+    }
```
sebastian-meyer: See above: please add this as default mapping to `Resources/Private/Data/MetadataDefaults.php`.

Classes/Format/Mods.php
```diff
@@ -78,56 +119,216 @@ public function extractMetadata(\SimpleXMLElement $xml, array &$metadata)
...
+     *   - alternative names
+     *
+     * @access private
+     *
+     * @return void
+     */
+    private function getObjectNames() {
+        $this->getSingleMetadata('object_name', './mods:relatedItem/mods:titleInfo[not(@displayLabel="alternative")]/mods:title/text()');
+        $this->getSingleMetadata('object_alternative_names', './mods:relatedItem/mods:titleInfo[@displayLabel="alternative"]/mods:title/text()');
+    }
```
sebastian-meyer: See above: please add this as default mappings to `Resources/Private/Data/MetadataDefaults.php`.

Classes/Format/Mods.php
```diff
@@ -78,56 +119,216 @@ public function extractMetadata(\SimpleXMLElement $xml, array &$metadata)
...
+     * @access private
+     *
+     * @return void
+     */
+    private function getObjectLocationMetadata() {
+        $this->getSingleMetadata('object_location', './mods:relatedItem/mods:location/mods:physicalLocation/text()');
+        $this->getSingleMetadata('geonames', './mods:relatedItem/mods:location/mods:url[@displayLabel="geonames"]/text()');
+        $this->getSingleMetadata('wikidata', './mods:relatedItem/mods:location/mods:url[@displayLabel="wikidata"]/text()');
+        $this->getSingleMetadata('wikipedia', './mods:relatedItem/mods:location/mods:url[@displayLabel="wikipedia"]/text()');
+    }
```
sebastian-meyer: See above: please add this as default mappings to `Resources/Private/Data/MetadataDefaults.php`.

Classes/Format/Mods.php
```diff
@@ -78,56 +119,216 @@ public function extractMetadata(\SimpleXMLElement $xml, array &$metadata)
...
+     * @return void
+     */
+    private function getSingleMetadata($metadataIndex, $xpath) {
+        $results = $this->xml->xpath($xpath);
+        if (!empty($results)) {
+            foreach ($results as $result) {
+                $this->metadata[$metadataIndex][0] = (string) $result;
+            }
+        }
+    }
```
sebastian-meyer: See above: this method isn't needed if all single metadata mappings are moved to the default configuration.

Classes/Format/Mods.php
```diff
@@ -78,56 +119,216 @@ public function extractMetadata(\SimpleXMLElement $xml, array &$metadata)
...
+                        } else {
+                            $name[$k] = (string) $namePart;
+                        }
+                        $k++;
+                    }
+                    ksort($name);
+                    $this->metadata['holder'][$i] = trim(implode(', ', $name));
                 }
                 // Append "valueURI" to name using Unicode unit separator.
                 if (isset($authors[$i]['valueURI'])) {
```
sebastian-meyer:
```suggestion
                if (isset($holders[$i]['valueURI'])) {
```

Classes/Format/Mods.php
```diff
@@ -78,56 +119,216 @@ public function extractMetadata(\SimpleXMLElement $xml, array &$metadata)
...
-                    $metadata['author'][$i] .= chr(31) . (string) $authors[$i]['valueURI'];
+                    $this->metadata['holder'][$i] .= chr(31) . (string) $authors[$i]['valueURI'];
```
sebastian-meyer:
```suggestion
                    $this->metadata['holder'][$i] .= chr(31) . (string) $holders[$i]['valueURI'];
```

Classes/Format/Mods.php
```diff
@@ -34,20 +48,47 @@ class Mods implements \Kitodo\Dlf\Common\MetadataInterface
...
+
+        $this->getAuthors();
+        $this->getHolders();
+        $this->getPlaces();
+        $this->getYears();
+        $this->getDescription();
+        $this->getIdentifier();
+        $this->getLicense();
+        $this->getObjectNames();
+        $this->getObjectLocationMetadata();
```
sebastian-meyer:
```suggestion
```

Classes/Common/MetsDocument.php
```diff
@@ -459,6 +460,13 @@ public function getMetadata($id, $cPid = 0)
             'out_of_print' => [],
             'rights_info' => [],
             'collection' => [],
+            'object_name' => [],
+            'object_alternative_names' => [],
+            'object_location' => [],
+            'geonames' => [],
+            'wikidata' => [],
+            'wikipedia' => [],
+            'object_reconstruction' => [],
```
sebastian-meyer: This array needs to contain only those metadata fields, which are strictly required by other plugins/modules. So `holder` seems to be a necessary addition, since it is referenced in `Classes/Format/Mods.php`, but all the other additions should be omitted.

sebastian-meyer (changes requested):

Classes/Format/Mods.php
beatrycze-volk: Thanks for info! I will make changes according to it in this PR.

Classes/Format/Mods.php
```diff
@@ -22,6 +22,20 @@
  */
 class Mods implements \Kitodo\Dlf\Common\MetadataInterface
 {
+        /**
```
sebastian-meyer:
```suggestion
    /**
```

Resources/Private/Data/MetadataDefaults.php
```diff
@@ -520,5 +520,138 @@
...
+        'is_sortable' => 0,
+        'is_facet' => 0,
+        'is_listed' => 0,
+        'index_autocomplete' => 0,
+    ],
+    'object_name' => [
+        'format' => [
+            [
+                'format_root' => 'mods',
+                'xpath' => './mods:relatedItem/mods:titleInfo[not(@displayLabel="alternative")]/mods:title',
```
sebastian-meyer: Are you sure about the xpath? Normally "alternative" would be a typical value for the `type` attribute, but here you are querying the `displayLabel` attribute for it. Also, all metadata about the original object (rather than the 3D model) should be bundled within `./mods:relatedItem[@type="original"]`. I'd expect the xpath to look something like this:
`./mods:relatedItem[@type="original"]/mods:titleInfo[not(@type="alternative")]/mods:title`

Otherwise this seems to be a very specific and non-generic way of mapping an object's name. It's OK to do so, but then I'd suggest not putting this mapping into the Kitodo.Presentation defaults. So maybe leave it out here and instead put it into a pull request for the DFG-Viewer?

Resources/Private/Data/MetadataDefaults.php
```diff
@@ -520,5 +520,138 @@
...
+        'is_sortable' => 0,
+        'is_facet' => 0,
+        'is_listed' => 0,
+        'index_autocomplete' => 0,
+    ],
+    'object_alternative_names' => [
+        'format' => [
+            [
+                'format_root' => 'mods',
+                'xpath' => './mods:relatedItem/mods:titleInfo[@displayLabel="alternative"]/mods:title',
```
sebastian-meyer: See above: I'd expect this mapping to be querying the `type` attribute rather than the `displayLabel`.

Resources/Private/Data/MetadataDefaults.php
```diff
@@ -520,5 +520,138 @@
...
+        'is_sortable' => 0,
+        'is_facet' => 0,
+        'is_listed' => 0,
+        'index_autocomplete' => 0,
+    ],
+    'object_location' => [
+        'format' => [
+            [
+                'format_root' => 'mods',
+                'xpath' => './mods:relatedItem/mods:location/mods:physicalLocation',
```
sebastian-meyer: The object's location should be mapped to
`./mods:relatedItem[@type="original"]/mods:location/mods:physicalLocation`

Resources/Private/Data/MetadataDefaults.php
```diff
@@ -520,5 +520,138 @@
...
+        'is_sortable' => 0,
+        'is_facet' => 0,
+        'is_listed' => 0,
+        'index_autocomplete' => 0,
+    ],
+    'geonames' => [
+        'format' => [
+            [
+                'format_root' => 'mods',
+                'xpath' => './mods:relatedItem/mods:location/mods:url[@displayLabel="geonames"]',
```
sebastian-meyer: The object's location's geonames URI should be mapped to
`./mods:relatedItem[@type="original"]/mods:location/mods:physicalLocation[@authority="geonames"]/@valueURI`

Resources/Private/Data/MetadataDefaults.php
```diff
@@ -520,5 +520,138 @@
...
+        'is_sortable' => 0,
+        'is_facet' => 0,
+        'is_listed' => 0,
+        'index_autocomplete' => 0,
+    ],
+    'wikidata' => [
+        'format' => [
+            [
+                'format_root' => 'mods',
+                'xpath' => './mods:relatedItem/mods:location/mods:url[@displayLabel="wikidata"]',
```
sebastian-meyer: Is Wikidata only used for geographical information? Then it should be mapped to
`./mods:relatedItem[@type="original"]/mods:location/mods:physicalLocation[@authority="wikidata"]/@valueURI`

Otherwise, if we are linking to general information about the object, we should map to
`./mods:relatedItem[@type="original"]/mods:location/mods:url[@note="wikidata"]`

Resources/Private/Data/MetadataDefaults.php
```diff
@@ -520,5 +520,138 @@
...
+        'is_sortable' => 0,
+        'is_facet' => 0,
+        'is_listed' => 0,
+        'index_autocomplete' => 0,
+    ],
+    'wikipedia' => [
+        'format' => [
+            [
+                'format_root' => 'mods',
+                'xpath' => './mods:relatedItem/mods:location/mods:url[@displayLabel="wikipedia"]',
```
sebastian-meyer: See above: Wikipedia should be mapped the same way as Wikidata.

sebastian-meyer (changes requested):
Please see my comments below: This pull request is fine regarding the code, but some of the metadata mappings should be further discussed. I am OK with merging it anyways, because the mappings can be adjusted later and don't affect anything other than 3D models.

A general remark though: All the metadata mappings you added don't have a default `wrap`, are not indexed, not used as a facet and not used for sorting by default. This is bad practice, since this means the mappings don't have any effect... Since most of the metadata are meant to be displayed, please add at least a default `wrap` like this:
`"key.wrap = <dt>|</dt>\nvalue.required = 1\nvalue.wrap = <dd>|</dd>"`

Resources/Private/Data/MetadataDefaults.php
beatrycze-volk: But this doesn't match the MODS structure which we are currently using:
```
<mods:location>
  <mods:physicalLocation valueURI=""/>
  <mods:url displayLabel="geonames"/>
  <mods:url displayLabel="wikidata"/>
 <mods:url displayLabel="wikipedia"/>
</mods:location>
```

sebastian-meyer: Yes, I know. I am already in discussions with our colleagues in order to change the MODS structure accordingly. We should leave this as it is for now, but will need to change it later when the MODS structure got changed.

Resources/Private/Data/MetadataDefaults.php
beatrycze-volk: As above.

Should be transformation rewritten to match this XPath?

Resources/Private/Data/MetadataDefaults.php
beatrycze-volk: As above.

Should be transformation rewritten to match this XPath?

Resources/Private/Data/MetadataDefaults.php
sebastian-meyer: > Should be transformation rewritten to match this XPath?

Yes, please. That'd be great!

Resources/Private/Data/MetadataDefaults.php
beatrycze-volk: So the end result should be like that?
```
<mods:location>
  <mods:physicalLocation valueURI=""/>
  <mods:physicalLocation authority="geonames" valueURI="https://geonames..." />
  <mods:url note="wikidata"/>
 <mods:url note="wikipedia"/>
</mods:location>
```

Wikipedia links point to general Wikipedia pages for objects, the same as Wikidata. I don't think that there are some exact pages which contain only geographical information.

sebastian-meyer: Since "object location" and "geonames" refers to the same location, it should look like this:
```
<mods:location>
    <mods:physicalLocation authority="geonames" valueURI="[GEONAMES URI]">[LOCATION]</mods:physicalLocation>
    <mods:url note="wikidata">[WIKIDATA URI]</mods:url>
    <mods:url note="wikipedia">[WIKIPEDIA URI]</mods:url>
</mods:location>
```

beatrycze-volk: Transformation updated in DFG Viewer

sebastian-meyer (approved):

PR #829, opened 2022-07-13
[FEATURE] Integrate ORCID API

beatrycze-volk:
The author data like address, email and names can be read from the API if ORCID identifier is included in MODS.

Depends on #828


beatrycze-volk:
I have discovered error with different formats of given URLs:
- https://orcid.org/my-orcid?orcid=0000-0002-0419-9171
- https://orcid.org/0000-0001-7032-6696
Current implementation covers only second. Changed to draft so first option also gets coverage.

sebastian-meyer:
The first link is not a valid ORCID identifier (but a deep link to a person's personal account page) and we should not support it. Otherwise we would need to support any kind of ORCID link.

Only ORCID URIs of the form `https://orcid.org/XXXX-XXXX-XXXX-XXXX`are persistent identifiers and should be supported.

If someone enters an invalid ORCID link, it should get rejected by the repo's ingest form and not even be added to the XML metadata.

beatrycze-volk:
In current test case, half of the links are inserted as `https://orcid.org/XXXX-XXXX-XXXX-XXXX `and other half as `https://orcid.org/my-orcid?orcid=XXXX-XXXX-XXXX-XXXX`. I will remove part which handles second type.

Classes/Api/Orcid/Client.php
```diff
@@ -0,0 +1,124 @@
...
+    }
+
+    /**
+     * Creates the qualified API endpoint for retrieving the desired data
+     *
+     * @return string
+     **/
+    private function getApiEndpoint()
+    {
+        $url  = 'https://' . $this->level . '.' .  self::HOSTNAME;
```
sebastian-meyer:
```suggestion
        $url  = 'https://' . $this->level . '.' . self::HOSTNAME;
```

Classes/Api/Orcid/Client.php
```diff
@@ -0,0 +1,124 @@
...
+        $url .= '/v' . self::VERSION . '/';
+        $url .= '0000-0001-9483-5161';
+        //$url .= $this->orcid;
```
sebastian-meyer, line 120: This seems to be some forgotten debug code or is it an intentional self-plug? ;o)

Classes/Format/Mods.php
```diff
@@ -34,98 +50,206 @@ class Mods implements \Kitodo\Dlf\Common\MetadataInterface
...
-                        ) {
-                            $name[3] = (string) $namePart;
-                        } else {
-                            $name[$k] = (string) $namePart;
+
+                $identifier = $authors[$i]->xpath('./mods:nameIdentifier');
+                if ($identifier[0]['type'] == 'orcid' && !empty((string) $identifier[0])) {
+                    $orcidIdParts = explode('/', (string) $identifier[0]);
+                    $orcidId = trim(end($orcidIdParts));
+                    if (!str_contains($orcidId, 'orcid=')) {
```
sebastian-meyer: This is only necessary because of a faulty metadata mapping. The ORCID ID should be mapped to `./mods:name/mods:nameIdentifier[@type="orcid"]` as a plain ID value, not as full URL. So there shouldn't be any need to do string magic here in order to get the ID...
This should be fixed in the metadata mapping and removed from the code here.

sebastian-meyer (changes requested):

Classes/Api/Orcid/Client.php
sebastian-meyer: I just saw that this is already fixed in pull request #834. So no need to change it here!

Classes/Format/Mods.php
beatrycze-volk: Currently we are getting this data as URL:
```
<author_orcid>https://orcid.org/0000-0001-8356-8483</author_orcid>`
<holder_viaf>http://viaf.org/viaf/170713989</holder_viaf>
```

Should be string magic made in XSLT transformation?

sebastian-meyer: Since `mods:nameIdentifier` unfortunately doesn't allow the `@valueURI` attribute (where we could put the whole URL into), I think we need to do the string magic in XSLT in order to get the plain ID.

sebastian-meyer (approved):

PR #833, opened 2022-07-14
[FEATURE] Update 3D objects list filtering with filtering by type

beatrycze-volk:
Depends on #803


sebastian-meyer:
During our trip to Mainz I've discussed with Piotr and Sander that we will drop support for object lists in the DFG Viewer in favor of the search and navigation interface in the repository itself. So this pull request and #836, #837 and #852 are obsolete now. I'll therefore not merge them, but kindly ask you to remove the already integrated handling of object lists.

sebastian-meyer:
Closing this because feature is no longer needed.

PR #834, opened 2022-07-19
[FEATURE] Integrate VIAF API

beatrycze-volk:
Depends on #829

### DIFF
--- a/Classes/Api/Orcid/OrcidClient.php
+++ b/Classes/Api/Orcid/OrcidClient.php
@@ -25,7 +25,7 @@ use TYPO3\CMS\Core\Utility\GeneralUtility;
  * @subpackage dlf
  * @access public
  **/
-class Client
+class OrcidClient
 {
     /**
      * constants for API endpoint
@@ -56,7 +56,7 @@ class Client
     private $level = 'pub';
 
     /**
-     * The ORCID to search for
+     * The login/registration page ORCID
      *
      * @var string
      **/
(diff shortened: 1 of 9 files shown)


sebastian-meyer (dismissed):

Classes/Api/Orcid/OrcidClient.php
```diff
@@ -25,7 +25,7 @@
  * @subpackage dlf
  * @access public
  **/
-class Client
+class OrcidClient
```
sebastian-meyer: It's more consistent to keep the naming scheme `Classes/Api/Orcid/Client` and `Classes/Api/Viaf/Client` respectively. To avoid namespace duplicates when using both classes within the same file, declaration should be like:
`use Kitodo\Dlf\Api\Orcid\Client as OrcidClient;`

Classes/Api/Orcid/OrcidProfile.php
```diff
@@ -20,12 +20,12 @@
...
  * ORCID API Profile class
  *
-* @author Beatrycze Volk <beatrycze.volk@slub-dresden.de>
+ * @author Beatrycze Volk <beatrycze.volk@slub-dresden.de>
  * @package TYPO3
  * @subpackage dlf
  * @access public
  **/
-class Profile
+class OrcidProfile
```
sebastian-meyer: See above.

sebastian-meyer (changes requested):
Please have a look at the Scrutinizer issues and my remarks.

sebastian-meyer:
Thank you! Looks good now except for the Scrutinizer issues.

beatrycze-volk:
> Thank you! Looks good now except for the Scrutinizer issues.

Which issues? The PR is marked green by checks.

sebastian-meyer:
Have a look at the "Files changed" tab. There are some inline comments from Scrutinizer. They are not critical, that's why the status checks went through anyway.

sebastian-meyer (approved):
Thank you!

PR #835, opened 2022-07-20
[FEATURE] Two-level metadata: Model and appearance

fraschz:

PR #839, opened 2022-07-27
[FEATURE] Display links in metadata

beatrycze-volk:
Depends on #838


sebastian-meyer (approved):
We should find a more generic way of linking metadata to external URLs, but I'll approve this for now. We should keep this in mind when working on #835.

PR #842, opened 2022-07-29
[MAINTENANCE] GitHub Testing CI: Cleanup branches, run on PRs

dvoracek-slub:
Currently, the (few) PHP tests are run after pushing/merging into master, which can be seen in the Actions tab and in the status icon of a commit.

This PR configures the GitHub workflow to also run tests on pull requests, so the results are shown alongside Codacy & co. It's already active on this PR.

Also, branches that have already been merged (dev-extbase-fluid, dev-testing) are removed from the configuration.


sebastian-meyer (approved):

sebastian-meyer:
Is there a possibility to exclude some tests for version 9.5? They always fail anyway because they try to call methods which don't exist in 9.5.

sebastian-meyer:
> Is there a possibility to exclude some tests for version 9.5? They always fail anyway because they try to call methods which don't exist in 9.5.

OK, that should be resolved with #846

dvoracek-slub:
> > Is there a possibility to exclude some tests for version 9.5? They always fail anyway because they try to call methods which don't exist in 9.5.
> 
> OK, that should be resolved with #846

Yes, in that particular case I rather made the test work with TYPO3 v9.

In the general case:
* There is a PHPUnit `@requires` annotation that apparently can be used to test for a particular method, see https://phpunit.readthedocs.io/en/9.5/incomplete-and-skipped-tests.html#skipping-tests-using-requires.
* I wonder (don't know) if TYPO3 somehow extended the annotations. Otherwise, it's of course possible to just add a version guard to the test case.

sebastian-meyer:
Closing this pull request as a duplicate of #846 and #847.

PR #844, opened 2022-07-29
[FEATURE] In calendar, fall back to METS if DB is empty

dvoracek-slub:
This PR is based on #821, which contains a fix for the calendar.

When assembling the newspaper issues that are shown in the calendar, Kitodo only considers the entries that are stored in database. This doesn't work for DFG Viewer, obviously, which (I guess) is why DFG Viewer has previously maintained its own calendar implementation using only data from the METS file. In this PR, I try to unify these: If no issue is found in the database, Kitodo then falls back to the issues listed in the METS file.

Other possible approaches might have been to introduce a feature flag (to run calendar in DB mode vs METS mode), or to use a different condition to determine METS fallback (e.g., whether or not the parent document is indexed).

### DIFF
--- a/Classes/Controller/CalendarController.php
+++ b/Classes/Controller/CalendarController.php
@@ -219,7 +219,7 @@ class CalendarController extends AbstractController
         $this->view->assign('documentId', $this->document->getUid());
         $this->view->assign('yearLinkTitle', $yearLinkTitle);
         $this->view->assign('parentDocumentId', $this->document->getPartof());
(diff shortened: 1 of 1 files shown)


sebastian-meyer:
> Other possible approaches might have been to introduce a feature flag (to run calendar in DB mode vs METS mode), or to use a different condition to determine METS fallback (e.g., whether or not the parent document is indexed).

I think the latter would be the best solution: Use the DB if the parent document is indexed and use the METS file otherwise.

sebastian-meyer (approved):

PR #845, opened 2022-08-04
[BUGFIX] Multidomain typoscript configuration

### LINKED ISSUE #841: Multi domain breaks hooks from ItemsProcFunc.php
## Description

If you use more than one root page (multi domain) in your installation, it is possible that the lists "show these facets" and "fields for extended search" inside the configuration of the search plugin are empty. These lists are build with hooks from ItemsProcFunc.php, which needs the typoscript configuration to extract the storage pid and build up these lists. But in case of a multi domain installation, the typoscript configuration is restricted to a page tree and the default storage pid is used, which could be a different page tree than the plugin is placed.

## Reproduction

1. Go to the typo3 backend
2. Place a new page on root
3. Mark this page as root page
4. Go to the search plugin
5. Open plugin tab

## Expected Behavior

It must be possible to configure the page tree where the typoscript configuration should be read from, or it must depend on the plugin placement. Otherwise, it is not possible to define different configurations for each page tree. In our case, we have three different configurations. Two of them are on the same page tree and the other one is on a different tree.

christopher-timm:
Read the typoscript configuration depending on the page where it is placed on.
Otherwise, multi domain installation only get the default site root typoscript.
This PR is linked to issue #841

### DIFF
--- a/Classes/Hooks/ItemsProcFunc.php
+++ b/Classes/Hooks/ItemsProcFunc.php
@@ -67,15 +67,15 @@ class ItemsProcFunc implements LoggerAwareInterface
         $pid = $params['flexParentDatabaseRow']['pid'];
         $rootline = \TYPO3\CMS\Backend\Utility\BackendUtility::BEgetRootLine($pid);
         $siterootRow = [];
-        foreach($rootline as $_uid=>$_row) {
-            if($_row['is_siteroot'] == '1') {
+        foreach ($rootline as $_row) {
+            if ($_row['is_siteroot'] == '1') {
                 $siterootRow = $_row;
                 break;
             }
         }
 
         try {
-            $ts = $objectManager->get(\TYPO3\CMS\Core\TypoScript\TemplateService::class,[$siterootRow['uid']]);
+            $ts = $objectManager->get(\TYPO3\CMS\Core\TypoScript\TemplateService::class, [$siterootRow['uid']]);
             $ts->rootLine = $rootline;
             $ts->runThroughTemplates($rootline, 0);
             $ts->generateConfig();
--- a/Classes/Hooks/ItemsProcFunc.php
+++ b/Classes/Hooks/ItemsProcFunc.php
@@ -13,6 +13,8 @@
 namespace Kitodo\Dlf\Hooks;
 
 use Kitodo\Dlf\Common\Helper;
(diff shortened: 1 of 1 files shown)


Classes/Hooks/ItemsProcFunc.php
```diff
@@ -50,19 +50,38 @@ public function toolList(&$params)
...
+            $ts = $objectManager->get(\TYPO3\CMS\Core\TypoScript\TemplateService::class,[$siterootRow['uid']]);
+            $ts->rootLine = $rootline;
+            $ts->runThroughTemplates($rootline, 0);
+            $ts->generateConfig();
+        } catch (\Exception $e) {
+            die($e->getMessage());
+        }
+
+        $typoscriptConfig = $ts->setup;
+        $this->storagePid = $typoscriptConfig["plugin."]["tx_dlf."]["persistence."]["storagePid"];
```
sebastian-meyer:
```suggestion
        $this->storagePid = $typoscriptConfig['plugin.']['tx_dlf.']['persistence.']['storagePid'];
```

Classes/Hooks/ItemsProcFunc.php
```diff
@@ -50,19 +50,38 @@ public function toolList(&$params)
...
+            }
+        }
+
+        try {
+            $ts = $objectManager->get(\TYPO3\CMS\Core\TypoScript\TemplateService::class,[$siterootRow['uid']]);
+            $ts->rootLine = $rootline;
+            $ts->runThroughTemplates($rootline, 0);
+            $ts->generateConfig();
+        } catch (\Exception $e) {
+            die($e->getMessage());
```
sebastian-meyer: Please use proper error handling here instead of just exiting the script.

sebastian-meyer (changes requested):

christopher-timm:
Requested changes should be fixed now

sebastian-meyer (changes requested):
Please check the codacy issues. Apart from that, everything looks good, now!

sebastian-meyer (dismissed):
Perfect! I can merge this as soon as you bring your branch up to date with the base branch.

(When opening a pull request you can choose to allow me to commit changes to your branch. If you do so, I can update your branch myself.)

sebastian-meyer (approved):

PR #846, opened 2022-08-04
[BUGFIX] Don't rely on `LanguageService::create` in test

dvoracek-slub:
This PR fixes a test case that I added in d98b9b7d (PR #821) when reworking `Helper::getLanguageName()`. The test case uses `LanguageService::create()`, which is only available in TYPO3 v10. So when available, `LanguageServiceFactory` is used (this seems to be the recommended way, esp. from v11 onwards, see https://docs.typo3.org/c/typo3/cms-core/main/en-us/Changelog/11.3/Deprecation-94414-DeprecateLanguageServiceContainerEntry.html).


sebastian-meyer (approved):

sebastian-meyer:
Thank you! I'll merge this one, since it resolves the issue with `LanguageService::create()`. But there still is another issue with `SolrSearch::getNumHits()`. Could you please have a look into this one as well?

PR #857, opened 2022-10-13
[BUGFIX] Show search result range only if there exists any search result

stweil:
This avoids such messages:

    Einträge 1 bis 0 von 0.


stweil:
The broken CI check is unrelated to the code changed here.

sebastian-meyer (approved):

PR #858, opened 2022-10-13
[BUGFIX] Handle incomplete OAI plugin configuration

stweil:
Don't throw an exception when $library is null because the OAI plugin was not configured, but use the default values for oai_label and contact.

That makes the code more robust and helps admins who install Kitodo.Presentation.


stweil:
Related extract from TYPO3 log:

```
Uncaught TYPO3 Exception: Call to a member function getContact() on null
Uncaught TYPO3 Exception: Call to a member function getOaiLabel() on null
```

sebastian-meyer (approved):

PR #859, opened 2022-10-14
[BUGFIX] Remove text id AddToBasket and replace it by basket.addBasket

stweil:


stweil:
Both texts and their translations are identical, so only one of them should be needed.

sebastian-meyer (approved):

PR #861, opened 2022-10-25
[BUGFIX] Use title as title_sorting as default

### LINKED ISSUE #855: Indexing fails with latest code because of missing value for "title_sorting"
## Description

Running `vendor/bin/typo3 kitodo:index [...]` aborts with an exception.

## Reproduction

1. Run `kitodo:index` with valid arguments.
2. See error

```
Index single document into database and Solr.
=============================================


Uncaught TYPO3 Exception Argument 1 passed to Kitodo\Dlf\Domain\Model\Document::setTitleSorting() must be of the type string, null given, called in /var/www/dfgviewer/public/typo3conf/ext/dlf/Classes/Command/BaseCommand.php on line 187
thrown in file /var/www/dfgviewer/public/typo3conf/ext/dlf/Classes/Domain/Model/Document.php
in line 361
```

That exception is caused by an unset `$metadata['title_sorting'][0]`.
Setting it to `$metadata['title'][0]` might be a possible fix which makes indexing possible again.

## Expected Behavior

Indexing should run successfully.

sebastian-meyer:
This fixes #855


Classes/Common/MetsDocument.php
```diff
@@ -649,6 +649,9 @@ class_exists($class)
             $metadata['title'][0] = '';
             $metadata['title_sorting'][0] = '';
         }
+        // Set title_sorting to title as default.
+        if (empty($metadata['title_sorting'][0])) {
+            $metadata['title_sorting'][0] = $metadata['title'][0];
```
sebastian-meyer, line 654:
```suggestion
        // Set title_sorting to title as default.
        if (empty($metadata['title_sorting'][0])) {
            $metadata['title_sorting'][0] = $metadata['title'][0];
        }
```

beatrycze-volk (approved):

PR #869, opened 2022-11-24
[FEATURE] Integrate datesearch

### LINKED ISSUE #1852: [BUG] Date search fails if only "from" date is given
https://github.com/kitodo/kitodo-presentation/pull/869/files/475031f3f2c90fdf7930a2937e4abcda10b69683#diff-3ad891a4f6faca777eb6755324956d09c396c0dcbaf53b7ec276e6d9afcf5a54

@michaelkubina, @sebastian-meyer, I am afraid that this part of the code was never tested and never worked. Ordering date strings which can also be "*" or "NOW" leads to unexpected results. I noticed this when I entered a "from" date, and the search moved this date to the "to" field.

_Originally posted by @stweil in https://github.com/kitodo/kitodo-presentation/pull/869#discussion_r2845013967_

michaelkubina:
This PR integrates a datesearch, that can be activated within the search plugin and will be displayed on the website. When activated it presents two input fields for dates, so the users can input a range if necessary. When a datesearch was performed both the sorting form and the facets will recognise it and apply the filterquery as well. The filterquery is a simple range search (via join query parser over the uid) on the new mandatory solr-field "date". All language localizations have been applied.

This feature allows for an easy and convenient date-range search in any collection for all objects that have dates, e.g. in a newspaper-collection.


Classes/Controller/SearchController.php
```diff
@@ -106,6 +106,28 @@ public function mainAction()
...
+            $this->searchParams['dateTo'] = $this->searchParams['dateFrom'];
+        }
+        /* // alternative sanitize logic
+        if(empty($this->searchParams['dateFrom'])) {
+            $this->searchParams['dateFrom'] = "0000-01-01";
+        }
+        if(empty($this->searchParams['dateTo'])) {
+            $this->searchParams['dateTo'] = "9999-12-31";
+        }
+        */
```
sebastian-meyer: I suggest using your alternative logic, but with different default values: the default _from_ date should be the earliest timestamp available in the index (i. e. `*`) and the default _to_ date should be the current date (i. e. `NOW`).

This would result in queries like:
`[ * TO 1999-12-31 ]` (get everything older than the 21st century)
`[ 2000-01-01 TO NOW ]` (get everything from the 21st century)

Configuration/ApacheSolr/configsets/dlf/conf/schema.xml
```diff
@@ -142,6 +142,8 @@ limitations under the License.
         <!-- Next two fields are mandatory for identifying documents. -->
         <field name="title" type="standard" indexed="true" stored="true" multiValued="false" default="" />
         <field name="volume" type="standard" indexed="true" stored="true" multiValued="false" default="" />
+        <!-- The date e.g a newspaper was issued or a letter was written -->
+        <field name="date" type="string" indexed="true" stored="true" multiValued="false" default="" />
```
sebastian-meyer: Why don't you use a _DateRangeField_ field type for this? See https://solr.apache.org/guide/8_8/working-with-dates.html

Configuration/Flexforms/Search.xml
```diff
@@ -41,6 +41,27 @@
                             </config>
                         </TCEforms>
                     </settings.fulltext>
+                    <settings.datesearch>
+                        <TCEforms>
+                            <exclude>1</exclude>
+                            <label>LLL:EXT:dlf/Resources/Private/Language/locallang_be.xlf:plugins.search.flexform.datesearch</label>
+                            <config>
+                                <type>radio</type>
```
sebastian-meyer: Please use a checkbox, if it's just an on/off switch. It's more in line with other options and needs a bit less code.

Classes/Common/Indexer.php
```diff
@@ -336,6 +336,7 @@ protected static function processLogical(Document $document, array $logicalUnit)
             $solrDoc->setField('toplevel', $logicalUnit['id'] == $doc->toplevelId ? true : false);
             $solrDoc->setField('title', $metadata['title'][0], self::$fields['fieldboost']['title']);
             $solrDoc->setField('volume', $metadata['volume'][0], self::$fields['fieldboost']['volume']);
+            $solrDoc->setField('date', $metadata['date']);
```
sebastian-meyer: We need to make sure that `$metadata['date']` is always set when using it this way. Most documents will have a date, but there certainly are cases, where no date is given and therefore `$metadata['date']` is not set while retrieving the document's metadata. Please at least set `$metadata['date']` to an empty string or an appropriate default value in `MetsDocument->getMetadata()` and `IiifManifest->getMetadata()` to avoid this.

Furthermore, we have to validate the retrieved value and make sure it is properly formatted (i. e. according to XML's `dateTime`).

Resources/Private/Data/MetadataDefaults.php
```diff
@@ -80,6 +80,25 @@
         'is_listed' => 1,
         'index_autocomplete' => 0,
     ],
+    'date' => [
+        'format' => [
+            [
+                'format_root' => 'mods',
+                'xpath' => './mods:originInfo[@eventType="publication"]/mods:dateIssued[@encoding="iso8601"][@keyDate="yes"]',
```
sebastian-meyer: We should at least drop the `[@eventType="publication"]` condition and also allow the `w3cdtf` encoding in order to be applicable to a broader range of documents. I think, the most important indicator should be `[@keyDate="yes"]`. But it's still limited to `dateIssued`, so maybe we can even include `dateCreated` as a fallback? What do you think?

```suggestion
                'xpath' => './mods:originInfo/mods:dateIssued[@encoding="iso8601" or @encoding="w3cdtf"][@keyDate="yes"]',
```

sebastian-meyer (changes requested):

Please see my comments below and please resolve the merge conflicts.

Resources/Private/Data/MetadataDefaults.php
michaelkubina: I agree with removing the `[@eventType]` attribute and with adding the `w3cdtf` encoding. We could include `dateCreated` as a fallback...but the best i can think of is this XPATH 1.0 expression, that might return a set of two dates in some cases:

`./mods:originInfo/*[local-name()="dateIssued" or local-name()="dateCreated"][@encoding="iso8601" or @encoding="w3cdtf"][@keyDate="yes"]`

Classes/Controller/SearchController.php
michaelkubina: Changed it to the alternative logic, where an empty _from_ field fills to `*` and an empty _to_ field to `NOW` (we use `solr.dateRangeField` now). It does so only if the counterpart is not empty. This prevents the usage of the filterquery, when none of both dates is provided.

Configuration/ApacheSolr/configsets/dlf/conf/schema.xml
michaelkubina: Actually i was not aware of it, and i was focused on a narrow solution for dates in the format YYYY-MM-DD. After reading carefully i see why its so beneficial and that we could now even index fields in other formats and still do daterange searches on them - both for points and ranges in time. Thank you for the hint!

I added a `daterange` fieldtype to the schema and changed the type of the `date` field to it.

Classes/Common/Indexer.php
michaelkubina: The indexer now checks the date format. Dates in the format `YYYY`, `YYYY-MM`, `YYYY-MM-DD` will be kept as they are - especially `YYYY` and `YYYY-MM` to benefit from how the `dateRangeField` works with them. `YYYYMMDD` will be transformed to `YYYY-MM-DD`, to be complient to the necessary formating for this field type. Any `datetime` will be changed to the complient `YYYY-MM-DDThh:mm:ssZ` if necessary.  Empty and invalid dates will be skipped. Below a test script:

```
<?php
$dates = [
  "1987",
  "198710",
  "871002",
  "19871002",
  "19871002T19",
  "19871002T19Z",
  "19871002T1902",
  "19871002T1902Z",
  "19871002T190215",
  "19871002T190215Z",
  "19871002T190215.222",
  "19871002T190215.222Z",
  "1987",
  "1987-10",
  "87-10-02",
  "1987-10-02",
  "1987-10-02T19",
  "1987-10-02T19Z",
  "1987-10-02T19:02",
  "1987-10-02T19:02Z",
  "1987-10-02T19:02:15",
  "1987-10-02T19:02:15Z",
  "1987-10-02T19:02:15.222",
  "1987-10-02T19:02:15.222Z",
  
];

$isoDate = date('Y-m-d\TH:i:s\Z', strtotime('1987-10-02T19:02:15.222Z'));

echo $isoDate . "\n";
foreach ($dates as $date) {
  if (strtotime($date)) {
    echo $date . " is a valid date : " . date('c', strtotime($date));
  } else {
    echo $date . " is not a valid date!";
  }
  echo "\n";
}

?>
```
```
1987-10-02T19:02:15Z
1987 is a valid date : 1987-12-09T09:28:28+00:00
198710 is not a valid date!
871002 is not a valid date!
19871002 is a valid date : 1987-10-02T00:00:00+00:00
19871002T19 is not a valid date!
19871002T19Z is not a valid date!
19871002T1902 is a valid date : 1987-10-02T19:02:00+00:00
19871002T1902Z is a valid date : 1987-10-02T19:02:00+00:00
19871002T190215 is a valid date : 1987-10-02T19:02:15+00:00
19871002T190215Z is a valid date : 1987-10-02T19:02:15+00:00
19871002T190215.222 is not a valid date!
19871002T190215.222Z is not a valid date!
1987 is a valid date : 1987-12-09T09:28:28+00:00
1987-10 is a valid date : 1987-10-01T00:00:00+00:00
87-10-02 is a valid date : 1987-10-02T00:00:00+00:00
1987-10-02 is a valid date : 1987-10-02T00:00:00+00:00
1987-10-02T19 is not a valid date!
1987-10-02T19Z is not a valid date!
1987-10-02T19:02 is a valid date : 1987-10-02T19:02:00+00:00
1987-10-02T19:02Z is a valid date : 1987-10-02T19:02:00+00:00
1987-10-02T19:02:15 is a valid date : 1987-10-02T19:02:15+00:00
1987-10-02T19:02:15Z is a valid date : 1987-10-02T19:02:15+00:00
1987-10-02T19:02:15.222 is a valid date : 1987-10-02T19:02:15+00:00
1987-10-02T19:02:15.222Z is a valid date : 1987-10-02T19:02:15+00:00
```

Resources/Private/Data/MetadataDefaults.php
michaelkubina: went shorter on it, to take any `keyDate` with the specified encoding:
`./mods:originInfo/*[@encoding="iso8601" or @encoding="w3cdtf"][@keyDate="yes"]`

Resources/Private/Language/de.locallang_metadata.xlf
```diff
@@ -17,6 +17,10 @@
 				<source><![CDATA[Coordinates]]></source>
 				<target><![CDATA[Geoinformationen]]></target>
 			</trans-unit>
+			<trans-unit id="metadata.date" approved="yes">
+				<source><![CDATA[Date]]></source>
+				<target><![CDATA[Datum]]></target>
+      </trans-unit>
```
sebastian-meyer:
```suggestion
			<trans-unit id="metadata.date" approved="yes">
				<source><![CDATA[Date]]></source>
				<target><![CDATA[Datum]]></target>
			</trans-unit>
```

Resources/Private/Language/locallang_metadata.xlf
```diff
@@ -14,6 +14,9 @@
 			<trans-unit id="metadata.coordinates">
 				<source><![CDATA[Coordinates]]></source>
 			</trans-unit>
+			<trans-unit id="metadata.date">
+				<source><![CDATA[Date]]></source>
+      </trans-unit>
```
sebastian-meyer:
```suggestion
			<trans-unit id="metadata.date">
				<source><![CDATA[Date]]></source>
			</trans-unit>
```

sebastian-meyer (changes requested):

sebastian-meyer (approved):

Classes/Controller/SearchController.php
```diff
@@ -106,6 +106,19 @@ public function mainAction()
...
         }
 
+        // sanitize date search input
+        if(empty($this->searchParams['dateFrom']) && !empty($this->searchParams['dateTo'])) {
+            $this->searchParams['dateFrom'] = '*';
+        }
+        if(empty($this->searchParams['dateTo']) && !empty($this->searchParams['dateFrom'])) {
+            $this->searchParams['dateTo'] = 'NOW';
+        }
+        if($this->searchParams['dateFrom'] > $this->searchParams['dateTo']) {
```
stweil, line 116: @michaelkubina, @sebastian-meyer, I am afraid that this part of the code was never tested and never worked. Ordering date strings which can also be "*" or "NOW" leads to unexpected results. I noticed this when I entered a "from" date, and the search moved this date to the "to" field.

sebastian-meyer: You are right. Ordering may only happen when both dates are set!

PR #871, opened 2022-11-30
[BUGFIX] Reactivate the "back to listview" link.

christopher-timm:
Search parameters are now temporarily saved in the typo3 session to generate the link "back to the listview".


sebastian-meyer (changes requested):
This is fine apart from the Scrutinizer hint in `Classes\Controller\NavigationController.php`. Please have a look at that!

sebastian-meyer (approved):

PR #895, opened 2023-01-30
[MAINTENANCE] Added additional repository functional test instead of unit tests.

### LINKED ISSUE #781: [FUND] Add tests and significantly improve test coverage
## Description

## Related Issues

None.

haogatyp:
Instead of unit testing repositories it is recommended to do functional tests.


sebastian-meyer:
Could you please bring this pull request up to date with master? Alternatively you can allow me to update your branch, then I can do it myself via GitHub most of the time.

haogatyp:
I did a rebase on the current master.

oliver-stoehr:
@sebastian-meyer we have to remove the temporary file in TokenRepositoryTest.php:46, but according to Codacy
`The use of function shell_exec() is forbidden` and
`The use of function unlink() is discouraged`. 
Do you see any other way to remove the file or should I use `unlink` as it is "just" discouraged?
(I will fix the other Codacy issues.)

sebastian-meyer:
Yes, please revert this back to `unlink()`, since this is more portable than `shell_exec()` (which only works on UNIX and is disabled on many shared hosting services for security reasons). I'll then look into which rule actually discourages using `unlink()` and how to make an exception in this case.

sebastian-meyer (approved):

PR #901, opened 2023-02-06
[FEATURE] Add an option to list empty years in calender plugin

michaelkubina:
This PR adds an additional option, to list empty years in the calender plugin. This allows for an easier overview of the actual years available, since we can group whole decades in single rows. The benefit is a quicker eye browsing over the years, especially when there are gaps without any issues being published.

### DIFF
--- a/Classes/Controller/CalendarController.php
+++ b/Classes/Controller/CalendarController.php
@@ -282,7 +282,7 @@ class CalendarController extends AbstractController
                 ];
             }
             // create an array that includes years without issues
-            if($this->settings['showEmptyYears']){
+            if (!empty($this->settings['showEmptyYears'])){
                 $yearFilled = [];
                 $min = $yearArray[0]['title'];
                 // round the starting decade down to zero for equal rows
--- a/Classes/Controller/CalendarController.php
+++ b/Classes/Controller/CalendarController.php
@@ -282,20 +282,18 @@ class CalendarController extends AbstractController
                 ];
             }
             // create an array that includes years without issues
-            if (!empty($this->settings['showEmptyYears'])){
+            if (!empty($this->settings['showEmptyYears'])) {
                 $yearFilled = [];
                 $min = $yearArray[0]['title'];
                 // round the starting decade down to zero for equal rows
-                $min = substr_replace($min,"0",-1);
+                $min = substr_replace($min, "0", -1);
                 $max = $yearArray[count($yearArray) - 1]['title'];
                 // if we have an actual documentId it should be used, otherwise leave empty
-                for($i = 0; $i < $max-$min+1; $i++) {
-                    $key = array_search($min+$i,array_column($yearArray, 'title'));
-                    if (is_int($key))
-                    {
+                for ($i = 0; $i < $max - $min + 1; $i++) {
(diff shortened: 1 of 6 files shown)


michaelkubina:
unfilled
filled

Classes/Controller/CalendarController.php
```diff
@@ -281,6 +281,27 @@ public function yearsAction()
                     'title' => $year['title']
                 ];
             }
+            // create an array that includes years without issues
+            if($this->settings['showEmptyYears']){
```
sebastian-meyer: We need to check configuration options with `empty()`, because when updating from an older version, new options don't always get set accordingly, so `$this->settings['showEmptyYears']` may be unset. `empty()` doesn't issue a warning when given an unset variable.

```suggestion
            if (!empty($this->settings['showEmptyYears'])){
```

Resources/Private/Language/de.locallang_be.xlf
```diff
@@ -121,6 +121,10 @@
 				<source><![CDATA[Initial document to show (e.g. of type "newspaper" or "ephemera")]]></source>
 				<target><![CDATA[Anfangsdokument (z.B. vom Typ "newspaper" oder "ephemera)]]></target>
 			</trans-unit>
+			<trans-unit id="plugins.calendar.flexform.showEmptyYears" approved="yes">
+				<source><![CDATA[Show months without issues in calendar?]]></source>
```
sebastian-meyer:
```suggestion
				<source><![CDATA[Show years without issues in calendar?]]></source>
```

Configuration/Flexforms/Calendar.xml
```diff
@@ -60,6 +60,16 @@
                             </config>
                         </TCEforms>
                     </settings.initialDocument>
+                    <settings.showEmptyYears>
+                      <TCEforms>
+                        <exclude>1</exclude>
+                        <label>LLL:EXT:dlf/Resources/Private/Language/locallang_be.xlf:plugins.calendar.flexform.showEmptyYears</label>
+                        <config>
+                          <type>check</type>
+                          <default>1</default>
```
sebastian-meyer: If possible, when introducing new options the default should always be the old behavior in order to make updates easier. So this should be `0`.

```suggestion
                          <default>0</default>
```

Documentation/Plugins/Index.rst
```diff
@@ -219,6 +219,13 @@ now.
        :ref:`t3tsref:data-type-integer`
    :Default:
 
+ - :Property:
+       showEmptyYears
+   :Data Type:
+       :ref:`t3tsref:data-type-boolean`
+   :Default:
+       1
```
sebastian-meyer: See above.

sebastian-meyer (changes requested):
Please see my comments and also have a look at the inline remarks from Codacy.

Classes/Controller/CalendarController.php
michaelkubina: Thank you for the explaination, will keep it in mind next time.

michaelkubina:
Have applied the requested changes, thank you for your feedback!

sebastian-meyer (approved):
Thank you very much!

PR #905, opened 2023-02-07
[FEATURE] Add an option to preselect fulltext search as default

michaelkubina:
This PR adds an option to the search plugin, that allows for the fulltextsearch being preselected instead of the default search in metadata. If selected it gets applied as long as we dont have a `tx_dlf_listview[searchParameter][fulltext]` parameter, which can happen either through the landing page or when entering a collection resultset. As soon as we have any real user defined search (`tx_dlf_listview[searchParameter][fulltext]` being then `0 or 1`), the user preference will be used instead.

### DIFF
--- a/Configuration/FlexForms/Search.xml
+++ b/Configuration/FlexForms/Search.xml
@@ -41,6 +41,16 @@
                             </config>
                         </TCEforms>
                     </settings.fulltext>
+                    <settings.fulltextPreselect>
+                        <TCEforms>
+                            <exclude>1</exclude>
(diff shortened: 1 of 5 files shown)


sebastian-meyer (dismissed):
This looks fine. Could you please rebase to master and resolve the merge conflict?

michaelkubina:
Rebase done and the conflict resolved itself => false negative due to folder-renaming commit.

sebastian-meyer (approved):

PR #907, opened 2023-02-08
[MAINTENANCE] Fix OaiPmh functional tests

beatrycze-volk:
Remove slashes from search URLs - reason for failing tests:
```
Phpoaipmh\Exception\HttpException: Client error: `GET http://web:8000/public/typo3temp/var/tests/functional-3567482//index.php?id=20001&verb=Identify` resulted in a `404 Not Found` response
```

### DIFF
--- a/Tests/Functional/Api/OaiPmhTest.php
+++ b/Tests/Functional/Api/OaiPmhTest.php
@@ -37,8 +37,8 @@ class OaiPmhTest extends FunctionalTestCase
     {
         parent::setUp();
 
-        $this->oaiUrl = $this->baseUrl . '/index.php?id=' . $this->oaiPage;
-        $this->oaiUrlNoStoragePid = $this->baseUrl . '/index.php?id=' . $this->oaiPageNoStoragePid;
+        $this->oaiUrl = $this->baseUrl . 'index.php?id=' . $this->oaiPage;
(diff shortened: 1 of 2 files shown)


sebastian-meyer:
Hm, all the URLs look good, now. Some of them are missing the `index.php` part, but that shouldn't be an issue. Do you have any idea why the tests still fail?

beatrycze-volk:
> Hm, all the URLs look good, now. Some of them are missing the `index.php` part, but that shouldn't be an issue. Do you have any idea why the tests still fail?

I'm still searching the reason, but I'm starting to suspecting that somehow there is also a problem inside the preparation of the testing environment. I'm on checking out our code for inserting documents into the SOLR.

sebastian-meyer:
Nice try! So it's a change in TYPO3 which breaks the tests...

beatrycze-volk:
> Nice try! So it's a change in TYPO3 which breaks the tests...

Yes, but I'm still not able to pinpoint the place where the problem appears.

sebastian-meyer:
Those are the only two commits since TYPO3 10.4.34 which IMHO could've caused our problem:

- https://github.com/typo3/typo3/commit/fa9832df37
- https://github.com/typo3/typo3/commit/5e1585ea79

sebastian-meyer (approved):
I have no idea why, but it works. Good job! ;o)

PR #912, opened 2023-02-13
[MAINTENANCE] Add 'SameSite' attribute to cookies

csidirop:
Under the Firefox browser, it has been noted for some time that cookies without SameSite attributes will soon no longer be supported and will instead be treated as "lax".

This is the message for the `tx-dlf-pageview-fulltext-select` cookie: 
```
Das Cookie "tx-dlf-pageview-fulltext-select" verfügt über keinen gültigen Wert für das "SameSite"-Attribut. Bald werden Cookies ohne das "SameSite"-Attribut oder mit einem ungültigen Wert dafür als "Lax" behandelt. Dadurch wird das Cookie nicht länger an Kontexte gesendet, die zu einem Drittanbieter gehören. 
Falls Ihre Anwendung das Cookie in diesen Kontexten benötigt, fügen Sie bitte das Attribut "SameSite=None" zu ihm hinzu. Weitere Informationen zum "SameSite"-Attribut finden Sie unter https://developer.mozilla.org/docs/Web/HTTP/Headers/Set-Cookie/SameSite.
```

This PR adds another parameter to the [setCookie function](https://github.com/kitodo/kitodo-presentation/blob/master/Resources/Public/JavaScript/PageView/Utility.js#L583) which can set the SameSite attribute individually.

### DIFF
--- a/Resources/Public/JavaScript/PageView/Utility.js
+++ b/Resources/Public/JavaScript/PageView/Utility.js
@@ -579,10 +579,22 @@ dlfUtils.parseDataDic = function (element) {
  *
  * @param {string} name The key of the value
  * @param {?} value The value to save
+ * @param {string} samesite Sets the SameSite attribute: lax, strict or none
+ * 
  */
-dlfUtils.setCookie = function (name, value) {
-
-    document.cookie = name + "=" + decodeURI(value) + "; path=/";
+dlfUtils.setCookie = function (name, value, samesite) {
+    switch(samesite) {
+        case "lax":
+        case "strict": 
+            break;
+        case "none":
+            samesite+= ";secure"
+            break;
+        default:
+            samesite = "lax"
+    }
+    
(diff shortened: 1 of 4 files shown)


Resources/Public/JavaScript/PageView/Utility.js
```diff
@@ -579,10 +579,22 @@ dlfUtils.parseDataDic = function (element) {
...
+        case "strict": 
+            break;
+        case "none":
+            samesite+= ";secure"
+            break;
+        default:
+            samesite = "lax"
+    }
+    
+    document.cookie = name + "=" + decodeURI(value) + "; path=/" + "; SameSite="+samesite;
```
sebastian-meyer:
```suggestion
    document.cookie = name + "=" + decodeURI(value) + "; path=/" + "; SameSite=" + samesite;
```

sebastian-meyer (approved):

Issue #918, opened 2023-02-27
[Question] How can I display the name of the current collection in a listview page?


BFallert:
When a collection is clicked on a page that contains a collection list (plugin: Collection), a normal ListView is called.

Currently, however, it is not visible in this ListView page that it was called via the link of a collection.
In the contained Typo3 variables it can be recognized via `lastSearch[collection]` that the template was called via a collection link, but instead of the collection name only the collection number is contained (i.e. the ID in the Typo3 database).

Is there a possibility to output the data of the collection list, especially **display label, description, thumbnail,** also in the ListView of a collection?

michaelkubina:
The Listview-Controller provides such information via the `currentCollection` object. In the code it is even assigned to the view, but from what i see it does not get applied anywhere in the template ([\Resources\Private\Templates\ListView\Main.html](https://github.com/kitodo/kitodo-presentation/blob/master/Resources/Private/Templates/ListView/Main.html)) by default.

It would be required add something like the snippet shown below to the template - but with a lot collections this information block might grow huge. And of course it needs to be styled in your CSS somehow.

I can only speak for certain that it works for the current master branch (not the last release). The current master has already the changes to the collection-plugin that i wrote. This being said i am not sure if the last release already returned a list of collections or just a single item. There were several severe bugs, that made collections not reliable or functional back then.

**\Resources\Private\Templates\ListView\Main.html**
```
...
            <f:for each="{currentCollection}" as="currentCollectionItem">
                <div class="collection">
                    <div class="collection-label">
                        <f:if condition="{currentCollectionItem.thumbnail}">
                            <f:image image="{currentCollectionItem.thumbnail}" />
                        </f:if>
                    <div>
                    <div class="collection-thumbnail">
                        <f:if condition="{currentCollectionItem.label}">
                            {currentCollectionItem.label}
                        </f:if>
                    </div>
                    <div class="collection-description">
                        <f:if condition="{currentCollectionItem.description}">
                            {currentCollectionItem.description}
                        </f:if>
                    </div>
                </div>
            </f:for>
...
```

BFallert:
I am working with the 4.x branch, but `currentCollection` does not exist.
`currentCollection` is not found anywhere in the source code? 
Main-Branch and 4.x do not differ significantly.
Have your changes been incorporated elsewhere?
Am I looking in the wrong place?

michaelkubina:
[...]

PR #920, opened 2023-03-01
ListView: add called collection

BFallert:
idea from @michaelkubina
see: https://github.com/kitodo/kitodo-presentation/issues/918

When a collection is called in a collection list (Kitodo:Collection), a ListView is called. In this ListView it is desirable to be able to output the name of the selected collection (Headline) and its associated data (Thumbnail and Description).

With these adjustments this is possible without further configuration.

### DIFF
--- a/Classes/Controller/ListViewController.php
+++ b/Classes/Controller/ListViewController.php
@@ -107,5 +107,6 @@ class ListViewController extends AbstractController
 
         $this->view->assign('sortableMetadata', $sortableMetadata);
         $this->view->assign('listedMetadata', $listedMetadata);
+        $this->view->assign('currentCollection', $collection);
     }
 }
--- a/Resources/Private/Templates/ListView/Main.html
+++ b/Resources/Private/Templates/ListView/Main.html
@@ -13,15 +13,40 @@
       data-namespace-typo3-fluid="true">
 
     <div class="tx-dlf-listview">
+        <f:if condition="{lastSearch.collection}">
+            <f:then>
+                <f:for each="{currentCollection}" as="currentCollectionItem">
+                    <div class="collection">
+                        <f:if condition="{currentCollectionItem.label}">
+                            <h2>{currentCollectionItem.label}</h2>
+                        </f:if>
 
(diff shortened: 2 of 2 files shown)


BFallert:
Was not sure which branch to take master or 4.x.

csidirop:
I think all PRs going to master, and master is merged at regular intervals into 4.x

sebastian-meyer:
Please have a look at [Resources/Private/Templates/Collection/List.html](https://github.com/kitodo/kitodo-presentation/blob/master/Resources/Private/Templates/Collection/List.html). This template already contains your proposed additions and is meant to be used with the `Collection` controller.

I am OK with unifying the templates and using only one template for all cases, but then we need to remove the duplicate template and make sure the unified template is always used.

sebastian-meyer:
> I think all PRs going to master, and master is merged at regular intervals into 4.x

Yes, that's true until development of a new major version begins. Currently I am working on the 4.1 release, so all pull requests should still target the `master` branch. After that the next version will be 5.0 (with support for TYPO3 v11) and any upcoming changes to 4.1 should then be targeted at the `4.x` branch.

sebastian-meyer:
How should we proceed with this pull request, @BFallert?

BFallert:
> Please have a look at [Resources/Private/Templates/Collection/List.html](https://github.com/kitodo/kitodo-presentation/blob/master/Resources/Private/Templates/Collection/List.html). This template already contains your proposed additions and is meant to be used with the Collection controller.

But when i clicked on a collection on a page that contains a collection list (plugin: Collection), a normal ListView is called.

> I am OK with unifying the templates and using only one template for all cases, but then we need to remove the duplicate template and make sure the unified template is always used.

Small adjustments would probably still have to be made in the existing template. 
But I don't know how I can call up the same template from two different situations (collection and listView)?
Can anyone offer support here? 

On the other hand, if I'm the only one with this problem, we can close the PR.

sebastian-meyer:
> On the other hand, if I'm the only one with this problem, we can close the PR.

PR #924, opened 2023-03-07
[BUG] Fix use a scalar value as an array

beatrycze-volk:
Fix for the error:

`PHP Warning: Cannot use a scalar value as an array in /var/www/extensions/kitodo-presentation/Classes/Controller/MetadataController.php line 265`

### DIFF
--- a/Classes/Controller/MetadataController.php
+++ b/Classes/Controller/MetadataController.php
@@ -261,8 +261,6 @@ class MetadataController extends AbstractController
                         foreach ($metadata[$i][$name] as &$langValue) {
                             $langValue = Helper::getLanguageName($langValue);
                         }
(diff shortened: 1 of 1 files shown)


sebastian-meyer (approved):
Although I am very confused how this was ever meant to work in the first place, the removal is obviously fine!
(There are more strange statements in this file... Guess we should clean up the whole class sometime!)

beatrycze-volk:
> Although I am very confused how this was ever meant to work in the first place, the removal is obviously fine!

I have actually traced the changes to this line of the code. It seems in the past this value was sanitized and then sanitation was removed but assignment of the value to itself was left.

> (There are more strange statements in this file... Guess we should clean up the whole class sometime!)

Right.

PR #926, opened 2023-03-15
[BUG] Fix messages for New Tenant module

sebastian-meyer:
This fixes some bugs in the New Tenant BE module:
- Added missing translations for message about missing namespace configuration (`newTenant.formatNotOkay` and `newTenant.formatNotOkayMsg`)
- Removed old translation keys which were now longer in use (`newTenant.formatCreated` and `newTenant.formatCreatedMsg`)
- Changed disabled submit form field for metadata to a button to allow its label to be visible


christopher-timm (approved):
Looks good from my point of view

PR #927, opened 2023-03-15
[BUG] Make Solr core name mandatory

sebastian-meyer:
If the creation of a new Solr core fails in New Tenant BE module or via list module, there is still a database record created and no error message shown. This leads to the false impression that everything is fine.
This pull request makes the core name field mandatory, thereby preventing the creation of a database record if Solr core creation failed. This results in the list module and the New Tenant BE module no longer falsly reporting the existence of a working Solr core.


christopher-timm (commented):
In my local environment, I still have the case that a database entry is created, although no core could be created. The index_name field remains empty

sebastian-meyer:
You have to deactivate and then re-activate the extension in order for the TCA configuration to be updated.

But I just encountered another problem with my solution: Although it works for me when using the "New Tenant" module, creating Solr cores manually via the list module doesn't work anymore (because you can't save without giving a core name...).

I'll rework this!

sebastian-meyer:
I've now decided on another approach:
- If creation of a new Solr core fails the field array is cleared in datamap post-processing in order to prevent creating a faulty database record.
- I've added the `readOnly` flag to prevent changing the Solr core name of existing records, because that would almost always lead to failure. Since this flag only affects field rendering in the backend, it's still possible to change the core name in database or through extension methods.
- I've removed the `required` flag from TCA again to allow creating new Solr cores by simply creating new backend records via the list module. (This just reverts my earlier changes.)

christopher-timm:
> You have to deactivate and then re-activate the extension in order for the TCA configuration to be updated.

This is what I have made, but I'm not sure if I might be making a mistake.
If I remove the existing SolrCore record from my system folder and then create a new Solr core via the NewTenant module, the record is also created accordingly without any error message. However, the index_name field remains empty.

christopher-timm:
While debugging, I saw that an empty string is returned to the NewTenantController on line 282 via "createCore()". This empty string is then written to the database in $newRecord via add. The hook doesn't seem to be executed at all because I don't get any debug output there.

sebastian-meyer:
Thanks for debugging! I have no idea why Extbase doesn't use the TYPO3 DataHandler for database processing. Seems like we need to use Extbase Validators instead to ensure all the fields meet their requirements. But there is no unified way to implement Validators for TYPO3v9 and TYPO3v10, so for the time being I just added another check for an empty core name in the New Tenant controller.

christopher-timm (approved):
Works from my point of view!

PR #885, opened 2022-12-28
[MAINTENANCE] TYPO3 v11 compatibility

christopher-timm:
This PR ensures that the DLF Extension is also compatible with version 11 of Typo3.
The biggest adjustment was the reorganization of the pagination, since the widgets were removed from Typo3. This required an own "Pagniator" for the Solr, as well as for the display of the "PageGrid"

PR #930, opened 2023-03-24
[BUG] Fix pagenavigation on sorted resultsets

michaelkubina:
This PR is a suggested fix for the issue related to failed pagenavigation in sorted lists. The underlying reason is, that from what i see, all searches require paramters via GET, while the sorting-form does POST. Thus the parameters go missing and an empty search will be returned, after using the page-navigation. Also the sorting-form does not know of the filterqueries applied by the facets. Those need to be applied to the hidden form as well.

This PR would fix the current issue, but its a rather quick fix.

(I have closed the prior PR, because i branched off from a wrong branch with unrelated commits. This is the actual one)

### DIFF
--- a/Resources/Private/Partials/ListView/SortingForm.html
+++ b/Resources/Private/Partials/ListView/SortingForm.html
@@ -26,7 +26,7 @@
             </p>
         </f:if>
 
-        <f:form section="showResults" action="{action}" controller="{Controller}" name="searchParameter" method="post" class="tx-dlf-search-form">
+        <f:form section="showResults" action="search" controller="Search" name="searchParameter" method="get" class="tx-dlf-search-form">
             <div>
                 <label for="form-orderBy-{viewData.uniqueId}">
                     <f:translate key="listview.form.orderBy"/>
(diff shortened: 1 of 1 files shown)


sebastian-meyer:
I think this should be fixed in the [SearchController](https://github.com/kitodo/kitodo-presentation/blob/master/Classes/Controller/SearchController.php), not in the Fluid template. There already is a method for getting the POST parameters, which is even called twice (lines 77 & 100), but for some reason it seems not to do what it is supposed to.

Adding the missing parameter for query filters to the template is fine, though!

michaelkubina:
The problem i see, is that the pagination-widget triggers a GET-request (**always** since typo 10.2, see this issue: https://forge.typo3.org/issues/90003 ), so we cant change the method for the widget itself anymore.

The controller extract the POST-parameters properly, but the pagination-link does not use them anymore. Because of this i did change the method in the sorting-form-template, in order to preserve the URL-Parameters.

Sadly i have no idea, how to make it work in the controller to reapply the POST-parameters to the URL...but this is essentially the reason for it not to work.

michaelkubina:
In a talk with Chris (@chrizzor ) i was told, that with Typo3 11 the pagination-widget was removed for a pagination-api instead. So this fix would only work for a limited time (see: https://github.com/kitodo/kitodo-presentation/pull/885) and would be obsolete with Typo3 11. Just as an additional note...how should we proceed?

sebastian-meyer (approved):

sebastian-meyer:
OK, I'll merge this for 4.x.

PR #937, opened 2023-04-03
[SECURITY] Fix potential XSS vulnerability

sebastian-meyer:
There is a potential security issue in Kitodo.Presentation allowing attackers to forge XSS URLs using the `tx_dlf['page']` parameter. In order to be vulnerable to this attack the instance must not use TYPO3 routing and the Fluid template of the `Navigation` plugin must fail to properly escape the `page` value of the request data.

This pull request adds proper sanitizing for `tx_dlf['page']` and moves sanitation of `tx_dlf[id]` to plugin initialization (the latter was already sanitized before, but at a later point).

### DIFF
--- a/Classes/Controller/AbstractController.php
+++ b/Classes/Controller/AbstractController.php
@@ -89,28 +89,7 @@ abstract class AbstractController extends \TYPO3\CMS\Extbase\Mvc\Controller\Acti
         $this->requestData = GeneralUtility::_GPmerged('tx_dlf');
 
         // Sanitize user input to prevent XSS attacks.
-
-        // tx_dlf[id] may only be an UID or URI.
-        if (
-            !empty($this->requestData['id'])
-            && !MathUtility::canBeInterpretedAsInteger($this->requestData['id'])
-            && !GeneralUtility::isValidUrl($this->requestData['id'])
-        ) {
-            $this->logger->error('Invalid ID or URI "' . $this->requestData['id'] . '" for document loading');
-            unset($this->requestData['id']);
-        }
-
-        // tx_dlf[page] may only be a positive integer or valid XML ID.
-        if (
-            !empty($this->requestData['page'])
-            && !MathUtility::canBeInterpretedAsInteger($this->requestData['page'])
-            && !Helper::isValidXmlId($this->requestData['page'])
-        ) {
-            $this->requestData['page'] = 1;
-        }
-
-        // tx_dlf[double] may only be 0 or 1.
-        $this->requestData['double'] = MathUtility::forceIntegerInRange($this->requestData['double'], 0, 1, 0);
+        $this->sanitizeRequestData();
 
         // Get extension configuration.
         $this->extConf = GeneralUtility::makeInstance(ExtensionConfiguration::class)->get('dlf');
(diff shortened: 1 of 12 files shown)


Classes/Controller/AbstractController.php
```diff
@@ -88,9 +87,28 @@ public function injectDocumentRepository(DocumentRepository $documentRepository)
...
+            && !GeneralUtility::isValidUrl($this->requestData['id'])
+        ) {
+            unset($this->requestData['id']);
+            $this->logger->error('Invalid ID or URI "' . $this->requestData['id'] . '" for document loading');
+        }
+
+        // tx_dlf[page] may only be a positive integer or valid XML ID.
+        if (
+            empty($this->requestData['page'])
+            || !preg_match('/^[1-9][0-9]*$|^[_a-z][_a-z0-9-.]*$/', $this->requestData['page']) === 1
```
ByteParty: Why do you use the MathUtility::canBeInterpretedAsInteger above and here is a regex for the integer check?

sebastian-meyer: Because I tried to incorporate both checks (for integer and valid XML IDs) into one statement. Otherwise I'd have to nest multiple `||` and `&&` operations into this condition check, which would be even more confusing, I guess.

beatrycze-volk: So maybe move the checks to own method to avoid long conditions in if statement?

Classes/Controller/AbstractController.php
```diff
@@ -108,28 +126,33 @@ protected function initialize()
...
-        if (!empty($requestData['id'])) {
+        if (!empty($documentId)) {
 
             $doc = null;
 
-            if (MathUtility::canBeInterpretedAsInteger($requestData['id'])) {
+            if (MathUtility::canBeInterpretedAsInteger($documentId)) {
                 // find document from repository by uid
-                $this->document = $this->documentRepository->findOneByIdAndSettings((int) $requestData['id']);
+                $this->document = $this->documentRepository->findOneByIdAndSettings((int) $documentId);
```
beatrycze-volk: Do you still need parse it here to int?

In method paramater it is already passed as int.

```suggestion
                $this->document = $this->documentRepository->findOneByIdAndSettings($documentId);
```

Classes/Controller/AbstractController.php
sebastian-meyer: You are right. I hope it's easier to read now.

beatrycze-volk: Actually maybe it would be good just put the sanitation of the input in the own function:

```
private function sanitizeUserInput() {
    // Sanitize user input to prevent XSS attacks.

        // tx_dlf[id] may only be an UID or URI.
        if (
            !empty($this->requestData['id'])
            && !MathUtility::canBeInterpretedAsInteger($this->requestData['id'])
            && !GeneralUtility::isValidUrl($this->requestData['id'])
        ) {
            $this->logger->error('Invalid ID or URI "' . $this->requestData['id'] . '" for document loading');
            unset($this->requestData['id']);
        }

        // tx_dlf[page] may only be a positive integer or valid XML ID.
        if (
            !empty($this->requestData['page'])
            && !MathUtility::canBeInterpretedAsInteger($this->requestData['page'])
            && !Helper::isValidXmlId($this->requestData['page'])
        ) {
            $this->requestData['page'] = 1;
        }
}
```

Also logger for $this->requestData['id'] needs to be called before unsetting the value.

sebastian-meyer: > Also logger for $this->requestData['id'] needs to be called before unsetting the value.

Uh, this is embarrassing... ;o)

sebastian-meyer: OK, I moved the input sanitation to a separate method.

beatrycze-volk: Great! The code is ready merge :)

beatrycze-volk (approved):

PR #940, opened 2023-04-04
[MAINTENANCE] Move setPage() to AbstractController

beatrycze-volk:
Avoid code redundancy

Depends on #939


sebastian-meyer (approved):

sebastian-meyer (changes requested):
Tests fail because in `ToolboxController` a private method `setPage` is declared although the parent class now has a protected method of the same name. I think the method in `ToolboxController` is redundant and can be removed.

sebastian-meyer (approved):

PR #941, opened 2023-04-05
[MAINTENANCE] Change way for rendering tools in the Toolbox

beatrycze-volk:


beatrycze-volk:
Codacy checks are false positives. The functions are used inside the renderToolByName() function.

sebastian-meyer (approved):

PR #953, opened 2023-04-28
[MAINTENANCE] Add tests for Common package

### LINKED ISSUE #781: [FUND] Add tests and significantly improve test coverage

oliver-stoehr:


Tests/Fixtures/Common/solrcores.xml
sebastian-meyer, line 1: This file is superseded by `Tests/Fixtures/Common/solrcores.csv`, isn't it? Then it should be removed from the PR.

sebastian-meyer (changes requested):

Tests/Functional/Common/SolrSearchTest.php
```diff
@@ -0,0 +1,106 @@
...
+        $this->setUpData(self::$databaseFixtures);
+        $this->solr = $this->setUpSolr(4, 0, self::$solrFixtures);
+    }
+
+    /**
+     * @test
+     */
+    public function canPrepareAndSubmit()
+    {
+        $this->markTestSkipped('Does not work in combination with other tests.');
```
sebastian-meyer: Hm, a skipped test is quite useless... There should be a way to make this work even in combination with the other tests. Where exactly does the test fail?

sebastian-meyer (changes requested):

Tests/Fixtures/Common/solrcores.xml
oliver-stoehr: Thank you for pointing out, I removed this file.

Tests/Functional/Common/SolrSearchTest.php
oliver-stoehr: I reenabled the test. It fails on line 55 at the first `assert` using the `SolrSearch` object from line 53. I could not find any difference in any variables accesses in the tested `SolrSearch::prepare` method.
I also switched to a different Solr core, but that didn't change the result.
However I'm not sure about the `DocumentRepository` passed to the constructor of the `SolrSearch` – I'm don't fully understand how the `DocumentRepository` object is created (or maybe reused?).

sebastian-meyer:
I fixed the PHPStan issue in `master`. It should disappear as soon as you update your branch.

Tests/Functional/Common/SolrSearchTest.php
sebastian-meyer: I am sorry, but I can't help you with that. Maybe @beatrycze-volk can give more insight into using the `DocumentRepository` correctly?

beatrycze-volk: > I am sorry, but I can't help you with that. Maybe @beatrycze-volk can give more insight into using the `DocumentRepository` correctly?

Actually more suitable person to answer this question would be @chrizzor as he has implemented this class.

oliver-stoehr:
I fixed the test. It looks like some database query were cached, that were necessary to determine the correct solr core. The solr cores are now determined more directly.

sebastian-meyer (approved):
Great job! Thank you very much!

sebastian-meyer:
Now you only need to update your branch to master and then I can merge it!

PR #954, opened 2023-04-28
[MAINTENANCE] Add tests for Controller package

### LINKED ISSUE #781: [FUND] Add tests and significantly improve test coverage

oliver-stoehr:
~Tests for CalendarController will be added in a separate pull request.~
Update: Added tests for CalendarController in another commit.


sebastian-meyer:
I fixed the PHPStan issue in `master`. It should disappear as soon as you update your branch.

composer.json
```diff
@@ -43,6 +43,7 @@
     "symfony/process": "^5.4|^6.4"
   },
   "require-dev": {
+	  "fluidtypo3/vhs": "^6.1",
```
sebastian-meyer: Das müsste 7.0 sein, damit TYPO3 v11 und v12 unterstützt werden.

```suggestion
    "fluidtypo3/vhs": "^7.0",
```

sebastian-meyer (changes requested):

Tests/Functional/Controller/AbstractControllerTest.php
```diff
@@ -0,0 +1,95 @@
...
+    public $currentSolrUid = 1;
+
+    public $currentCoreName = '';
+
+    protected function setUpData($databaseFixtures): void
+    {
+        foreach ($databaseFixtures as $filePath) {
+            $this->importCSVDataSet($filePath);
+        }
+        $this->persistenceManager = $this->objectManager->get(PersistenceManager::class);
```
sebastian-meyer: This seems to be the problem why the tests fail. `$this->objectManager->get()` throws the error `Call to a member function get() on null`, so `$this->objectManager` isn't initialized.

christopher-timm: This fails because as of version 12 the GeneralUtilityManager is used to get an instance of a class. However, if I change this, all tests fail. The response only contains an empty string. This must be related to the changes to version 12.
`$this->persistenceManager = GeneralUtility::makeInstance(PersistenceManager::class);`

sebastian-meyer: Yes, this fixed the issue. Now the tests are running for both, v11 and v12. Unfortunately, some of the tests fail...

stweil:
More tests would be really useful. Therefore I appreciate this pull request.

@oliver-stoehr, are you still working on it?

I suggest to squash some of the commits (each commit should be complete and produce a working software). A rebase on main would be helpful, too.

christopher-timm:
Tests created for:
- Classes/Controller/AbstractController.php (Abstract, therefore tested through the other, implementing controllers)
- Classes/Controller/AudioPlayerController.php
- Classes/Controller/CalendarController.php
- Classes/Controller/CollectionController.php
- Classes/Controller/FeedsController.php
- Classes/Controller/ListViewController.php
- Classes/Controller/MetadataController.php
- Classes/Controller/NavigationController.php
- Classes/Controller/PageGridController.php
- Classes/Controller/PageViewController.php
- Classes/Controller/SearchController.php
- Classes/Controller/StatisticsController.php
- Classes/Controller/TableOfContentsController.php
- Classes/Controller/ToolboxController.php

In the StatisticsController in the MainAction, the test has so far failed because there was an error in the action. The remaining tests have otherwise passed. However, the problem now is that the new Typo3 version has changed the handling of the response objects. The interface `\Psr\Http\Message\ServerRequestInterface` is now used. This means that calling the actions in the PR is no longer compatible with the current presentation code. The attempt to rebuild the controllers according to the mentioned interface unfortunately did not work with reasonable effort. In principle, the tests should work if the problem with the request/response objects is resolved.

sebastian-meyer:
Successor: #1708

stweil:
> Successor: #1708

So can this pull request be closed?

PR #957, opened 2023-05-04
[BUGFIX] Display volumes correctly in ToC plugin

beatrycze-volk:
Depends on #956

This works but all in all it is a bit confusing.

I would propose that fields `mets_orderlabel`, `mets_order` and `volume_sorting` in database are getting consolidated into one field called `order`. Then we should add general method to keep ToC arrays sorted by this field. Is there any difference between those 3 fields? Or they usage depends on the document type?

In general we are using now for title 3 different fields: `title`, `label` or `orderlabel`. And they have also two forms of naming `label` or `mets_label`, the same for `orderlabel`. Here should be also considered the naming of this fields while querying data because it is also quiet confusing. 

Example:
```
$entry = [
    'label' => !empty($resArray['mets_label']) ? $resArray['mets_label'] : $resArray['title'],
    'orderlabel' => $resArray['mets_orderlabel'],
```
then it is used here
```
$entryArray['title'] = !empty($entry['label']) ? $entry['label'] : $entry['orderlabel'];
```

### DIFF
--- a/Classes/Command/BaseCommand.php
+++ b/Classes/Command/BaseCommand.php
@@ -196,7 +196,7 @@ class BaseCommand extends Command
         $document->setAuthor(implode('; ', $metadata['author']));
         $document->setThumbnail($doc->thumbnail ? : '');
         $document->setMetsLabel($metadata['mets_label'][0] ? : '');
-        $document->setMetsOrderlabel($metadata['mets_orderlabel'][0] ? : '');
+        $document->setMetsOrderlabel($metadata['mets_orderlabel'][0] ? : $metadata['mets_order'][0] ? : '');
 
         $structure = $this->structureRepository->findOneByIndexName($metadata['type'][0], 'tx_dlf_structures');
         $document->setStructure($structure);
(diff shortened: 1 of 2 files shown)


Classes/Command/BaseCommand.php
```diff
@@ -196,7 +194,7 @@ protected function saveToDatabase(Document $document)
         $document->setAuthor(implode('; ', $metadata['author']));
         $document->setThumbnail($doc->thumbnail ? : '');
         $document->setMetsLabel($metadata['mets_label'][0] ? : '');
-        $document->setMetsOrderlabel($metadata['mets_orderlabel'][0] ? : '');
+        $document->setMetsOrderlabel($metadata['mets_orderlabel'][0] ? : $metadata['mets_order'][0] ? : '');
```
sebastian-meyer, line 197: The idea was to keep the document instance as close to the METS standard as possible in order to allow for future applications where there is a difference between having an `ORDERLABEL` or it being empty/omitted. If you need to have a fallback (i. e. `ORDER`) then this should be done when applying the property in the respective context.

But you are right, maybe we should consolidate this and just use a single field for sorting.

Nevertheless, since the change in this line seems to be unnecessary to achieve the goal of fixing the display of volumes, I'd rather keep it unchanged for now until we do a complete refactoring of the sorting mechanisms (like you suggested). Otherwise this could confuse people even more.

```suggestion
        $document->setMetsOrderlabel($metadata['mets_orderlabel'][0] ? : '');
```

sebastian-meyer (changes requested):
See my comment. Other than that this seems fine!

Classes/Command/BaseCommand.php
sebastian-meyer: Nevermind, I just saw that the document object doesn't have a property which holds the `ORDER` value so using this as a fallback wouldn't be that easy.

So, please ignore my comment! I'll approve the pull request and merge it!

sebastian-meyer (approved):

PR #959, opened 2023-05-05
[MAINTENANCE] Change initialization of repositories in BaseCommand

beatrycze-volk:


sebastian-meyer:
Does this still work in TYPO3v9? If not I'll merge it while adding TYPO3v11 support (see #885).

beatrycze-volk:
> Does this still work in TYPO3v9? If not I'll merge it while adding TYPO3v11 support (see #885).

I didn't test it for TYPO3v9 so it is better to add with TYPO3v11 support

beatrycze-volk:
Already implemented in other PR

PR #961, opened 2023-05-05
[MAINTENANCE] Initialize metadata array with empty values in abstract Doc class

beatrycze-volk:


sebastian-meyer (approved):

sebastian-meyer:
Just an idea: Would it be feasible to build get the metadata fields dynamically from configuration (https://github.com/kitodo/kitodo-presentation/blob/master/Configuration/TCA/tx_dlf_documents.php) in order to not forget about new metadata fields added in the future?

beatrycze-volk:
> Just an idea: Would it be feasible to build get the metadata fields dynamically from configuration (https://github.com/kitodo/kitodo-presentation/blob/master/Configuration/TCA/tx_dlf_documents.php) in order to not forget about new metadata fields added in the future?

I don't think that's possible. The included file contains also fields which are not included in metadata table. We would probably need to somehow mark in this file which fields should be initialized.

PR #964, opened 2023-05-16
[BUGFIX] Display issues correctly in ToC plugin

beatrycze-volk:
The issues were not displayed correctly because their METS structure doesn't contain `LABEL` nor `ORDERLABEL`:

mets-issue

It is fixed by taking `LABEL` from `ORDERLABEL` of day (parent) structure:

issues-fixed

Probably more suitable solution would be to have METS which already contains `LABEL` or `ORDERLABEL` for issue type.

Depends on #957

### DIFF
--- a/Classes/Common/MetsDocument.php
+++ b/Classes/Common/MetsDocument.php
@@ -333,7 +333,7 @@ final class MetsDocument extends Doc
         $details['orderlabel'] = (isset($attributes['ORDERLABEL']) ? $attributes['ORDERLABEL'] : '');
         $details['contentIds'] = (isset($attributes['CONTENTIDS']) ? $attributes['CONTENTIDS'] : '');
         $details['volume'] = '';
-        // Set volume information only if no label is set and this is the toplevel structure element.
+        // Set volume any year information only if no label is set and this is the toplevel structure element.
         if (
             empty($details['label'])
             && $details['id'] == $this->_getToplevelId()
@@ -341,6 +341,9 @@ final class MetsDocument extends Doc
             $metadata = $this->getMetadata($details['id']);
             if (!empty($metadata['volume'][0])) {
                 $details['volume'] = $metadata['volume'][0];
+            }
+            if (!empty($metadata['year'][0])) {
+                $details['year'] = $metadata['year'][0];
             }
         }
         $details['pagination'] = '';
--- a/Classes/Common/MetsDocument.php
+++ b/Classes/Common/MetsDocument.php
@@ -181,8 +181,10 @@ final class MetsDocument extends Doc
         $details = $this->getLogicalStructure($id);
         if (!empty($details)) {
             $metadata['mets_order'][0] = $details['order'];
-            $metadata['mets_label'][0] = $details['label'];
-            $metadata['mets_orderlabel'][0] = $details['orderlabel'];
+            if ($metadata['type'][0] != 'issue') {
+                $metadata['mets_label'][0] = $details['label'];
+                $metadata['mets_orderlabel'][0] = $details['orderlabel'];
+            }
         }
     }
 
@@ -630,6 +632,14 @@ final class MetsDocument extends Doc
         if (empty($metadata['date'][0])) {
             $metadata['date'][0] = '';
         }
+        // Set mets_label for issues
+        if ($metadata['type'][0] == 'issue' && empty($metadata['mets_label'][0])) {
+            $dayLabel = $this->mets->xpath('./mets:structMap[@TYPE="LOGICAL"]//mets:div[@TYPE="day"]/@ORDERLABEL');
+            if (!empty($dayLabel)) {
+                $metadata['mets_label'] = [(string) $dayLabel[0]];
+            }
+        }
+
         // Files are not expected to reference a dmdSec
         if (isset($this->fileInfos[$id]) || isset($hasMetadataSection['dmdSec'])) {
             return $metadata;
--- a/Classes/Common/MetsDocument.php
+++ b/Classes/Common/MetsDocument.php
@@ -181,10 +181,8 @@ final class MetsDocument extends Doc
         $details = $this->getLogicalStructure($id);
         if (!empty($details)) {
             $metadata['mets_order'][0] = $details['order'];
-            if ($metadata['type'][0] != 'issue') {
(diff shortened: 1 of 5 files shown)


sebastian-meyer:
I don't like this approach because it tries to fix an issue with METS by adding a workaround in the code instead of fixing the METS in the first place. Furthermore, this is problematic, because there are several types of _issues_, not just newspaper _issues_, so we can't just do this for every _issue_ which does not have a `LABEL`.

In general, in the ToC the `LABEL` of each entry is displayed (with `ORDERLABEL` being a possible fallback). If this is not available, we use the translated `TYPE` value as a generic title for the entry. This is how METS is meant to be interpreted and therefore how it should work in Kitodo.

In this case the problem is, that the entry for the issue doesn't have a `LABEL` or `ORDERLABEL`. This is fine from a METS perspective, but this results (correctly) in just having the `TYPE` as a generic title. In order to have another title, the correct way of doing this would be to add a `LABEL` to METS (as it is described in the [application profile](https://dfg-viewer.de/fileadmin/groups/dfgviewer/METS-Anwendungsprofil_2.3.1.pdf) and shown in the example on page 12).

Another approach could be to use the entry's metadata to get the publication date from there and add this to the ToC. If I remember correctly we already do something like this for volumes of multivolume works. You can even achieve this solely by using typoscript (have a look at the `plugin.tx_dlf_tableofcontents.menuConf` array of digital.slub-dresden.de).

Anyway, we shouldn't rely on other structures (like `day`) to get the information needed to display a meaningful title, because this works only in the special edge case you are encountering with newspapers.

In the current state I can't merge this, unfortunately. Could you please consider one of the options above (either fix the METS or get the issue's publication date from its metadata)?

beatrycze-volk:
Changes are made. Instead overwriting the label field, the year field was added. 

Title replacement is now configured in typoscript. It is possible to define for which type wich fields should be used.

Classes/Common/MetsDocument.php
```diff
@@ -333,7 +333,7 @@ protected function getLogicalStructureInfo(\SimpleXMLElement $structure, $recurs
         $details['orderlabel'] = (isset($attributes['ORDERLABEL']) ? $attributes['ORDERLABEL'] : '');
         $details['contentIds'] = (isset($attributes['CONTENTIDS']) ? $attributes['CONTENTIDS'] : '');
         $details['volume'] = '';
-        // Set volume information only if no label is set and this is the toplevel structure element.
+        // Set volume any year information only if no label is set and this is the toplevel structure element.
         if (
             empty($details['label'])
             && $details['id'] == $this->_getToplevelId()
```
sebastian-meyer: Should we really limit this to toplevel elements? It could be confusing if I'd configured some replacements in the ToC plugin but they don't work because the respective elements are not toplevel...

I suggest dropping this condition and just checking for empty labels and orderlabels:
```suggestion
            && empty($details['orderlabel'])
```

sebastian-meyer (commented):
Your current approach is way better than the last one, but it still has some issues.
For example: Using replacements is currently limited to `type`, `volume` and `year`, because those are the only information resolved in `MetsDocument->getLogicalStructureInfo()`. It's not possible to configure a replacement by any other metadata field. But on the other hand, documentation and implementation in `TableOfContentsController` suggests that you could use any metadata field (which is also what I'd expect).
Furthermore, doing part of the replacement in Typoscript and code (`volume` and `year`), and part of it in the Fluid template (`type`) seems inconsistent.

What do you think about the following suggestions:

1. Add a general extension configuration option to enable/disable loading the metadata for structure elements. The default should be "disabled" because of performance implications, since the metadata are loaded from METS.
2. In `MetsDocument->getLogicalStructureInfo()`, if for any structure `LABEL` and `ORDERLABEL` are empty and the above mentioned option is enabled, the whole metadata array for the corresponding structure element is saved in `$details['metadata']`.
3. Do the actual replacement/fallback in the Fluid template (i. e. `Resources/Private/Partials/TableOfContents/Children.html`), just like the already present fallback for using `type` if `title` is empty.

That way we would need less code, have a more consistent implementation and even more flexibility. The only downside I see is that if a lot of structure elements don't have a `LABEL` or `ORDERLABEL`, getting all the metadata might lower performance.

sebastian-meyer:
@beatrycze-volk What do you think about my proposal above? Since this is the last open bugfix for 4.1, I'd like to merge this soon in order to move on with TYPO3 v11 compatibility.

beatrycze-volk:
> @beatrycze-volk What do you think about my proposal above? Since this is the last open bugfix for 4.1, I'd like to merge this soon in order to move on with TYPO3 v11 compatibility.

Your proposal sounds interesting but I would make it in the new PR. The solution implemented here is tested and works so at the moment I would keep it that way.

Classes/Common/MetsDocument.php
```diff
@@ -333,15 +333,18 @@ protected function getLogicalStructureInfo(\SimpleXMLElement $structure, $recurs
         $details['orderlabel'] = (isset($attributes['ORDERLABEL']) ? $attributes['ORDERLABEL'] : '');
         $details['contentIds'] = (isset($attributes['CONTENTIDS']) ? $attributes['CONTENTIDS'] : '');
         $details['volume'] = '';
-        // Set volume information only if no label is set and this is the toplevel structure element.
+        // Set volume any year information only if no label is set and this is the toplevel structure element.
```
```suggestion
        // Set volume and year information only if no label is set and this is the toplevel structure element.
```

Classes/Controller/TableOfContentsController.php
```diff
@@ -272,30 +288,70 @@ private function getTranslatedType($type) {
...
+     *
+     * @return string
+     */
+    private function setTitle($entry) {
+        if (empty($entry['label']) && empty($entry['orderlabel'])) {
+            foreach ($this->settings['titleReplacements'] as $titleReplacement) {
+                if ($entry['type'] == $titleReplacement['type']) {
+                    $fields = explode(",", $titleReplacement['fields']);
+                    $title = '';
+
```
sebastian-meyer: Fix codacy issue (whitespaces in empty line):
```suggestion

```

sebastian-meyer (changes requested):
OK, then. I'll merge this as soon as those two little things are fixed. Thank you for your patience!

sebastian-meyer (approved):

PR #982, opened 2023-06-27
[MAINTENANCE] Fix and move Solr classes

beatrycze-volk:


sebastian-meyer (approved):
Thank you very much for cleaning up!

PR #985, opened 2023-07-17
[BUGFIX] Resolve link to the parent if the parent is multivolume_work or multipart_manuscript

beatrycze-volk:
The backlink for the parent if the parent is the `multivolume_work` /  `multipart_manuscript` was working correctly only for the last one child, for the others the METS link was used. It was happening because only for the last one condition `$entry['points'] === $doc->parentHref` returned `true`.
Reason: In the `multivolume_work` /  `multipart_manuscript` every child has separated anchor METS file (relation 1 to 1 not 1 to many children). All other children are read out from database and their anchor files are not matching.

### DIFF
--- a/Classes/Controller/TableOfContentsController.php
+++ b/Classes/Controller/TableOfContentsController.php
@@ -232,7 +232,7 @@ class TableOfContentsController extends AbstractController
         $doc = $this->document->getDoc();
         if (
             $doc instanceof MetsDocument
-            && $entry['points'] === $doc->parentHref
+            && ($entry['points'] === $doc->parentHref || $entry['type'] === 'multivolume_work')
             && !empty($this->document->getPartof())
         ) {
             unset($entry['points']);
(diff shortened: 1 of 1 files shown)


Classes/Controller/TableOfContentsController.php
```diff
@@ -244,18 +244,35 @@ protected function resolveMenuEntry($entry)
...
      * @return string
      */
     private function getTranslatedType($type) {
         return Helper::translate($type, 'tx_dlf_structures', $this->settings['storagePid']);
     }
 
+    /**
+     * Check if element has type 'multivolume_work' or 'multipart_manuscript'.
+     * Those elements have for each child one parent anchor file instead one
+     * parent file for all children elements.
```
sebastian-meyer:
```suggestion
     * For Kitodo.Production prior to version 3.x, hierarchical child documents
     * always come with their own METS file for their parent document, even
     * if multiple documents in fact have the same parent. To make sure that all
     * of them point to the same parent document in Kitodo.Presentation, we
     * need some workaround here.
     *
     * @todo Should be removed when Kitodo.Production 2.x is no longer supported.
```

sebastian-meyer (changes requested):
This looks fine, I just suggested a bit more documentation.

sebastian-meyer (approved):

PR #988, opened 2023-07-20
[BUGFIX] Remove not existing getUid and getLocation

### LINKED ISSUE #968: [BUG] Get METS url with newest presentation version
## Description

With presentation 3.3 there was a variable called `$uid` stored in a `Document` object. It holds the UID (solr indexed) or the URL (mets) of the document.
Now with presentation 4.x the mentioned variable is empty for all METS files I've tested (around 30) or any solr indexed documents. 

Is this on purpose or a bug? If on purpose, what would be the correct way to get the url?

## Reproduction

Try any METS files and check if `$doc->uid` (for PageView) or `$this->document->getDoc()` (PageViewController) is empty or not. 
Example: https://digital.slub-dresden.de/oai/?verb=GetRecord&metadataPrefix=mets&identifier=oai:de:slub-dresden:db:id-512520879
But you can try any of those too: https://github.com/UB-Mannheim/kitodo-presentation/wiki/METS-Examples

## Environment

 - OS version: Debian
 - RDBMS version: Maria DB
 - Apache Solr version: -
 - TYPO3 version: 10.4.37
 - PHP version: 7.4.33

beatrycze-volk:
Parameters `$uid` and `$location` are currently accessible in class `Kitodo\Dlf\Domain\Model\Document`. The references in `Kitodo\Dlf\Common\Doc` needs to be removed.

### DIFF
--- a/Classes/Common/Doc.php
+++ b/Classes/Common/Doc.php
@@ -30,7 +30,6 @@ use Ubl\Iiif\Tools\IiifHelper;
  * @access public
  * @property int $cPid This holds the PID for the configuration
  * @property-read bool $hasFulltext Are there any fulltext files available?
- * @property-read string $location This holds the documents location
  * @property-read array $metadataArray This holds the documents' parsed metadata array
  * @property-read int $numPages The holds the total number of pages
  * @property-read int $parentId This holds the UID of the parent document or zero if not multi-volumed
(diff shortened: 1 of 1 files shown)


csidirop:
Is there a reason why we do not use them anymore?

As I mentioned in #968 we could use those till v3.3.4 .

beatrycze-volk:
We are using them. They are now part of `Kitodo\Dlf\Domain\Model\Document` class.

sebastian-meyer (approved):

PR #993, opened 2023-07-21
[BUGFIX] Minor fixes in ToolboxController

beatrycze-volk:


sebastian-meyer (changes requested):
I suggest renaming method `emptyFullText` to something like `isFulltextEmpty` or `hasFulltext` to make it's purpose more clear.

sebastian-meyer (approved):
Thank you!

PR #1005, opened 2023-09-01
[FEATURE] Display ORDERLABEL as year in the search results if the structure type is year

beatrycze-volk:
year


sebastian-meyer:
Originally the idea was to control the display of metadata in list views (i. e. search results) by setting the `is_listed` switch: https://github.com/kitodo/kitodo-presentation/blob/28a365c435d2defda6c1e29f3dc8fa7be3178be7/Configuration/TCA/tx_dlf_metadata.php#L214
Thus showing/hiding a metadata field could be achieved by simply changing the field's configuration in the backend instead of having to edit the Fluid template. Therefore I would expect the template to just iterate over all metadata fields where `is_listed` is `true` and display them all.
What do you think?

beatrycze-volk:
After checking it out again I have fund out that actually structure **year** doesn't contain any metadata and that's why solution with `is_listed` doesn't work correctly there:

year

Can this solution stay or should i change the way how documents with structure **year** are indexed?

sebastian-meyer (approved):

PR #1006, opened 2023-09-04
[MAINTENANCE] Remove not needed sanitize for double parameter

beatrycze-volk:
It happens already in `sanitizeRequestData()` function of `AbstractController`


beatrycze-volk:
Tests failed because of the problems with installing dependencies:
```
The archive may contain identical file names with different capitalization (which fails on case insensitive filesystems)
    Unzip with unzip command failed, falling back to ZipArchive class
    Install of typo3/cms-fluid-styled-content failed
```

Probably they should be restarted.

sebastian-meyer (approved):

PR #1015, opened 2023-09-15
[MAINTENANCE] Remove database fields from ext_tables.sql which are referenced in TCA

### LINKED ISSUE #399: Open Maintenance Tasks
### General Maintenance and Modernization
- [x] Support Fluid templates for easy content rendering (especially for metadata)
- [x] #1018
- [x] #1045
- [x] Complete and fix phpDoc (in the given order)
  - [x] #1012 (`@package`, `@subpackage`, `@access`, `@abstract`, `@property`)
  - [x] #1013 (`@access`, `@static`, `@var`)
  - [x] #1014 (`@access`, `@static`, `@param`, `@return`)
  - [x] #1011 (since it only works on class level)
- [ ] #1015 (see [Auto generated structure](https://docs.typo3.org/m/typo3/reference-coreapi/master/en-us/ExtensionArchitecture/FilesAndLocations/Index.html#ext-tables-sql))
- [x] #1010

### Update Dependencies
- [x] Clean up and update Javascript libraries
  - [x] jPlayer
  - [x] jQuery
  - [x] jQuery UI
  - [x] OpenLayers

beatrycze-volk:
Change for https://github.com/kitodo/kitodo-presentation/issues/399

### DIFF
--- a/ext_tables.sql
+++ b/ext_tables.sql
@@ -2,16 +2,6 @@
 -- Table structure for table 'tx_dlf_documents'
 --
 CREATE TABLE tx_dlf_documents (
-    uid int(11) NOT NULL auto_increment,
-    pid int(11) DEFAULT '0' NOT NULL,
-    tstamp int(11) DEFAULT '0' NOT NULL,
-    crdate int(11) DEFAULT '0' NOT NULL,
-    cruser_id int(11) DEFAULT '0' NOT NULL,
-    deleted smallint(6) DEFAULT '0' NOT NULL,
-    hidden smallint(6) DEFAULT '0' NOT NULL,
-    starttime int(11) DEFAULT '0' NOT NULL,
-    endtime int(11) DEFAULT '0' NOT NULL,
-    fe_group varchar(100) DEFAULT '' NOT NULL,
     prod_id varchar(255) DEFAULT '' NOT NULL,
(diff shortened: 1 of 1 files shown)


beatrycze-volk:
I'm not really sure if this an error in the application or an error in the test setup. I'm going to make some updates in test setup and ten come back to this PR.

beatrycze-volk:
@sebastian-meyer I have found the reason why those tests are failing. DB data must be imported as CSV instead of deprecated XML format. I'm currently transforming those files, but probably rest of newly inserted tests should be inserted with CSV files to avoid double work.

sebastian-meyer:
Thanks for investigating! I didn't even know they changed that for TYPO3...

@oliver-stoehr Could you please consider this for the remaining pull requests which add new fixtures?

beatrycze-volk:
At the end it looks that even after those changes it doesn't pass the tests. The database tables are still created incorrectly. They are  missing fields removed from `ext_tables.sql.`

sebastian-meyer:
According to [documentation](https://docs.typo3.org/m/typo3/reference-coreapi/main/en-us/ExtensionArchitecture/FileStructure/ExtTablesSql.html) this looks fine, but tests are still failing. Are we not using the extension manager to set the database up?

beatrycze-volk:
It is not going to work with our test set up.

PR #1023, opened 2023-09-27
Add namespace for SolrSearch

beatrycze-volk:
Class SolrSearch was moved to Solr directory


sebastian-meyer:
After updating from `master` there are no more changes in this pull request.

PR #1024, opened 2023-09-27
[BUGFIX] Make Toolbox render again

frank-ulrich-weber:
$this->settings['tool'] does not exist, insted there is $this->settings['tools'] with a comma seperated list of tools that cannot be passed to switch directly


beatrycze-volk:
I'm not sure in which way you are testing that you have this settings not renamed but maybe this change is missing for you (https://github.com/slub/dfg-viewer/pull/235). Configurations for toolbox are placed in kitodo-presentation and dfg-viewer.

frank-ulrich-weber:
ok, maybe I understood something wrong here. I added the Toolbox Plugin, the Toolbox Default Templates and than select the required tools within the Toolbox Plugin Setup. No Tools are displayed with that setup. Then I stepped through renderTools() method and realised that there is no $this->settings['tool'], but instead $this->settings['tools'] with a comma seperated list of the selected tools and thought this must be a bug. Where is my mistake in thinking?

sebastian-meyer:
I think there is a misunderstanding about how the toolbox is supposed to work. The original idea was, that you can add the toolbox to your site once and it then can contain multiple tools. That's why we had a comma-separated list of tools which where all rendered inside one toolbox.
As far as I understand, the changes from @beatrycze-volk for the DFG-Viewer are now using multiple instances of toolboxes, each containing a single tool.

I suggest restoring the old behavior, because that wouldn't break current installations and gives more flexibility (as you can add one toolbox with multiple tools as well as several toolboxes with one tool each).

@beatrycze-volk and @frank-ulrich-weber: What do you think?

beatrycze-volk:
> As far as I understand, the changes from @beatrycze-volk for the DFG-Viewer are now using multiple instances of toolboxes, each containing a single tool.
> 
> I suggest restoring the old behavior, because that wouldn't break current installations and gives more flexibility (as you can add one toolbox with multiple tools as well as several toolboxes with one tool each).
> 
> @beatrycze-volk and @frank-ulrich-weber: What do you think?

But I didn't change configuration for tools, only renamed it... I was not reimplementing the logic, only refactored methods and renamed configuration. The split onto separated tools in configuration needed to happen earlier.

sebastian-meyer:
OK, if the functionality is unchanged, we should revert to `tools` instead of `tool` for the settings, because it can in fact be a comma-separated list of multiple tools.

sebastian-meyer:
So, I'll merge this for Kitodo.Presentation. @beatrycze-volk Do you revert changes for the DFG-Viewer accordingly?

Classes/Controller/ToolboxController.php
```diff
@@ -62,40 +62,46 @@ public function mainAction(): void
...
-                    $this->renderToolByName('renderSearchInDocumentTool');
-                    break;
-                default:
-                    $this->logger->warning('Incorrect tool configuration: "' . $this->settings['tool'] . '". This tool does not exist.');
+        if (!empty($this->settings['tools'])) {
+
+            $tools = explode(',', $this->settings['tools']);
+
+            foreach($tools as $tool)
+            {
```
sebastian-meyer:
```suggestion
            foreach($tools as $tool) {
```

sebastian-meyer (changes requested):

Classes/Controller/ToolboxController.php
```diff
@@ -62,40 +62,45 @@ public function mainAction(): void
...
-                case 'searchindocumenttool':
-                    $this->renderToolByName('renderSearchInDocumentTool');
-                    break;
-                default:
-                    $this->logger->warning('Incorrect tool configuration: "' . $this->settings['tool'] . '". This tool does not exist.');
+        if (!empty($this->settings['tools'])) {
+
+            $tools = explode(',', $this->settings['tools']);
+
+            foreach($tools as $tool) {
```
sebastian-meyer:
```suggestion
            foreach ($tools as $tool) {
```

sebastian-meyer (changes requested):

sebastian-meyer (approved):

PR #1025, opened 2023-09-28
[BUGFIX] Fix initialization and checks for rawData in API profiles

beatrycze-volk:


sebastian-meyer (commented):
I suggest using type-sensitive comparisons to avoid false-positives (i. e. `!== ` instead of `!=`).

sebastian-meyer (approved):

PR #1026, opened 2023-09-28
[BUGFIX] Fix getting AbstractDocument instance from cache

beatrycze-volk:


Classes/Common/AbstractDocument.php
```diff
@@ -401,12 +401,15 @@ public static function &getInstance(string $location, array $settings = [], bool
         $xml = null;
         $iiif = null;
 
-        if ($instance = self::getDocCache($location) && !$forceReload) {
-            return $instance;
-        } else {
-            $instance = null;
+        if (!$forceReload) {
+            $instance = self::getDocumentCache($location);
+            if ($instance != false) {
```
```suggestion
            if ($instance !== false) {
```

sebastian-meyer (changes requested):

sebastian-meyer (approved):

PR #1027, opened 2023-09-28
[BUGFIX] Fix returns in functions of IiifManifest class

beatrycze-volk:


Classes/Common/IiifManifest.php
```diff
@@ -404,7 +404,7 @@ public function getFileInfo($id)
     public function getFileLocation(string $id): string
     {
         if ($id == null) {
-            return null;
+            return "";
```
sebastian-meyer:
```suggestion
            return '';
```

sebastian-meyer (approved):

PR #1018, opened 2023-09-25
[MAINTENANCE] Consistently use type declaration for parameters and return values - part 1

beatrycze-volk:
Change for https://github.com/kitodo/kitodo-presentation/issues/399

PR #1045, opened 2023-10-17
[MAINTENANCE] Consistently use type declaration for parameters and return values - part 2

beatrycze-volk:
Changes for controller and domain classes - follow up to https://github.com/kitodo/kitodo-presentation/pull/1018 and part of https://github.com/kitodo/kitodo-presentation/issues/399 issue.

### DIFF
--- a/Classes/Controller/Backend/NewTenantController.php
+++ b/Classes/Controller/Backend/NewTenantController.php
@@ -442,7 +442,7 @@ class NewTenantController extends AbstractController
      *
      * @access public
      *
-     * @return ResponseInterface
+     * @return void
      */
     public function errorAction()
     {
--- a/Classes/Controller/Backend/NewTenantController.php
+++ b/Classes/Controller/Backend/NewTenantController.php
@@ -22,7 +22,6 @@ use Kitodo\Dlf\Domain\Repository\FormatRepository;
 use Kitodo\Dlf\Domain\Repository\MetadataRepository;
 use Kitodo\Dlf\Domain\Repository\StructureRepository;
 use Kitodo\Dlf\Domain\Repository\SolrCoreRepository;
-use Psr\Http\Message\ResponseInterface;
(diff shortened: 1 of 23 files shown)


Classes/Controller/Backend/NewTenantController.php
```diff
@@ -441,7 +442,7 @@ public function indexAction()
      *
      * @access public
      *
-     * @return void
+     * @return ResponseInterface
```
sebastian-meyer: How does this return a `ResponseInterface`? The method is empty...

Classes/Domain/Model/Document.php
```diff
@@ -637,8 +637,6 @@ public function addCollection(Collection $collection): void
      * Removes a collection
      *
      * @param Collection $collection
-     *
-     * @return ObjectStorage<Collection> collections
      */
     public function removeCollection(Collection $collection)
```
sebastian-meyer: This should have a `void` return value. You removed the (wrong) return value from PHPDoc, but didn't add it to the function's signature. Or am I missing something?
```suggestion
    public function removeCollection(Collection $collection): void
```

sebastian-meyer (changes requested):

Classes/Controller/Backend/NewTenantController.php
```diff
@@ -22,6 +22,7 @@
 use Kitodo\Dlf\Domain\Repository\MetadataRepository;
 use Kitodo\Dlf\Domain\Repository\StructureRepository;
 use Kitodo\Dlf\Domain\Repository\SolrCoreRepository;
+use Psr\Http\Message\ResponseInterface;
```
sebastian-meyer: This is never used.
```suggestion
```

sebastian-meyer (changes requested):

sebastian-meyer (approved):

PR #1052, opened 2023-10-17
[MAINTENANCE] Introduce usage of PHPStan checks for new PR

beatrycze-volk:
Depends on #1045


Classes/Common/AbstractDocument.php
```diff
@@ -441,9 +441,11 @@ public static function &getInstance(string $location, array $settings = [], bool
         // Sanitize input.
         $pid = max(intval($settings['storagePid']), 0);
         if ($documentFormat == 'METS') {
-            $instance = new MetsDocument($location, $pid, $xml);
+            $instance = new MetsDocument($pid,$location, $xml, $settings);
```
```suggestion
            $instance = new MetsDocument($pid, $location, $xml, $settings);
```

.github/phpstan.neon
```diff
@@ -0,0 +1,32 @@
...
+            - '#Call to an undefined method Ubl\\Iiif\\Presentation\\Common\\Model\\Resources\\IiifResourceInterface::getWidth\(\)\.#'
+            - '#Call to an undefined method Ubl\\Iiif\\Presentation\\Common\\Model\\Resources\\IiifResourceInterface::getPossibleTextAnnotationContainers\(\)\.#'
+            - '#Call to an undefined method Ubl\\Iiif\\Presentation\\Common\\Model\\Resources\\IiifResourceInterface::getTextAnnotations\(\)\.#'
+            - '#Call to an undefined method Ubl\\Iiif\\Presentation\\Common\\Model\\Resources\\ManifestInterface::getOriginalJsonArray\(\)\.#'
+            - '#Call to an undefined method Ubl\\Iiif\\Presentation\\Common\\Model\\Resources\\RangeInterface::getMemberRangesAndRanges\(\)\.#'
+            - '#Constant LOG_SEVERITY_ERROR not found\.#'
+            - '#Constant LOG_SEVERITY_NOTICE not found\.#'
+            - '#Constant LOG_SEVERITY_WARNING not found\.#'
+            - '#Constant TYPO3_MODE not found\.#'
+            - '#Method Kitodo\\Dlf\\Controller\\Backend\\NewTenantController::errorAction\(\) should return Psr\\Http\\Message\\ResponseInterface but return statement is missing\.#'
```
sebastian-meyer: This should be fixed in the code instead of ignoring the error message. But maybe this is just a side-effect of the recursion from #1045 (see below).
```suggestion
```

.github/phpstan.neon
```diff
@@ -0,0 +1,32 @@
...
+    level: 5
+    paths:
+        - ../Classes/
+    excludePaths:
+        - ../Classes/Controller/OaiPmhController.php
```
sebastian-meyer, line 30: Why do you exclude this controller? Is it so bad? ;o)

.github/phpstan.neon
```diff
@@ -0,0 +1,32 @@
...
+    treatPhpDocTypesAsCertain: false
```
sebastian-meyer:
```suggestion
    treatPhpDocTypesAsCertain: false

```

.github/workflows/phpstan.yml
```diff
@@ -0,0 +1,25 @@
...
+      - name: Install dependencies
+        uses: php-actions/composer@v6
+        with:
+          command: update
+          php_version: "7.4"
+
+      - name: PHPStan Static Analysis
+        uses: php-actions/phpstan@v3
+        with:
+          configuration: ./.github/phpstan.neon
```
sebastian-meyer:
```suggestion
          configuration: ./.github/phpstan.neon

```

Classes/Common/AbstractDocument.php
```diff
@@ -1253,7 +1257,7 @@ public function __set(string $var, $value): void
      *
      * @param string $location
      *
-     * @return AbstractDocument|false
+     * @return MetsDocument|IiifManifest|false
```
sebastian-meyer: What's wrong with having an abstract class as return type? Can't we make this work with PHPStan?

Classes/Common/Solr/SearchResult/Highlight.php
```diff
@@ -33,6 +33,7 @@ class Highlight
      * @access private
      * @var int The parent region's identifier
      */
+    // @phpstan-ignore-next-line
```
sebastian-meyer: Why do we need to ignore the next line? Because it can be NULL and not just an integer? Shouldn't we then declare the property as `private ?int $parentRegionId;` instead of ignoring PHPStan errors?

Classes/Common/Solr/SolrSearchQuery.php
```diff
@@ -55,6 +55,7 @@ public function __construct($solrSearch)
         $this->limit = count($solrSearch);
     }
 
+    // @phpstan-ignore-next-line
```
sebastian-meyer, line 59: Again: Can't we fix the PHPStan configuration instead of adding those ignore comments? It otherwise suggests there is an error in the code, which it isn't in this case.

Classes/Controller/Backend/NewTenantController.php
```diff
@@ -22,6 +22,7 @@
 use Kitodo\Dlf\Domain\Repository\MetadataRepository;
 use Kitodo\Dlf\Domain\Repository\StructureRepository;
 use Kitodo\Dlf\Domain\Repository\SolrCoreRepository;
+use Psr\Http\Message\ResponseInterface;
```
sebastian-meyer: This is a recursion from #1045 and should be removed.
```suggestion
```

Classes/Controller/Backend/NewTenantController.php
```diff
@@ -441,7 +442,7 @@ public function indexAction(): void
      *
      * @access public
      *
-     * @return void
+     * @return ResponseInterface
```
sebastian-meyer: This is a recursion from #1045 and should be kept unchanged or removed entirely.
```suggestion
     * @return void
```
Furthermore, since the method is empty, it should declare it's return type in it's signature as well.

sebastian-meyer (changes requested):
Thank you very much for this pull request! I appreciate your hard work on improving and maintaining the code quality! Overall this looks very good and I'd like to add PHPStan to our array of static code analyzers.
What I don't like are a lot of those `// @phpstan-ignore-next-line` comments, which suggest, that there is some error in the code. Please, either also add a `TODO` to indicate what is wrong (as you did many times) or try to change PHPStan's configuration in order to not report false positives. (Disclaimer: I am no expert with PHPStan, so maybe my request is unrealistic...)

.github/phpstan.neon
beatrycze-volk: Yeah, I was changing there to ResponseInterface, but as you good pointed out currently we are not returning there anything so we should ignore this error.

.github/phpstan.neon
beatrycze-volk: Yes, it is bad and any tries fixing it are causing all tests to fail (functioniality itself works, but tests are failing). I will try to fix in the future but for now I decided to put it aside.

Classes/Common/Solr/SearchResult/Highlight.php
beatrycze-volk: Here actually problem is that PHPStan doesn'T allow properties which are assigned but never used. At the moment we are not using this property but I was adding it to keep the full result returned from SOLR.

Classes/Common/Solr/SolrSearchQuery.php
beatrycze-volk: This class in general is also problematic becasue it inherits a lot of methods but actually doesn't implement them. PHPStan doesn't like such a situation.

Classes/Controller/Backend/NewTenantController.php
beatrycze-volk: PHPStan will not accept this method declaring void because it conflict with the method of parent class.
So there are 2 posibilities:
1. declare return as void and keep ignore of error message 
2. declare return as ResponseInterface and remove ignore of error message

Which one do you prefer?

.github/phpstan.neon
beatrycze-volk: https://github.com/kitodo/kitodo-presentation/pull/1052#discussion_r1368926320

Classes/Common/AbstractDocument.php
beatrycze-volk: Reversed

Classes/Common/Solr/SearchResult/Highlight.php
sebastian-meyer: I think PHPStan is right on this one. ;o) We should remove the property as long as we don't need it. Maybe leave a comment if you want to keep a reminder of how to access the full result should we need it someday?

Classes/Common/Solr/SolrSearchQuery.php
sebastian-meyer: Yes, and PHPStan is right in doing so, I think! ;o)

But I understand that we can't do anything about it at the moment, so let's keep the comments in this file!

Classes/Controller/Backend/NewTenantController.php
sebastian-meyer: Then why not simply remove this method? As far as I can see the parent class' declaration isn't abstract, so we could just let it be inherited...

sebastian-meyer (dismissed):

Classes/Common/Helper.php
```diff
@@ -603,7 +603,7 @@ public static function getURN(string $base, string $id): string
         for ($i = 0, $j = strlen($digits); $i < $j; $i++) {
             $checksum += ($i + 1) * intval(substr($digits, $i, 1));
         }
-        $checksum = substr(strval(intval($checksum / intval(substr($digits, -1, 1)))), -1, 1);
+        $checksum = substr((string) (intval($checksum / intval(substr($digits, -1, 1))), -1, 1);
```
beatrycze-volk: `$checksum = substr((string) (intval($checksum / intval(substr($digits, -1, 1)))), -1, 1);`

Classes/Common/Helper.php
```diff
@@ -605,7 +603,7 @@ public static function getURN(string $base, string $id): string
         for ($i = 0, $j = strlen($digits); $i < $j; $i++) {
             $checksum += ($i + 1) * intval(substr($digits, $i, 1));
         }
-        $checksum = substr(intval($checksum / intval(substr($digits, -1, 1))), -1, 1);
+        $checksum = substr((string) (intval($checksum / intval(substr($digits, -1, 1)))), -1, 1);
```
sebastian-meyer:
```suggestion
        $checksum = substr((string) floor($checksum / (int) substr($digits, -1, 1)), -1, 1);
```

sebastian-meyer (changes requested):

sebastian-meyer (approved):

PR #1062, opened 2023-10-24
[BUGFIX] Do not assume that return value from `Extbase\Persistence\Repository::findOneByRecordId` is not null

csidirop:
This PR fixes an exception thrown when loading a document:

`Uncaught TYPO3 Exception: Typed property Kitodo\Dlf\Controller\AbstractController::$document must be an instance of Kitodo\Dlf\Domain\Model\Document, null used | TypeError thrown in file /var/www/typo3/public/typo3conf/ext/dlf/Classes/Controller/AbstractController.php in line 143.`

Now with type declarations we should check if returnvalues outside our repository (here from `Extbase\Persistence\Repository::findOneByRecordId` ) are not null.

### DIFF
--- a/Classes/Controller/AbstractController.php
+++ b/Classes/Controller/AbstractController.php
@@ -140,13 +140,15 @@ abstract class AbstractController extends ActionController implements LoggerAwar
 
                 if ($doc !== null) {
                     if ($doc->recordId) {
-                        $this->document = $this->documentRepository->findOneByRecordId($doc->recordId);
-                    }
-
-                    if ($this->document === null) {
-                        // create new dummy Document object
-                        $this->document = GeneralUtility::makeInstance(Document::class);
-                    }
+                        // find document from repository by recordId
+                        $docFromRepository = $this->documentRepository->findOneByRecordId($doc->recordId);
+                        if ($docFromRepository !== null) {
+                            $this->document = $docFromRepository;
+                        } else {
+                            // create new dummy Document object
+                            $this->document = GeneralUtility::makeInstance(Document::class);
+                        }
+                     }
 
                     // Make sure configuration PID is set when applicable
                     if ($doc->cPid == 0) {
--- a/Classes/Controller/AbstractController.php
+++ b/Classes/Controller/AbstractController.php
@@ -148,7 +148,7 @@ abstract class AbstractController extends ActionController implements LoggerAwar
                             // create new dummy Document object
                             $this->document = GeneralUtility::makeInstance(Document::class);
                         }
-                     }
+                    }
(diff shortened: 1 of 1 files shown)


csidirop:
Regarding codestyle. Do you prefer one combined if-statement `if ($doc->recordId & $this->documentRepository->findOneByRecordId($doc->recordId) === null)` or two separate nested?

Classes/Controller/AbstractController.php
```diff
@@ -137,15 +137,12 @@ protected function loadDocument(int $documentId = 0): void
...
-
                 if ($doc !== null) {
                     if ($doc->recordId) {
-                        $this->document = $this->documentRepository->findOneByRecordId($doc->recordId);
-                    }
-
-                    if ($this->document === null) {
-                        // create new dummy Document object
-                        $this->document = GeneralUtility::makeInstance(Document::class);
+                        if ($this->documentRepository->findOneByRecordId($doc->recordId) === null) {
```
sebastian-meyer: There is an `else` statement missing for when there is a valid document in the repository.

(The original code sets `$this->document` to the document retrieved from the repository and uses a dummy document only as a fallback.)

sebastian-meyer (changes requested):
I personally prefer all conditions in a single `if` statement, because I consider it easier to read.

Classes/Controller/AbstractController.php
csidirop: Yes, forgot the else statement. Ill add it, but than leave the nested if statements:

```
                    if ($doc->recordId) {
                        if ($this->documentRepository->findOneByRecordId($doc->recordId) === null) {
                            // create new dummy Document object
                            $this->document = GeneralUtility::makeInstance(Document::class);
                        } else {
                            // find document from repository by recordId
                            $this->document = $this->documentRepository->findOneByRecordId($doc->recordId);
                        }
                    }
```

sebastian-meyer: You should avoid requesting the repository twice:
 ```
if ($doc->recordId) {
    // find document from repository by recordId
    $docFromRepository = $this->documentRepository->findOneByRecordId($doc->recordId);
    if ($docFromRepository !== null) {
        $this->document = $docFromRepository;
    } else {
        // create new dummy Document object
        $this->document = GeneralUtility::makeInstance(Document::class);
    }
}
```

csidirop: You're right

sebastian-meyer (dismissed):
Thanks a lot!

sebastian-meyer (approved):

PR #1067, opened 2023-11-02
[BUGFIX] Fix call to extension configuration

csidirop:
After 0be9160a566e093c0b8394454e52a38186549300 there is a wrong call in [MetsDocument.php#L476](https://github.com/kitodo/kitodo-presentation/blob/master/Classes/Common/MetsDocument.php#L476).

`$this->settings` are not the extension configuration but some other settings:

```
array(7) {
  ["storagePid"]=>
  string(1) "3"
  ["useInternalProxy"]=>
  string(1) "1"
  ["linkTitle"]=>
  string(1) "0"
  ["getTitle"]=>
  string(1) "0"
  ["showFull"]=>
  string(1) "1"
  ["rootline"]=>
  string(1) "1"
  ["separator"]=>
  string(1) "#"
}
```
Therefore we got following exception:

```
Core: Exception handler (WEB): Uncaught TYPO3 Exception: Argument 3 passed to Kitodo\Dlf\Format\Mods::extractMetadata() must be of the type bool, null given, called in /var/www/typo3/public/typo3conf/ext/dlf/Classes/Common/MetsDocument.php on line 477 | TypeError thrown in file /var/www/typo3/public/typo3conf/ext/dlf/Classes/Format/Mods.php in line 58
```

I propose a simple fix based on similar calls in this class.

### DIFF
--- a/Classes/Common/MetsDocument.php
+++ b/Classes/Common/MetsDocument.php
@@ -473,7 +473,7 @@ final class MetsDocument extends AbstractDocument
                         class_exists($class)
                         && ($obj = GeneralUtility::makeInstance($class)) instanceof MetadataInterface
                     ) {
-                        $obj->extractMetadata($this->mdSec[$dmdId]['xml'], $metadata, $this->settings['useExternalApisForMetadata']);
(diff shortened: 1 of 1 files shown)


sebastian-meyer (approved):
`$this->settings` should contain the plugin's settings merged with the extension's settings, but the latter seem to be missing in this case.

But now it at least is consistent within this class!

PR #1069, opened 2023-11-02
[BUGFIX] Change way of key generation for collection

beatrycze-volk:
There is no field 'priority' so some other field needs to be used for random key generation.

First commit contains the change.
Second is for taking collection processing to the own function - no logical change in the code working.


sebastian-meyer:
There should be a field `priority` (see TCA for `tx_dlf_collections`). So instead of removing calculations with priority we should investigate why the priority value is not read from the DB table.

The idea of this priority is that you can display your collections in a randomized order but still have important collections shown more towards the top of the list.

beatrycze-volk:
> There should be a field `priority` (see TCA for `tx_dlf_collections`). So instead of removing calculations with priority we should investigate why the priority value is not read from the DB table.

The result which is used here to create the collections array is taken out of SOLR not from the DB.

sebastian-meyer:
Well, then the problem happens even earlier when indexing the collections in Solr. ;o)

beatrycze-volk:
> Well, then the problem happens even earlier when indexing the collections in Solr. ;o)

So should be this field indexed?

sebastian-meyer:
Sorry, I checked if someone actually uses this feature, but then forgot to write an answer here. Unfortunately we cannot get rid of the priority for collections since we use this ourselves in the SLUB Digitale Sammlungen. ;o)

So: Yes, I suppose we need to index the priority field for collections.

beatrycze-volk:
I have found actual the problem. The variable `$collection` was renamed to `$collectionInfo`, but it looks that in that place it shouldn't be renamed: https://github.com/kitodo/kitodo-presentation/commit/3d17862a0c01f3438d9e9f9bde4943d643971ae4

sebastian-meyer (approved):
Nice catch!

PR #1074, opened 2023-11-02
[BUGFIX] Make `useOriginalIiifManifestMetadata` class parameter

beatrycze-volk:


sebastian-meyer (commented):
Why did you add an abstract method to `AbstractDocument`? That would make sense if it'd be a generic `getMetadata` method, but since it's specifically only for IIIF metadata, I think it shouldn't be part of the base class. Don't you think?

beatrycze-volk:
Ok, I will remove it and put ignore for PHPStan.

sebastian-meyer (approved):

PR #1075, opened 2023-11-02
fix order of "Previous Page" and page + last page and "Next Page"

BFallert:
"Previous Page" and "<number of first page>"  instead of "<number of first page>" and "Previous Page" and
"<number of last page>" and "Next Page" instead of "Next Page" and "<number of last page>"

### DIFF
--- a/Resources/Private/Partials/Lists/Pagination.html
+++ b/Resources/Private/Partials/Lists/Pagination.html
@@ -3,19 +3,19 @@
     <ul class="pagination">
         <f:if condition="{pagination.previousPageNumber} && {pagination.previousPageNumber} >= {pagination.firstPageNumber}">
             <f:then>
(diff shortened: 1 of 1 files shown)


BFallert:
If you use icons for the links instead of text, the original order makes perfect sense, e.g. 1 corresponds to |< and 5 corresponds to >|.
|< < 2 [3] 4 > >|

sebastian-meyer:
As you already mentioned: when using icons this is perfectly fine. Since this is up to personal preference and easily changeable via template, I'd rather not change anything for consistency reasons. Otherwise everybody who is using the default templates would have to fix their templates after upgrading to keep the previous behavior.

Do you agree?

BFallert:
Yes, I agree with this.

PR #1076, opened 2023-11-03
[BUGFIX] Fix missing pagination between first and last page

BFallert:
Currently still a problem: no summary (...) is formed yet with too many pages. But better all pages then no pages.

Maybe someone has an idea on how to reinstall this functionality.


BFallert:
In the documentation for Typo3 11 ([https://docs.typo3.org/m/typo3/reference-coreapi/11.5/en-us/ApiOverview/Pagination/Index.html](https://docs.typo3.org/m/typo3/reference-coreapi/11.5/en-us/ApiOverview/Pagination/Index.html)) an extension is mentioned that solves the problem with the dots. But this work only with the old pagination, so it is no option (as i cab see).

And in the documentation for Typo3 12 [https://docs.typo3.org/m/typo3/reference-coreapi/12.4/en-us/ApiOverview/Pagination/Index.html#sliding-window-pagination](https://docs.typo3.org/m/typo3/reference-coreapi/12.4/en-us/ApiOverview/Pagination/Index.html#sliding-window-pagination) it is mentioned how this problem can be solved with Typo3 version 12. 

The description of the behavior, in Typo3 12, is taken verbatim from the extension:

_**Sliding window pagination**_
 _New in version 12.0._

_The sliding window pagination can be used to paginate array items or query results from Extbase. The main advantage is that it reduces the amount of pages shown._

_Example: Imagine 1000 records and 20 items per page which would lead to 50 links. Using the SlidingWindowPagination, you will get something like < 1 2 ... 21 22 23 24 ... 100 >._


But at the moment, I don't see how that could benefit us at the moment.

BFallert:
I moved my discussion about the ... into an issue see https://github.com/kitodo/kitodo-presentation/issues/1077

sebastian-meyer:
Again, this seems to be a case of personal preference. I am hesitant to change the default behavior, because it means that people using the provided templates have to update theirs to keep the old behavior.
But I appreciate your work in providing an alternative display for the pagination. Maybe add your changes only as comments, so that the default behavior isn't changed, but somebody who prefers your pagination style can easily adapt their template by using the commented variant?

BFallert:
I cannot fully agree.
The line `<f:for each="{pagination.allPageNumbers}" as="page">` contains an error, the variable name is wrong and should be `pagination.pages` correctly. Without this correction, no output will be made at this point.

I have commented out the line that prevents the duplicate output of the first / last page, as suggested

sebastian-meyer (approved):

sebastian-meyer:
That makes totally sense! Thank you very much!

PR #1079, opened 2023-11-03
[MAINTENANCE] Reorder keywords and abstract functions in AbstractDocument class

beatrycze-volk:


beatrycze-volk:
It is necessary to trigger tests again as changes in the commit couldn't cause the failure.

sebastian-meyer:
Beware: The `_get*` methods in `AbstractDocument`, `MetsDocument` and `IiifManifest` are used by the magic getter method (see `AbstractDocument->__get()`! Renaming those methods means that a lot of protected class members are no longer accessible (which most likely causes the tests to fail).
On the other hand, simply changing the magic getter method would make a lot of purposely protected class members accessible as well. So we have to distinguish magic getter methods from "normal" `getSomething()` methods.

I understand that using a preceding underscore is deprecated and even disallowed by PSR-12 coding styles. I already discussed this issue with @frank-ulrich-weber. Maybe we can find some other unique prefix to use for those getter methods? Something like `magicGetSomething()`?

beatrycze-volk:
> Beware: The `_get*` methods in `AbstractDocument`, `MetsDocument` and `IiifManifest` are used by the magic getter method (see `AbstractDocument->__get()`! Renaming those methods means that a lot of protected class members are no longer accessible (which most likely causes the tests to fail). On the other hand, simply changing the magic getter method would make a lot of purposely protected class members accessible as well. So we have to distinguish magic getter methods from "normal" `getSomething()` methods.
> 
> I understand that using a preceding underscore is deprecated and even disallowed by PSR-12 coding styles. I already discussed this issue with @frank-ulrich-weber. Maybe we can find some other unique prefix to use for those getter methods? Something like `magicGetSomething()`?

I'm fine with reverting this last change if the Codacy rules are getting disabled. I'm also fine with rename to some other like proposed by you. So tell me which way you prefer and I will adjust this PR.

sebastian-meyer:
I think the best option would be to rename all `_get*` methods to `magicGet*` and change `AbstractDocument->__get()` accordingly. This way we'd fulfill all PSR-12 requirements checked by Codacy and still distinguish between magic and normal getter methods.

beatrycze-volk:
```
Issues
======
+ Solved 72
- Added 1
```
`Protected method name "AbstractDocument::magicGetCPid" is not in camel caps format`

sebastian-meyer:
That's fine! I'll add an exception in the Codacy rules.

sebastian-meyer (approved):

PR #1080, opened 2023-11-03
[BUGFIX] DataHandler: Change uid type from `int` to `string`

### LINKED ISSUE #1078: [BUG] Page creation not possible anymore
## Description

After creating and saving a new subpage following exception is thrown: 

`
Oops, an error occurred!
Argument 3 passed to Kitodo\Dlf\Hooks\DataHandler::processDatamap_postProcessFieldArray() must be of the type int, string given, called in /var/www/typo3/public/typo3/sysext/core/Classes/DataHandling/DataHandler.php on line 1178
`

## Reproduction

0. Get newest code (in my case 4dd7001b643607fb879c006288e8aa3a139fd5d0)
1. Go BE -> Page
2. Create a new subpage
3. Set a title and click on save
4. Exception is thrown

grafik

## Environment

 -  OS version: Debian (in Docker)
 -  RDBMS version: Maria DB (MySQL 5.5.5-10.11.2-MariaDB-1:10.11.2+maria~ubu2204)
 -  Apache Solr version: -
 -  TYPO3 version: 10.4.38-dev
 -  PHP version: 7.4.33

csidirop:
This should fix #1078. 

For `processDatamap_postProcessFieldArray()` this is save because both times `id` is used, it gets cast to int `intval($id))`

For `processDatamap_afterDatabaseOperations()` this should also be save. I believe the uid will always be an int when dealing with solr and indexed documents.

For `processCmdmap_postProcess()` it should be the same as for the last function. Although I'm not sure.

### DIFF
--- a/Classes/Hooks/DataHandler.php
+++ b/Classes/Hooks/DataHandler.php
@@ -72,7 +72,7 @@ class DataHandler implements LoggerAwareInterface
      *
      * @return void
      */
-    public function processDatamap_postProcessFieldArray(string $status, string $table, int $id, array &$fieldArray): void
+    public function processDatamap_postProcessFieldArray(string $status, string $table, string $id, array &$fieldArray): void
     {
         if ($status == 'new') {
             switch ($table) {
@@ -204,7 +204,7 @@ class DataHandler implements LoggerAwareInterface
      *
      * @return void
      */
-    public function processDatamap_afterDatabaseOperations(string $status, string $table, int $id, array &$fieldArray): void
+    public function processDatamap_afterDatabaseOperations(string $status, string $table, string $id, array &$fieldArray): void
     {
         if ($status == 'update') {
             switch ($table) {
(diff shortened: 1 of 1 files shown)


sebastian-meyer:
According to [TYPO3 Datahandler](https://github.com/TYPO3/typo3/blob/11.5/typo3/sysext/core/Classes/DataHandling/DataHandler.php) `$id` can be both an integer (in case of existing records) or a string (in case of new records). So both should be supported by our hook. Thus we need to remove the type-hint altogether.

Could you please change your pull request accordingly:
- [x] Remove type-hint for parameter `$id` in signature of `processDatamap_afterDatabaseOperations` and `processDatamap_postProcessFieldArray`.
- [x] Change PHPDoc of both methods to `int|string` for parameter `$id`.
- [x] Add comments to remind us of changing the signature again when dropping support for PHP 7 (because PHP 8 supports union types).

Regarding the PHPStan issue: When we accept `$id` to be a string, we have to cast it to integer in line 260. (It's always an integer at that point, because we are dealing with existing records, but PHPStan doesn't know that and it doesn't hurt to cast it to integer anyway.)

csidirop:
Thank you for taking on this

sebastian-meyer (approved):

PR #1081, opened 2023-11-03
[MAINTENANCE] Rename getTitledata to getToplevelMetadata

beatrycze-volk:
This name more clearly describes the purpose of this function.


sebastian-meyer:
I suppose from a developer's point of view you are right, but `getTitledata` is not completely wrong since "Titeldaten" is a technical term for the toplevel metadata of an object (i. e. those which would be found in a library catalog).

But since most developers are no librarians I'll accept your pull request. ;o)

sebastian-meyer (approved):

Issue #1029, opened 2023-09-28
[BUG] Issues after merging #885


sebastian-meyer:
**Tests under TYPO3 v10 with [our DFG-Viewer](https://github.com/UB-Mannheim/dfg-viewer/tree/6.x-ocr):**

- Everything search related is still not working. See my comment from [24 feb](https://github.com/kitodo/kitodo-presentation/pull/885#issuecomment-1443367080). Is it just me?
- The thumbnail preview (gridpage) shows only the first 24 pages and has no option to load more pages: (dfg-viewer issue?)
![grafik](https://github.com/kitodo/kitodo-presentation/assets/43964592/4cbd4f32-513a-4826-80c6-d22bca50ddcc)
Short addition: On an other document the text part was partially linked. First 24 pages with working link for the next 24:
![grafik](https://github.com/kitodo/kitodo-presentation/assets/43964592/67777cbf-0915-4105-9958-b6dae89c8d9d)
but the next page had only a working link back:
![grafik](https://github.com/kitodo/kitodo-presentation/assets/43964592/bfc0bbe8-9d45-4978-b913-5e80074f1e4b)
The two mentioned mets: https://digi.bib.uni-mannheim.de/fileadmin/vl/ubmaosi/59087/59087.xml and https://digi.bib.uni-mannheim.de/fileadmin/vl/ubmaosi/59088/59088.xml
- In addition, the numbers 1 and 31 refer to the number of review pages. But are somehow incorrectly displayed / placed. For reference, how it currently looks (and should be looking):
![grafik](https://github.com/kitodo/kitodo-presentation/assets/43964592/51a2087d-0936-4f37-8d68-2ec2a05ffbe8)

- The image manipulation tools are missing: (dfg-viewer issue?)
![grafik](https://github.com/kitodo/kitodo-presentation/assets/43964592/0aee3153-981b-4edf-b6b5-07a24dac4619)

- the same Metadata pairing/grouping issue described later

**Tests under TYPO3 v11:**
- I could not get the viewer running. But that can be on me. I had always issues manually setting presentation up. Ill try more. Is there anywhere a step by step guide?
  -> I could not try the viewer or anything related 
- New Tenant Module:
  - Can not create a new Solr Core:
![grafik](https://github.com/kitodo/kitodo-presentation/assets/43964592/7876cff6-45ad-4b9d-b649-4436936f2f4d)
    -> I did not set the correct solr configuration prior. But till v10 this wasn't necessary.
[...]

PR #1087, opened 2023-11-07
[BUGFIX] Use paginator in 'SearchController.php' and 'CollectionController.php'

csidirop:
This PR should fix my long standing issue with missing search results raised here https://github.com/kitodo/kitodo-presentation/pull/885#issuecomment-1443367080 and further discussed here #1029.

After a lot of digging, I found the root of the issue. But to be honest I'm not sure if my solution is the best. I just copied existing working code from https://github.com/kitodo/kitodo-presentation/blob/master/Classes/Controller/ListViewController.php#L104 and made it fit into the SearchController and the CollectionController, without understanding if it was missing for a reason. I assume that it was simply overlooked.

Now both ways to set up the search are working (https://github.com/kitodo/kitodo-presentation/issues/1029#issuecomment-1792648218) as well as the collections.

As you can see, both are working:
grafik

### DIFF
--- a/Classes/Controller/CollectionController.php
+++ b/Classes/Controller/CollectionController.php
@@ -201,10 +201,8 @@ class CollectionController extends AbstractController
 
         // get all documents of given collection
         $solrResults = null;
-        // $numResults = 0;
         if (is_array($searchParams) && !empty($searchParams)) {
             $solrResults = $this->documentRepository->findSolrByCollection($collection, $this->settings, $searchParams, $listedMetadata);
-            // $numResults = $solrResults->getNumFound();
 
             $itemsPerPage = $this->settings['list']['paginate']['itemsPerPage'];
             if (empty($itemsPerPage)) {
@@ -219,7 +217,6 @@ class CollectionController extends AbstractController
 
         $this->view->assign('viewData', $this->viewData);
         $this->view->assign('documents', $solrResults);
-        // $this->view->assign('numResults', $numResults);
         $this->view->assign('collection', $collection);
         $this->view->assign('page', $currentPage);
         $this->view->assign('lastSearch', $searchParams);
--- a/Classes/Controller/CollectionController.php
+++ b/Classes/Controller/CollectionController.php
@@ -201,11 +201,10 @@ class CollectionController extends AbstractController
 
         // get all documents of given collection
         $solrResults = null;
-        $numResults = 0;
+        // $numResults = 0;
         if (is_array($searchParams) && !empty($searchParams)) {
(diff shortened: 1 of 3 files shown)


csidirop:
As a side effect, this also seams to have some influence the issue with the gridpage:

prior: 
grafik

after: 
grafik

Unfortunately clicking on the numbers only reloads the first page. I will look further into that asap. But let me know if you sse why this happens!

Classes/Controller/CollectionController.php
```diff
@@ -198,10 +200,26 @@ public function showAction(Collection $collection): void
         $sortableMetadata = $this->metadataRepository->findByIsSortable(true);
 
         // get all documents of given collection
-        $solrResults = $this->documentRepository->findSolrByCollection($collection, $this->settings, $searchParams, $listedMetadata);
+        $solrResults = null;
+        // $numResults = 0;
```
sebastian-meyer: Please remove inactive and debug code.

Classes/Controller/CollectionController.php
```diff
@@ -198,10 +200,26 @@ public function showAction(Collection $collection): void
...
+        if (is_array($searchParams) && !empty($searchParams)) {
+            $solrResults = $this->documentRepository->findSolrByCollection($collection, $this->settings, $searchParams, $listedMetadata);
+            // $numResults = $solrResults->getNumFound();
```
sebastian-meyer: Please remove inactive and debug code.

Classes/Controller/CollectionController.php
```diff
@@ -198,10 +200,26 @@ public function showAction(Collection $collection): void
...
+            $solrPaginator = new SolrPaginator($solrResults, $currentPage, $itemsPerPage);
+            $simplePagination = new SimplePagination($solrPaginator);
+
+            $pagination = $this->buildSimplePagination($simplePagination, $solrPaginator);
+            $this->view->assignMultiple([ 'pagination' => $pagination, 'paginator' => $solrPaginator ]);
+        }
 
         $this->view->assign('viewData', $this->viewData);
         $this->view->assign('documents', $solrResults);
+        // $this->view->assign('numResults', $numResults);
```
sebastian-meyer: Please remove inactive and debug code.

sebastian-meyer (changes requested):

Classes/Controller/CollectionController.php
csidirop: Right, I forgot to ask if `numResults` should be set. But it seams not. I removed it

sebastian-meyer (approved):

PR #1090, opened 2023-11-10
[FEATURE] Enhance fulltext retrieval on virtual collections by providing database independent identifier

michaelkubina:
While logical structure elements have an information about the METS file they are referenced from, this is missing in case of physical structure elements. The currently only connecting identifier for logical and physical structures is the uid, which has several shortcomings:
- depends on the current state of the database (current uid in the typo3 document repository)
- it's not immidiate obvious from the uid, which document is associated with it (other then a PPN, Kalliope-ID, METS-path)
- if used in a virtual collection, can and will fail, if the document repository is build up again from the ground and the uid changes

This PR has the goal, that both logical and physical structures in the solr provide an information of where they come from.
- On the plus side they share a common identifier, that is independend of the typo3 document repository id.
- And with it removing an extra (lookup) step into the document repository to reference an object, when using it in virtual collections.

Otherwise fulltext retrieval in virtual collection would only work, if we query a document by its uid, which as stated lacks some comfortability and is not as static as other identifiers (like the location).


sebastian-meyer (approved):

sebastian-meyer:
Although I disagree with the idea that the METS file's location is more static than its database UID (and should therefor be the preferred identifier), I see no harm in indexing the location in addition to the primary identifier. At least it's more consistent with indexing logical structure elements and it actually can be helpful in edge-cases like dynamic collection queries.

Why I disagree with calling the file location an identifier: One of the basic principles of Kitodo.Presentation always was to be completely independent from storage implementations. That's why all file resources are always addressed as URL. In theory you should be able to migrate your files to a new location or a different infrastructure (like a repository) and would just need to update the location fields in the Kitodo database and index, but not rebuild collections, change metadata mappings or anything else.

But as I said: Your change is an improvement nevertheless, so it's of course appreciated!

PR #1091, opened 2023-11-10
[BUGFIX] The function `errorAction()` needs to exists in `NewTenantController`

beatrycze-volk:
Without this function the backend module doesn't work


sebastian-meyer:
I don't understand why there is a problem at all. `NewTenantController` extends `AbstractController` which extends Extbase's `ActionController` which in turn implements `ActionController::errorAction()`. So `NewTenantController` should've inherited this method. How can this function be missing?

beatrycze-volk:
> I don't understand why there is a problem at all. `NewTenantController` extends `AbstractController` which extends Extbase's `ActionController` which in turn implements `ActionController::errorAction()`. So `NewTenantController` should've inherited this method. How can this function be missing?

I don't know but the fact is that without this function creation of the data in backend doesn't work. The error says that this function is missing. No other more detailed message was displayed.

sebastian-meyer:
OK, but then the method should return the correct result type. Does something like `return parent::errorAction();` work? Otherwise we should at least return an empty response object to satisfy the interface.

beatrycze-volk:
Changes in the last commit are working correctly with TYPO3 11.x (https://github.com/TYPO3/typo3/blob/629eb841ea8e92399bdccd63e80f2681510665bc/typo3/sysext/extbase/Classes/Mvc/Controller/ActionController.php#L753), but with TYPO3 10.x there is an error because of incompatible return type (https://github.com/TYPO3/typo3/blob/808ef61a53bc1e7ef65b715c13727b0f2d086c24/typo3/sysext/extbase/Classes/Mvc/Controller/ActionController.php#L650).

sebastian-meyer:
The only difference seems to be that in TYPO3v10 the result of `getFlattenedValidationErrorMessage()` is returned directly, while in TYPO3v11 it's wrapped in `htmlResponse()`. Maybe we can distinguish those cases as well? Of course the return type then would have to be `string|ResponseInterface` (in PHPDocs) or `mixed` (in method signature), since PHP<8 doesn't support union types.

sebastian-meyer (approved):

sebastian-meyer:
Since Codacy and PHPStan are fine with the current state, I am as well. ;o)

PR #1100, opened 2023-11-22
[BUGFIX] Fix wrong caching backend and default lifetime

### LINKED ISSUE #884: Indexing related issue due to document persistency in cache
## Description

When deleting a document in the backend, it gets deleted from `tx_dlf_documents` and from the `solr-index`. But unbeknownst its still in the cache and can cause problems during reindexing - in our case specifically connecting documents to their parents (both in the `tx_dlf_documents` and `solr`. The only solution is to delete all caches to circumvent this issue - even if it means to delete all of the caches to just get one document properly reindexed. 

## Reproduction
e.g.
1. Have METS' where in the logical `structMap` the `mptr` is missing for the parent in the documents
2. You wont have the `partof` filled neither in `tx_dlf_documents` nor `solr`
3. Add the missing `mptr` and reindex the documents with the now corrected METS'
4. The partof is still not connected in `tx_dlf_documents` and `solr`
5. Delete the documents
6. Reindex them again and they still wont be connected
7. Delete all caches and reindex again to solve this issue

From what i see, the indexer retrieves some information partially from the document-cache. This way a deleted document even receives the same uid when reindexed without clearing all caches. So it does not have a fresh start, but instead some old data is gathered and processed.

I have not tested, what other implications this mechanism (= retrieving data from the cache) has on indexing or reindexing documents and which fields really get updated or not. But in this case we've got the defective document back from cache instead of really re-indexing the corrected one, which by itself took some time to troubleshoot.

## Expected Behavior

Deleting a document should remove it from `tx_dlf_document`, the `solr-index` and from the `cache`.

michaelkubina:
This PR fixes caching related issues, that arise due to the wrong cache backend for the caches `tx_dlf_solr` and `tx_dlf_doc`. For example, that resultsets dont get updated even if the default lifetime of the cache has passed and new documents have been indexed.

## Explanation:

The `SimpleFileBackend` does not care about expiry times:

> A caching backend which stores cache entries in files, but does not support or care about expiry times and tags. 

(see: https://api.typo3.org/10.4/class_t_y_p_o3_1_1_c_m_s_1_1_core_1_1_cache_1_1_backend_1_1_simple_file_backend.html)

This leaves once cached data permanently on disk if it is not manually removed by using "clear all caches" in the backend or a scheduler extension. The currently configured default lifetime does not get applied and has thus no function.

This PR uses the slightly more powerful `FileBackend` instead (see: https://api.typo3.org/10.4/class_t_y_p_o3_1_1_c_m_s_1_1_core_1_1_cache_1_1_backend_1_1_file_backend.html), which considers the default lifetime and updates caches accordingly (which can be tracked via the modification date of cache files on disk -> `[...]/cache/data/tx_dlf_solr`). This PR also fixes the default lifetime to an actual day (86400 seconds = 1 day).

### DIFF
--- a/ext_localconf.php
+++ b/ext_localconf.php
@@ -86,20 +86,20 @@ if (!is_array($GLOBALS['TYPO3_CONF_VARS']['SYS']['caching']['cacheConfigurations
     $GLOBALS['TYPO3_CONF_VARS']['SYS']['caching']['cacheConfigurations']['tx_dlf_solr'] = [];
 }
 if (!isset($GLOBALS['TYPO3_CONF_VARS']['SYS']['caching']['cacheConfigurations']['tx_dlf_solr']['backend'])) {
-    $GLOBALS['TYPO3_CONF_VARS']['SYS']['caching']['cacheConfigurations']['tx_dlf_solr']['backend'] = 'TYPO3\\CMS\\Core\\Cache\\Backend\\SimpleFileBackend';
+    $GLOBALS['TYPO3_CONF_VARS']['SYS']['caching']['cacheConfigurations']['tx_dlf_solr']['backend'] = 'TYPO3\\CMS\\Core\\Cache\\Backend\\FileBackend';
 }
 if (!isset($GLOBALS['TYPO3_CONF_VARS']['SYS']['caching']['cacheConfigurations']['tx_dlf_solr']['options']['defaultLifeTime'])) {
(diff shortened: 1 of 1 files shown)


michaelkubina:
I believe this is related to or the reason for " Indexing related issue due to document persistency in cache" #884 .

sebastian-meyer (approved):

PR #1102, opened 2023-11-22
[BUGFIX] Fix pagination in DFG-Viewer GridView

BFallert:
See Issue "Misbehavior in the GridView pagination" https://github.com/kitodo/kitodo-presentation/issues/1101. A detailed description can also be found there.

There are some errors in the current implementation of the pagination of the gridview (used in DFG-Viewer):

1. The links for the individual pagination pages [1,2,3,...] to the required document page [1,11,22,...] are incorrect
1. "previous page" link is incorrect
1. "next page" link ist incrorrect
1. "next page" link and "last page" link is set to deactivated too early
1. "last page" link is calculated incorrectly

These bugs will be fixed

PR #1103, opened 2023-11-23
[BUGFIX] Show only specific range of pages in pagination, add "..."

### LINKED ISSUE #1077: Pagination: pagination.hasLessPages and pagination.hasMorePages didn't exist
### Problem

In the Partial Resources/Private/Partials/Lists/Pagination.html two variables are used to insert "..." for extensive pagination. However, these variables do not currently exist. As a result, the "..." are not inserted.

### Proposed Solution
The behavior should be implemented sensibly in order to be able to display different variants of pagination.

Compare with the description of the extention mentioned in Typo3 11 [EXT:numbered_pagination](https://extensions.typo3.org/extension/numbered_pagination)

### Preferred variant
< 1 2 ... 21 22 23 24 ... 100 >

### Option 1
[1] 2 3 ... next
1 [2] 3 ... next
prev ... 2 [3] 4 ... next
prev ... 3 [4] 5
prev ... 3 4 [5]

### Option 2
[1] 2 3 ... last
1 [2] 3 ... last
first ... 2 [3] 4 ... last
first ... 3 [4] 5
first ... 3 4 [5]

### Option 3
`|<` means first
`<` means prev
`>` means next
`>|` means last

[1] 2 3 ... `>` `>|`
1 [2] 3 ... `>` `>|`
`|<` `<` ... 2 [3] 4 ... `>` `>|`
`|<` `<` ... 3 [4] 5
`|<` `<` ... 3 4 [5]

Option with possibly deactivated `>` and `>|` as well as `|<` and `<` when the last or first element has been reached

## Additional Context

In the documentation for Typo3 11 ([https://docs.typo3.org/m/typo3/reference-coreapi/11.5/en-us/ApiOverview/Pagination/Index.html](https://docs.typo3.org/m/typo3/reference-coreapi/11.5/en-us/ApiOverview/Pagination/Index.html)) an extension is mentioned that solves the problem with the dots. But this work only with the old pagination, so it is no option (as i cab see).

And in the documentation for Typo3 12 [https://docs.typo3.org/m/typo3/reference-coreapi/12.4/en-us/ApiOverview/Pagination/Index.html#sliding-window-pagination](https://docs.typo3.org/m/typo3/reference-coreapi/12.4/en-us/ApiOverview/Pagination/Index.html#sliding-window-pagination) it is mentioned how this problem can be solved with Typo3 version 12. 

The description of the behavior, in Typo3 12, is taken verbatim from the extension:

_**Sliding window pagination**_
 _New in version 12.0._

_The sliding window pagination can be used to paginate array items or query results from Extbase. The main advantage is that it reduces the amount of pages shown._

_Example: Imagine 1000 records and 20 items per page which would lead to 50 links. Using the SlidingWindowPagination, you will get something like < 1 2 ... 21 22 23 24 ... 100 >._

But at the moment, I don't see how that could benefit us at the moment.

Source code for the _**Sliding window pagination**_ from Typo3 V12 [[_**Sliding window pagination**_](https://github.com/TYPO3/typo3/blob/main/typo3/sysext/core/Classes/Pagination/SlidingWindowPagination.php)](https://github.com/TYPO3/typo3/blob/main/typo3/sysext/core/Classes/Pagination/SlidingWindowPagination.php)

BFallert:
With extensive list views or gridviews, a large number of page buttons are displayed. Depending on the layout, these do not fit into the respective frame. 

Currently, these adjustments display a maximum of 5 additional page buttons next to the active button. If there are more pages before or after, "..." is added in each case.

image

This PR extends the PR https://github.com/kitodo/kitodo-presentation/pull/1102 and would close the issue https://github.com/kitodo/kitodo-presentation/issues/1077


Resources/Private/Partials/Lists/Pagination.html
```diff
@@ -18,11 +18,14 @@
...
-        <li>…</li>
-    </f:if>
-    <f:comment>add pages between first and last page</f:comment>
-    <f:for each="{pagination.pagesG}" as="page">
+    <f:comment>is no longer needed, is provided in pagination.pagesR</f:comment>
+    <f:comment>
+        <f:if condition="{pagination.hasLessPages}">
+            <li>…</li>
+        </f:if>
+    </f:comment>
```
sebastian-meyer: If it is no longer needed (and working), please remove it entirely.

Resources/Private/Partials/Lists/Pagination.html
```diff
@@ -37,15 +40,28 @@
...
     </f:for>
-    <f:if condition="{pagination.hasMorePages}">
-        <li>…</li>
-    </f:if>
+    <f:comment>is no longer needed, is provided in pagination.pagesR</f:comment>
+    <f:comment>
+        <f:if condition="{pagination.hasMorePages}">
+            <li>…</li>
+        </f:if>
+    </f:comment>
```
sebastian-meyer: If it is no longer needed (and working), please remove it entirely.

sebastian-meyer (changes requested):
I am not a fan of all the calculations in the controller. I feel like it would be more logical to just provide the necessary variables there (`$nRange`) and do all the calculations in the Fluid template (because it's about visualization, not program logic). And yes, as you propose in the comments, we should make it configurable.

Resources/Private/Partials/Lists/Pagination.html
BFallert: done

Resources/Private/Partials/Lists/Pagination.html
BFallert: done

sebastian-meyer (approved):

Issue #1106, opened 2023-11-28
CollectionView shows ALL indexed documents instead of only the selected


csidirop:
## Description

Setting the collection plugin without a ListView causes all documents to be shown instead of just the selected ones:

Left: Collection -> Listview -> PageView
Right: Collection -> PageView
![grafik](https://github.com/kitodo/kitodo-presentation/assets/43964592/937e6816-5aed-4e2a-914e-1548bf47793b)

This error exists even before the TYPO3 v11 merge.

## Reproduction

Steps to reproduce the behaviour:

1. Index some documents
2. Create two pages with Collection Plugins, one with a ListView and one with a PageView plugin
3. Setup the first Collection plugin to use the site with the PageView directly, the second one uses the ListView which then uses the PageView
4. See behavior

## Expected Behavior

Only documents in given collection should be shown, exactly like in the ListView.

## Screenshots and Examples

see above

## Environment

 -   OS version: Debian (in Docker)
 -   RDBMS version: Maria DB (MySQL 5.5.5-10.11.2-MariaDB-1:10.11.2+maria~ubu2204)
 -   Apache Solr version: -
 -   TYPO3 version: 10.4.38-dev
 -   PHP version: 7.4.33

csidirop:
I searched a bit and found out the following: 

If we compare the `$collection` var from

https://github.com/kitodo/kitodo-presentation/blob/0e0c77f1d9fcc8a236eacb3e191ad46e133bb45c/Classes/Controller/ListViewController.php#L78

and 

https://github.com/kitodo/kitodo-presentation/blob/0e0c77f1d9fcc8a236eacb3e191ad46e133bb45c/Classes/Controller/CollectionController.php#L168

we see, that they are identical except that the first one has its `Kitodo\Dlf\Domain\Model\Collection` object in an array. If we encapsulating the collection object in an array `$collectionArray[] = $collection;` and replacing all references only the correct documents are shown:

![grafik](https://github.com/kitodo/kitodo-presentation/assets/43964592/1c03c3bf-792d-4935-9f38-97e3d37ec057)

Obviously this is just a workaround but I couldn't find the source of the call and why it differs. Maybe you can help me out.

csidirop:
In the showAction function the search parameter are build from the url query parameter:
https://github.com/kitodo/kitodo-presentation/blob/0e0c77f1d9fcc8a236eacb3e191ad46e133bb45c/Classes/Controller/CollectionController.php#L170

[...]

PR #1108, opened 2023-12-05
[BUGFIX] Fix parameter type for `$collections` in `CollectionController`

### LINKED ISSUE #1107: [BUG] Error in `CollectionController`
## Description

After #1069 my collection pages are no longer loading and giving this error:

`Uncaught TYPO3 Exception: Argument 1 passed to Kitodo\Dlf\Controller\CollectionController::processCollections() must be of the type array, object given, called in /var/www/typo3/public/typo3conf/ext/dlf/Classes/Controller/CollectionController.php on line 104 | TypeError thrown in file /var/www/typo3/public/typo3conf/ext/dlf/Classes/Controller/CollectionController.php in line 214.`

The ` $collections` Array gets overwritten by the return value of `collectionRepository->findAll()`:
https://github.com/kitodo/kitodo-presentation/blob/10248c53fd34bdd4bde3310a6bc8ccba46fe09d2/Classes/Controller/CollectionController.php#L97

Commenting out the line no longer causes the error, but no objects are loaded either.

Reverting the changes of #1069  is fixing the issue.

beatrycze-volk:
Fix for https://github.com/kitodo/kitodo-presentation/issues/1107

### DIFF
--- a/Classes/Common/Solr/SolrSearch.php
+++ b/Classes/Common/Solr/SolrSearch.php
@@ -6,7 +6,6 @@ use Kitodo\Dlf\Common\AbstractDocument;
 use Kitodo\Dlf\Common\Helper;
 use Kitodo\Dlf\Common\Indexer;
 use Kitodo\Dlf\Common\Solr\SearchResult\ResultDocument;
-use Kitodo\Dlf\Domain\Model\Collection;
 use Kitodo\Dlf\Domain\Repository\DocumentRepository;
 use Solarium\QueryType\Select\Result\Document;
 use TYPO3\CMS\Core\Cache\CacheManager;
@@ -38,8 +37,9 @@ class SolrSearch implements \Countable, \Iterator, \ArrayAccess, QueryResultInte
 
     /**
      * @access private
-     * @var QueryResult|Collection|null
+     * @var array|null
      */
+    // TODO: confusing naming, here is passed array of collections and it is used as array in prepare()
     private $collection;
 
     /**
@@ -84,7 +84,7 @@ class SolrSearch implements \Countable, \Iterator, \ArrayAccess, QueryResultInte
      * @access public
      *
      * @param DocumentRepository $documentRepository
-     * @param QueryResult|Collection|null $collection
+     * @param array|null $collection
      * @param array $settings
      * @param array $searchParams
      * @param QueryResult $listedMetadata
@@ -433,6 +433,7 @@ class SolrSearch implements \Countable, \Iterator, \ArrayAccess, QueryResultInte
         }
 
         // if collections are given, we prepare the collection query string
+        // TODO: this->collection should not be actually called collections?
         if ($this->collection) {
             $collectionsQueryString = '';
             $virtualCollectionsQueryString = '';
--- a/Classes/Controller/CollectionController.php
+++ b/Classes/Controller/CollectionController.php
@@ -253,7 +253,7 @@ class CollectionController extends AbstractController
 
             // Generate random but unique array key taking amount of documents into account.
             do {
-                $key = (count($collection['priority']) * 1000) + random_int(0, 1000);
+                $key = ($collection['priority'] * 1000) + random_int(0, 1000);
             } while (!empty($processedCollections[$key]));
 
             $processedCollections[$key]['collection'] = $collection;
--- a/Classes/Controller/CollectionController.php
+++ b/Classes/Controller/CollectionController.php
@@ -253,7 +253,12 @@ class CollectionController extends AbstractController
 
             // Generate random but unique array key taking amount of documents into account.
             do {
-                $key = ($collection['priority'] * 1000) + random_int(0, 1000);
+                if ($collection instanceof Collection) {
+                    $key = ($collection->getPriority() * 1000) + random_int(0, 1000);
+                }
+                if (is_array($collection)) {
+                    $key = ($collection['priority'] * 1000) + random_int(0, 1000);
+                }
             } while (!empty($processedCollections[$key]));
 
             $processedCollections[$key]['collection'] = $collection;
--- a/Classes/Controller/CollectionController.php
+++ b/Classes/Controller/CollectionController.php
@@ -206,12 +206,12 @@ class CollectionController extends AbstractController
      *
      * @access private
      *
-     * @param array $collections to be processed
+     * @param QueryResultInterface|array|object $collections to be processed
      * @param Solr $solr for query
      *
      * @return array
      */
-    private function processCollections(array $collections, Solr $solr): array
+    private function processCollections($collections, Solr $solr): array
     {
         $processedCollections = [];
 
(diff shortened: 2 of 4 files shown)


csidirop:
I'm getting following error: ` Uncaught TYPO3 Exception: Cannot use object of type Kitodo\Dlf\Domain\Model\Collection as array | Error thrown in file /var/www/typo3/public/typo3conf/ext/dlf/Classes/Controller/CollectionController.php in line 256`

csidirop:
Not sure if this helps, but the error occurs regardless of whether the collections plugin points to a list plugin or a page plugin

beatrycze-volk:
Weird, I have tested it by myself ad could without any problem display collection list and list of documents inside the collection.

csidirop:
Interesting, can it be some differences in the configuration?

beatrycze-volk:
That's possible - here is a ddev system which I have used to test those changes: https://github.com/slub/ddev-sachsendigital

csidirop:
Tested again and have the same error (https://github.com/kitodo/kitodo-presentation/pull/1108#issuecomment-1845001029). I get this error regardless of whether the collection plugin calls the page view plugin directly or via separate list view.

https://github.com/kitodo/kitodo-presentation/blob/7432a44716dc7def6873d5218864dfcefae1c207/Classes/Controller/CollectionController.php#L256

Why do you think that `$collection` is an array? I cant follow you here.
I changed the call to `$key = ($collection->getPriority() * 1000) + random_int(0, 1000);` which fixes the issue here (for me).

With this change both collection pages are loading again. But again only the "indirect" one shows the correct result. The other one shows all indexed documents.

A var_dump of `$this->collection` in `SolrSearch::prepare()`:
https://github.com/kitodo/kitodo-presentation/blob/25639525814f0fcbf0b07d922314a563dd2cbe9f/Classes/Common/Solr/SolrSearch.php#L436
shows that the object is encapsulated in an array again for the one case. Comparable to https://github.com/kitodo/kitodo-presentation/issues/1106#issuecomment-1830325863

beatrycze-volk:
> Why do you think that `$collection` is an array? I cant follow you here.

```
if ($this->settings['collections']) {
    $sortedCollections = [];
    foreach (GeneralUtility::intExplode(',', $this->settings['collections']) as $uid) {
        $sortedCollections[$uid] = $this->collectionRepository->findByUid($uid);
    }
    $collections = $sortedCollections;
} else {
    $collections = $this->collectionRepository->findAll();
}
```
Funxction `findByUid($uid)` returns `object|null`, but `findAll()` returns `QueryResultInterface|array`. So depending on the called function the returned collection can be `array` or `object` (it can be also `null` if doesn't exists).

csidirop:
Okay right, thanks! So we can simply unify this so that an array is always used at this point? If `findByUid()` returns an object, just push it into an array.

csidirop:
Can you tell me where `showAction()` is called from? I believe somewhere up the call stack will be a junction like your example where we get an object for one case and and array for the other case.

beatrycze-volk:
> Can you tell me where `showAction()` is called from? I believe somewhere up the call stack will be a junction like your example where we get an object for one case and and array for the other case.

`showAction()` is called fromList.html template. There are 2 calls: https://github.com/kitodo/kitodo-presentation/blob/8fe3ca79f1af9e42974176365dc688c2bd01d7c0/Resources/Private/Templates/Collection/List.html#L20 and https://github.com/kitodo/kitodo-presentation/blob/8fe3ca79f1af9e42974176365dc688c2bd01d7c0/Resources/Private/Templates/Collection/List.html#L41

beatrycze-volk:
> Okay right, thanks! So we can simply unify this so that an array is always used at this point? If `findByUid()` returns an object, just push it into an array.

I don't think that making such a conversion in this place makes sense. I would stay with the type check.

csidirop:
Okay, then I can't find anything else to suggest. Is there anything else that you can think of?

beatrycze-volk:
> Okay, then I can't find anything else to suggest. Is there anything else that you can think of?

Does it work for you with my type check or does the problem still appear?

csidirop:
> > Okay, then I can't find anything else to suggest. Is there anything else that you can think of?
> 
> Does it work for you with my type check or does the problem still appear?

Unfortunately it still appears. Btw. I am trying to look into your sachsendigital ddev, but I have had a few problems so far.

beatrycze-volk:
> > > Okay, then I can't find anything else to suggest. Is there anything else that you can think of?
> > 
> > 
> > Does it work for you with my type check or does the problem still appear?
> 
> Unfortunately it still appears.

In which line? Does it ignore the check for collection type?

csidirop:
> In which line?

There is no error or exception anymore. It just shows all indexed documents again.

> Does it ignore the check for collection type?

If you are [referring to this check in `processCollections()`](https://github.com/beatrycze-volk/kitodo-presentation/blob/984ac7efebb106e38de0876676596ddde8d9aaac/Classes/Controller/CollectionController.php#L257), then no it does not ignore it. 
`processCollections()` is called six times (for my six collections?) and every time `$collection instanceof Collection` is true and the block is triggered. There is no run where nether the collection nor the array check is true.

csidirop:
Not sure if that matters, but regardless whether I use the not-working "direct" or working "indirect" (with ListView) implementation, `$collection` is every time a `Collection` type.

But my [workaround ](https://github.com/kitodo/kitodo-presentation/issues/1106#issuecomment-1830325863) is still working. `showAction()` is only called in that "direct" implementation, while the "indirect" uses ListViewControllers `mainAction()`.

csidirop:
I managed to reproduce this error in your ddev environment (with your fix-type branch):

collage

beatrycze-volk:
Thank you for extensive testing! :)

As a summary:
With my change the error is not appearing but also the list of collections is not displayed, right? I will look into it and write you back.

csidirop:
> As a summary: With my change the error is not appearing but also the list of collections is not displayed, right? I will look into it and write you back.

Right I think. The TYPO3 Exception is fixed. But the issue that ALL documents are shown instead of just the ones in the selected collection still exists (like before).

beatrycze-volk:
@sebastian-meyer The issues https://github.com/kitodo/kitodo-presentation/issues/1106 and https://github.com/kitodo/kitodo-presentation/issues/1107 should be fixed by this changes but the problem of confusion between `collection` and `collections` in `SolrSearch` class should be tackled.

1. Should function `findSolrByCollection` search by no collection, one or many than maybe it should be renamed to `findSolrByCollections`? Or should we add to DocumentRepository second function and call one `findSolrByCollection` and second `findSolrByCollections`. First would take as parameter `Collection` object and second an `array` of `Collection` objects.
 
2. `$collection` attribute of `SolrSearch` class is clearly an array of collections. Should it be renamed to match its real use or was there some reason for this name?

csidirop (approved):
Collections working again for me. Also tested the search just in case, which works too.

Code looks good to me.

sebastian-meyer (approved):

PR #1116, opened 2023-12-18
[BUGFIX] Fix scrolling of text in #tx-dlf-fulltextselection to the text under the cursor

BFallert:
Fixed the scoll function for fulltext. When the mouse is positioned on a text passage, the text in the configured div scrolls to the respective fulltext line.

- Adjust the default setting of `settings.full-text-scroll-element` to `#tx-dlf-fulltextselection`, instead of `html, body`
- Calculating the scroll positions copied from `ddb_kitodo_zeitungsportal`
- `full-text-scroll-element` now consistently defines the ID of the scroll container. The previously used fixed value `#tx-dlf-fulltextselection` is now only available as a default value.

image

### DIFF
--- a/Resources/Public/JavaScript/PageView/FulltextControl.js
+++ b/Resources/Public/JavaScript/PageView/FulltextControl.js
@@ -336,6 +336,16 @@ dlfViewerFullTextControl.prototype.addActiveBehaviourForSwitchOff = function() {
 };
 
 /** 
+ * Recalculate position of text lines if full text container was resized
+ */
+dlfViewerFullTextControl.prototype.onResize = function() {
+    if (this.element != undefined && this.element.css('width') != this.lastHeight) {
+        this.lastHeight = this.element.css('width');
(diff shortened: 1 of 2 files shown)


sebastian-meyer (changes requested):
Apart from the CodeQL issues this is fine! Please check the automated comments above.

sebastian-meyer (approved):

csidirop:
This PR somehow broke my viewer. When opening a document with a fulltext it just shows "Volltext wird geladen"

grafik

Reverting this fixes that issue. In detail its the changes in [FulltextControl.js](https://github.com/kitodo/kitodo-presentation/pull/1116/files#diff-23b1fd74dbf112aa8dd3fbfcfe1529e09f9f936923b7b5943e763441fd2bee39). I have not yet looked into these changes, but I will as soon I found some time.

But I believe that the DFG-Viewer could be the reason why you didn't notice it. I did not test it with the presentation viewer.

PR #1117, opened 2023-12-21
[FEATURE] Add support for ALTO v3 and v4

michaelkubina:
This PR makes the ALTO Parser agnostic of the actual ALTO schema version in the provided file. It checks the document namespaces and registers it accordingly. This way we can index ALTO files in the schema versions 2, 3 and 4.

I have testet it in our development enviroment with SOLR 9.4.0 and the ocr-highlighting module in version 0.8.3. I have not encountered any issues here. It has been rolled out in our productive enviroment as well.

While it does not resolve the general namespace issue, it does at least resolve the ALTO issue.

This is related to #488 .


Classes/Format/Alto.php
```diff
@@ -36,7 +36,19 @@ class Alto implements \Kitodo\Dlf\Common\FulltextInterface
...
+        if (in_array('http://www.loc.gov/standards/alto/ns-v2#', $namespace, true)) {
+            $xml->registerXPathNamespace('alto', 'http://www.loc.gov/standards/alto/ns-v2#');
+        }
+        if (in_array('http://www.loc.gov/standards/alto/ns-v3#', $namespace, true)) {
+            $xml->registerXPathNamespace('alto', 'http://www.loc.gov/standards/alto/ns-v3#');
+        }
+        if (in_array('http://www.loc.gov/standards/alto/ns-v4#', $namespace, true)) {
+            $xml->registerXPathNamespace('alto', 'http://www.loc.gov/standards/alto/ns-v4#');
+        }
+
```
csidirop:
```suggestion
        if (in_array('http://www.loc.gov/standards/alto/ns-v2#', $namespace, true)) {
            $xml->registerXPathNamespace('alto', 'http://www.loc.gov/standards/alto/ns-v2#');
        } elseif (in_array('http://www.loc.gov/standards/alto/ns-v3#', $namespace, true)) {
            $xml->registerXPathNamespace('alto', 'http://www.loc.gov/standards/alto/ns-v3#');
        } elseif (in_array('http://www.loc.gov/standards/alto/ns-v4#', $namespace, true)) {
            $xml->registerXPathNamespace('alto', 'http://www.loc.gov/standards/alto/ns-v4#');
        }

```

Classes/Format/Alto.php
```diff
@@ -68,7 +80,17 @@ public function getRawText(\SimpleXMLElement $xml): string
...
+        $namespace = $xml->getDocNamespaces();
+        if (in_array('http://www.loc.gov/standards/alto/ns-v2#', $namespace, true)) {
+            $xml->registerXPathNamespace('alto', 'http://www.loc.gov/standards/alto/ns-v2#');
+        }
+        if (in_array('http://www.loc.gov/standards/alto/ns-v3#', $namespace, true)) {
+            $xml->registerXPathNamespace('alto', 'http://www.loc.gov/standards/alto/ns-v3#');
+        }
+        if (in_array('http://www.loc.gov/standards/alto/ns-v4#', $namespace, true)) {
+            $xml->registerXPathNamespace('alto', 'http://www.loc.gov/standards/alto/ns-v4#');
+        }
```
csidirop:
```suggestion
        if (in_array('http://www.loc.gov/standards/alto/ns-v2#', $namespace, true)) {
            $xml->registerXPathNamespace('alto', 'http://www.loc.gov/standards/alto/ns-v2#');
        } elseif (in_array('http://www.loc.gov/standards/alto/ns-v3#', $namespace, true)) {
            $xml->registerXPathNamespace('alto', 'http://www.loc.gov/standards/alto/ns-v3#');
        } elseif (in_array('http://www.loc.gov/standards/alto/ns-v4#', $namespace, true)) {
            $xml->registerXPathNamespace('alto', 'http://www.loc.gov/standards/alto/ns-v4#');
        }
```

csidirop:
I would suggest adding a feedback for future "unsupported" versions. This could be a log entry or writing direct to the command line (via the SymfonyStyle). This way at least we know that there is something off.

Proposal: Add an `else`-block which catches all other cases and handle some form of feedback.

csidirop (changes requested):

Classes/Format/Alto.php
```diff
@@ -36,7 +36,19 @@ class Alto implements \Kitodo\Dlf\Common\FulltextInterface
     public function getRawText(\SimpleXMLElement $xml): string
     {
         $rawText = '';
-        $xml->registerXPathNamespace('alto', 'http://www.loc.gov/standards/alto/ns-v2#');
+
```
stweil:
```suggestion
 
```

Classes/Format/Alto.php
stweil: What about adding a private function which registers the namespace and avoids the duplication of code?

BFallert:
I have testet it in our development enviroment with SOLR 8.11.2.
I have tested documents with ALTO 2.x / 3.x / 4.x.
All tests were successful. The data arrived in the solr dataset in all cases.

michaelkubina:
as proposed by @stweil i have moved the namespace registration to a private function to reduce code duplication.

...

> I would suggest adding a feedback for future "unsupported" versions. This could be a log entry or writing direct to the command line (via the SymfonyStyle). This way at least we know that there is something off.

> Proposal: Add an else-block which catches all other cases and handle some form of feedback.

while i like this idea, i currently have no time to look into the depths of how to achieve this. but we should keep that in mind for the future. looking at what ocr engines output currently, we cover the necessary schemas

stweil (approved):
Thanks!

sebastian-meyer (approved):

PR #1123, opened 2024-01-10
[BUGFIX] Fix and improve SolrSearch class

beatrycze-volk:
- rename `$collection` attribute to `$collections` in `SolrSearch` class
- extract two smaller functions from `prepare()` function
- split function` findSolrByCollection()` onto three different functions

Depends on https://github.com/kitodo/kitodo-presentation/pull/1108


sebastian-meyer (approved):

sebastian-meyer (changes requested):
Please see the PHPStan report: There is still one call to `DocumentRepository::findSolrByCollection()` left, which uses the old method's signature (in `CollectionController.php`).

sebastian-meyer (approved):

PR #1135, opened 2024-01-16
[MAINTENANCE] Improvements in MetadataController class

beatrycze-volk:


Classes/Controller/MetadataController.php
```diff
@@ -185,7 +185,7 @@ protected function printMetadata(array $metadata): void
             $this->view->assign('documentMetadataSections', $metadata);
             $this->view->assign('configMetadata', $metadataResult);
             $this->view->assign('separator', $this->settings['separator']);
-            $this->view->assign('metaCObjData', $this->buildMetaCObjData($metadata));
+            $this->view->assign('metaConfigObjectData', $this->buildMetaConfigObjectData($metadata));
```
sebastian-meyer, line 188: In the TYPO3 universe "cObj" actually is a common abbreviation for "content object". But yes, we could call it a "config object" as well, because it's not a real cObj anyways...

sebastian-meyer (approved):

PR #1138, opened 2024-01-17
[MAINTENANCE] Improvements in AbstractDocument class

beatrycze-volk:


Classes/Common/AbstractDocument.php
```diff
@@ -590,7 +590,7 @@ public static function &getInstance(string $location, array $settings = [], bool
             $instance = new IiifManifest($pid, $location, $iiif);
         }
 
-        if (!is_null($instance)) {
+        if ($instance != null) {
```
sebastian-meyer: Please always use strict comparisons!

```suggestion
        if ($instance !== null) {
```

sebastian-meyer (changes requested):

sebastian-meyer (approved):

PR #1140, opened 2024-01-18
[BUGFIX] Fix displaying pages for images in grid view

beatrycze-volk:
Previously instead of page numbers were displayed ' - '.
page

### DIFF
--- a/Classes/Controller/AbstractController.php
+++ b/Classes/Controller/AbstractController.php
@@ -329,6 +329,7 @@ abstract class AbstractController extends ActionController implements LoggerAwar
      * @param PaginatorInterface $paginator
      * @return array
      */
+    //TODO: clean this function
     protected function buildSimplePagination(PaginationInterface $pagination, PaginatorInterface $paginator): array
     {
         $firstPage = $pagination->getFirstPageNumber();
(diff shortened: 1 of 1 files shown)


sebastian-meyer:
I think that's not a bug, but a feature. There is a difference between `page` and `pagination`:

`page` is the image number, always an unique integer starting with 1 and continuously counted. In METS it's the value of the `@ORDER` attribute.

`pagination` is the page number as it is printed on the book. This can be any integer, string or even missing. It also isn't necessarily unique. In METS it's the value of the `@ORDERLABEL` attribute.

So my guess is, that the grid view is supposed to show the *page* numbers and displays "-" for pages which don't have one. Your change will now show the *image* number instead (which is fine for me, but I am not sure if this is intended).

sebastian-meyer (dismissed):
From a technical point of view this is fine! But please verify that this change really is intentional (see above).

csidirop:
Can we both selectable in the backend? I think there is a lot of scope here for how you would like something like this to be displayed.

beatrycze-volk:
> I think that's not a bug, but a feature. There is a difference between `page` and `pagination`:
> 
> `page` is the image number, always an unique integer starting with 1 and continuously counted. In METS it's the value of the `@ORDER` attribute.
> 
> `pagination` is the page number as it is printed on the book. This can be any integer, string or even missing. It also isn't necessarily unique. In METS it's the value of the `@ORDERLABEL` attribute.
> 
Oh, then shouldn't be this somehow passed to the people which are working on those systems? I have got error report about not working pages (https://www.digas.sachsen.de/werkansicht?tx_dlf%5Bdouble%5D=0&tx_dlf%5Bid%5D=7391&tx_dlf%5Bpage%5D=1&tx_dlf%5Bpagegrid%5D=1&cHash=8c785d23d941f7d0ea8f7bb872f4ff26, https://www.digas.sachsen.de/digas/LfULG_EB02521-00053306/LfULG_EB02521-00053306_mets.xml). Indeed order labels are used:

```
<mets:div ID="PHYS_0001" ORDER="1" ORDERLABEL=" - " TYPE="page">
  <mets:fptr FILEID="FILE_0001_FULLTEXT"/>
   ..
  <mets:fptr FILEID="FILE_0001_ORIGINAL"/>
</mets:div>
```

> So my guess is, that the grid view is supposed to show the _page_ numbers and displays "-" for pages which don't have one. Your change will now show the _image_ number instead (which is fine for me, but I am not sure if this is intended).

After your explanation I'm also not sure if it is intended.

PR #1149, opened 2024-01-26
[FEATURE] Implement document validation before document is saved to database

beatrycze-volk:
Allow index documents only if their metadata contains document format and record identifier

This functionality can be extended for more mandatory fields without which document should be treated as incomplete and as such should not be inserted into database and SOLR Index.

### DIFF
--- a/Classes/Command/BaseCommand.php
+++ b/Classes/Command/BaseCommand.php
@@ -22,6 +22,7 @@ use Kitodo\Dlf\Domain\Repository\StructureRepository;
 use Kitodo\Dlf\Domain\Model\Collection;
 use Kitodo\Dlf\Domain\Model\Document;
 use Kitodo\Dlf\Domain\Model\Library;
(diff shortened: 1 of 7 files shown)


Classes/Validation/DocumentValidator.php
sebastian-meyer, line 1: Could you please fix the Codacy issues?

Classes/Command/BaseCommand.php
sebastian-meyer, line 1: There are some merge conflicts. Could you please resolve?

sebastian-meyer (changes requested):

sebastian-meyer (dismissed):

sebastian-meyer:
Following #1150 the config option needs to be renamed from `requiredMetadataFields` to `general.requiredMetadataFields`. I already did it in `ext_conf_template.txt` and the localization files while updating from `master`, but it still has to be done in the code.

sebastian-meyer (approved):

PR #1163, opened 2024-02-03
[MAINTENANCE] Fix some typos (most of them found by codespell)

stweil:


sebastian-meyer (approved):

sebastian-meyer:
Thank you for the corrections! I'll merge them all, although some of the changes (i. e. those to `jPlayer`) would need to be fixed upstream to not be reintroduced with the next dependency updates.

stweil:
The last release of jPlayer was ten years ago, the latest commit is from 2017, and my pull request to fix the typos (https://github.com/jplayer/jPlayer/pull/415) is from 2018. Therefore I don't expect that anything will be fixed there upstream.

I wonder whether we should look for a replacement with recent code. Sticking to ten year old software with two high severity and four moderate severity Dependabot alerts might be a bad idea.

PR #1164, opened 2024-02-03
[MAINTENANCE] Document function magicGetMetadataArray as a "magic" function

stweil:


stweil:
Removing `magicGetMetadataArray` causes one test case to fail:
```
There was 1 failure:

1) Kitodo\Dlf\Tests\Functional\Common\SolrIndexingTest::canIndexAndSearchDocument
Failed asserting that 0 matches expected 1.
```

So it is still used, although that is not so obvious.

stweil:
Okay, I found the reason. The function is one of the magic functions where it is not obvious that it is used. I updated the commit to add a comment which describes that.

sebastian-meyer (approved):

PR #1169, opened 2024-02-08
[BUGFIX] Fix problems with translation of database records

### LINKED ISSUE #1147: [BUG] Metadata and structure translation entries
## Description

There is something off with the metadata created by the new tenant controller. I first stumbled across this, when I tried to edit a metadata entry and TYPO3 gave me an info  message that `: The value of the field "index_name" has been changed from "opac_id" to "opac_id0" as it is required to be unique.` And this holds for every metadata entry. 

So it seams, that in between the v3.3.x and now something broke. 
Here in comparison some screenshots:

Overview: (note the indentations indicating a translation of the entry instead of a separate one)
overview

Example:
title

Example translation: (apart from the obvious difference, you can see that the translation in v4.x has the same index_name as the original)
title-translation

This are exemplary DB entries for Type/Strukturtyp:
```
# v3.3.x:
INSERT INTO `tx_dlf_metadata` VALUES (28,3,1706263174,1706263174,2,0,0,0,'a:16:{s:5:\"label\";N;s:10:\"index_name\";N;s:15:\"index_tokenized\";N;s:12:\"index_stored\";N;s:13:\"index_indexed\";N;s:11:\"index_boost\";N;s:11:\"is_sortable\";N;s:8:\"is_facet\";N;s:9:\"is_listed\";N;s:18:\"index_autocomplete\";N;s:6:\"format\";N;s:13:\"default_value\";N;s:4:\"wrap\";N;s:16:\"sys_language_uid\";N;s:6:\"hidden\";N;s:6:\"status\";N;}',1,1026,NULL,'Strukturtyp','type',0,'','key.wrap = <dt class=\"tx-dlf-type\">|</dt>\nvalue.required = 1\nvalue.wrap = <dd class=\"tx-dlf-type\">|</dd>',0,1,1,1,0,0,1,0,0);
INSERT INTO `tx_dlf_metadata` VALUES (29,3,1706263174,1706263174,2,0,1,30,'a:3:{s:16:\"sys_language_uid\";N;s:11:\"l18n_parent\";N;s:6:\"format\";N;}',0,1025,NULL,'Context','context0',1,'','key.wrap = <dt>|</dt><dd>\nvalue.required = 1\nvalue.noTrimWrap = || &gt; |\nall.substring = 0,-6\nall.wrap = |</dd>',0,1,1,1,0,0,1,0,0);

```

```
# v4.x:
INSERT INTO `tx_dlf_metadata` VALUES (1,3,1706264096,1706264096,0,0,1,'',0,0,2,'Type','type',0,'','key.wrap = <dt>|</dt>\nvalue.required = 1\nvalue.wrap = <dd>|</dd>',0,0,0,1,0,0,0,0,0);
INSERT INTO `tx_dlf_metadata` VALUES (2,3,1706264096,1706264096,0,0,0,'',0,0,0,'Strukturtyp','type',2,'','key.wrap = <dt>|</dt>\nvalue.required = 1\nvalue.wrap = <dd>|</dd>',0,1,0,1,1,1,1,0,0);
```

And here an overview where you can clearly see the difference:
collage
What you also can see is, that the Data Formats differ extremely! On some cases there are multiple MODS data format entries and on most there different entries for the translations.

beatrycze-volk:
- Restore missing connection between `transOrigPointerField` and `l18n_parent` (after that problem with unique `index_name ` disappears so no need to append 0)
- Add missing database field

Problem with the differing formats will be resolved in separate PR


csidirop:
At first glance it looked like everything would work, but when you rename the label of an entry TYPO3 throws you two exceptions:
```
(1/2) Doctrine\DBAL\Exception\InvalidFieldNameException

An exception occurred while executing 'SELECT `uid`, `l10n_state`, `sys_language_uid`, `l18n_parent` FROM `tx_dlf_metadata` WHERE (`uid` IN (?)) AND (`tx_dlf_metadata`.`deleted` = 0)' with params [4]: Unknown column 'l10n_state' in 'field list'
```
and 
```
(2/2) Doctrine\DBAL\Driver\Mysqli\Exception\ConnectionError

Unknown column 'l10n_state' in 'field list'

```

A quick check on the DB showed that like the exception says, there is no 'l10n_state' column in the tx_dlf_metadata table. Can we add it.

sebastian-meyer (approved):

stweil:
This commit adds a field which should be removed according to the TCA migrations check (tested with TYPO3 11 and 12), so it looks like it was not the right fix:

```
Check TCA Migrations

Checks whether the current TCA needs migrations and displays the new migration paths which need to be adjusted manually.
TCA migrations need to be applied
Check the following list and apply needed changes.
The 'tx_dlf_collections' TCA tables transOrigPointerField 'l18n_parent' is defined as excluded field which is no longer needed and should therefore be removed.
The 'tx_dlf_libraries' TCA tables transOrigPointerField 'l18n_parent' is defined as excluded field which is no longer needed and should therefore be removed.
The 'tx_dlf_metadata' TCA tables transOrigPointerField 'l18n_parent' is defined as excluded field which is no longer needed and should therefore be removed.
The 'tx_dlf_structures' TCA tables transOrigPointerField 'l18n_parent' is defined as excluded field which is no longer needed and should therefore be removed.
[...]
```

According to https://docs.typo3.org/m/typo3/reference-tca/main/en-us/BestPractises/LanguageFields.html `l10n_parent`  instead of `l18n_parent` might be the correct name.

PR #1171, opened 2024-02-13
[MAINTENANCE] Suppress Codacy warnings for private methods...

stweil:
…  which are only used in a template


stweil:
The comments which are added in this commit hopefully suppress the Codacy warnings, but this cannot be tested in my personal fork.

sebastian-meyer (approved):

PR #1175, opened 2024-02-14
[SECURITY] Update dependencies to fix security issues (4.x branch)

sebastian-meyer:
This is kind of a double-edged sword: In order to enforce using only the newest TYPO3 ELTS versions, we would need to put those versions as dependencies in `composer.json`. But then all tests would fail, because those versions are not publicly available via `packagist.org`. On the other hand, if we put the latest publicly available versions in `composer.json` (as I did), Dependabot will (rightfully) complain about insecure dependencies...

(I've at least updated the dependencies in `README.md` to the latest ELTS versions.)


sebastian-meyer:
This also backports #1173 for the branch `4.x` in order to make the GitHub actions work again.

stweil (approved):
I did not test the changes, but they look good.

PR #1184, opened 2024-03-05
[BUGFIX] Add `sorting` key to metadata

### LINKED ISSUE #1183: [BUG] Metadata `sorting` value is `0` for every entry
## Description

The `sorting` key for the metadata entries is not set when setting up with the new tenant module. As a result moving a entry gives it a higher than 0 value and moves it to the end of the list.

This is not a big deal, but annoying. As a workaround the user needs to move every entry once to be able to set a custom sorting. 

## Screenshots and Examples

sorting-fin

csidirop:
See #1183 

I added the missing entries in [MetadataDefaults.php](https://github.com/csidirop/kitodo-presentation/blob/fix/metadatasorting/Resources/Private/Data/MetadataDefaults.php) and added a call in `addMetadataAction()` so that it is saved in the metadata object.
But it is not stored in the DB, where the value for the sorting key is still 0.

Any idea?

### DIFF
--- a/Classes/Controller/Backend/NewTenantController.php
+++ b/Classes/Controller/Backend/NewTenantController.php
@@ -248,6 +248,7 @@ class NewTenantController extends AbstractController
                 $newRecord->setIsFacet((int) $values['is_facet']);
                 $newRecord->setIsListed((int) $values['is_listed']);
                 $newRecord->setIndexAutocomplete((int) $values['index_autocomplete']);
+                $newRecord->setSorting((int) $values['sorting']);
 
                 if (is_array($values['format'])) {
                     foreach ($values['format'] as $format) {
--- a/Resources/Private/Data/MetadataDefaults.php
+++ b/Resources/Private/Data/MetadataDefaults.php
@@ -31,6 +31,7 @@ return [
         'is_facet' => 1,
         'is_listed' => 1,
         'index_autocomplete' => 0,
+        'sorting' => 1,
     ],
     'title' => [
         'format' => [
@@ -60,6 +61,7 @@ return [
         'is_facet' => 0,
         'is_listed' => 1,
         'index_autocomplete' => 1,
+        'sorting' => 2,
     ],
     'volume' => [
         'format' => [
@@ -79,6 +81,7 @@ return [
         'is_facet' => 0,
         'is_listed' => 1,
         'index_autocomplete' => 0,
+        'sorting' => 3,
     ],
     'date' => [
         'format' => [
@@ -98,6 +101,7 @@ return [
         'is_facet' => 0,
         'is_listed' => 1,
         'index_autocomplete' => 0,
+        'sorting' => 4,
     ],
     'author' => [
         'format' => [
@@ -122,6 +126,7 @@ return [
         'is_facet' => 1,
         'is_listed' => 1,
         'index_autocomplete' => 1,
+        'sorting' => 5,
     ],
     'place' => [
         'format' => [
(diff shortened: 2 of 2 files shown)


sebastian-meyer:
According to [documentation](https://docs.typo3.org/m/typo3/reference-tca/11.5/en-us/BestPractises/CommonFields.html#manual-sorting-in-the-backend) sorting should be handled by the DataHandler without any need for manually setting values in the database. It should be sufficient to set `sortby` in the TCA ([which we did](https://github.com/kitodo/kitodo-presentation/blob/master/Configuration/TCA/tx_dlf_metadata.php#L23)) and TYPO3 should do the rest. I have no idea why it doesn't work...

csidirop:
Okay, thanks for checking the documentation. Thats really weird than

beatrycze-volk:
> According to [documentation](https://docs.typo3.org/m/typo3/reference-tca/11.5/en-us/BestPractises/CommonFields.html#manual-sorting-in-the-backend) sorting should be handled by the DataHandler without any need for manually setting values in the database. It should be sufficient to set `sortby` in the TCA ([which we did](https://github.com/kitodo/kitodo-presentation/blob/master/Configuration/TCA/tx_dlf_metadata.php#L23)) and TYPO3 should do the rest. I have no idea why it doesn't work...

Currently DataHandler is not used to insert metadata. I was writing about it here https://github.com/kitodo/kitodo-presentation/issues/1147#issuecomment-1961467931 : 

> `Extbase` directly writes to the database, and thus bypasses `DataHandler`. `DataHandler` is actually the one which is responsible for the localization of records.

sebastian-meyer:
> `Extbase` directly writes to the database, and thus bypasses `DataHandler`. `DataHandler` is actually the one which is responsible for the localization of records.

But that's by choice, not by design, isn't it? I don't see a reason not to revert to using the `DataHandler` with `Extbase` (at least in backend context). The `DataHandler` is still available in TYPO3 v12. The old implementation (Extbase + DataHandler) can be seen here: https://github.com/kitodo/kitodo-presentation/blob/1ff03fc4c3ab4dfdc0102d5d98837f1e72f57103/Classes/Controller/NewTenantController.php

sebastian-meyer:
Oh, I just saw that you already suggested restoring the old method using the `DataHandler` in #1147... Sorry, I missed that!

But yes, I think that's the way we should proceed!

beatrycze-volk:
> But yes, I think that's the way we should proceed!

Ok, then I will revert it.

beatrycze-volk:
@csidirop As I'm still working on `DataHandler `usage and can't give the prediction when it will work (basic save works but no localization) ,I have found workaround for sorting problem: https://github.com/beatrycze-volk/kitodo-presentation/commit/6e02b6ec83eaf96f6bc1a7f0dbe6ca2335c2c3dc

It is necessary to add to `tx_dlf_metadata.php`:

```
        'sorting' => [
            'config' => [
                'type' => 'passthrough',
            ],
        ],
```

@sebastian-meyer Are you ok with this solution?

sebastian-meyer:
> It is necessary to add to `tx_dlf_metadata.php`:
> 
> ```
>         'sorting' => [
>             'config' => [
>                 'type' => 'passthrough',
>             ],
>         ],
> ```
> 
> @sebastian-meyer Are you ok with this solution?

Yes, that's fine! Thank you!

sebastian-meyer (approved):

sebastian-meyer:
Merging this as a temporary workaround (see #1209).

PR #1185, opened 2024-03-06
[BUGFIX] Fix search word highlighting

### LINKED ISSUE #1127: [BUG] Search word highlighting does not work (`tx_dlf[highlight_word]`)
## Description

With prior versions (~presentation v2?) there was a optical highlight for searched words in opened documents. This is missing in the last releases for at least presentation v3.4.3:

comparisson
Here in detail:
detail

**Left (v2.2.0):** https://digi.bib.uni-mannheim.de/suchergebnis/seitenansicht-suche/?tx_dlf[id]=3693&tx_dlf[page]=76&tx_dlf[highlight_word]=Exprorektor
 
**Right (master branch):** https://playground.bib.uni-mannheim.de/werksansicht?tx_dlf[highlight_word]=Exprorektor&tx_dlf[id]=2210&tx_dlf[page]=76
(Dev instance, might change, or be broken)

It seams that the `tx_dlf[highlight_word]` query is ignored.

## Reproduction

1. Setup and old (~v2.x) and a new presentation instance
2. Index a document with fulltext
3. Search for word that exists
4. There should be some hits
5. Take on and open it
6. The old instance will highlight the search prompt in the documents overlay, the new one doesn't.

## Expected Behavior

Highlight the search prompt in the documents overlay.

csidirop:
See #1127

In addition to fixing the search word highlighting, I changed the way toe strings are compared. Because the search shows results for substring matches (e.g: `Mannheim` in `Mannheim.`), but the search word highlighting does not (in the example because of the dot `.`). There is a downside and a upsite.

- Downside: Searching for `Mann` does not deliver `Mannheim`, but the search word highlighting would still mark `Mannheim`.
- Upsite: Searching with a wildcard for `Mann*` does deliver `Mannheim`, and now the search word highlighting would marks `Mannheim` too.

I could programme the comparison more precisely, but that would require more code. What do you think?

Last but not least: We should consider renaming the function and its parameters. e.g: `coordinates` ist misleading. As far as I see `searchFeatureCollectionForCoordinates()` gets only called in [PageView.js](https://github.com/kitodo/kitodo-presentation/blob/master/Resources/Public/JavaScript/PageView/PageView.js#L508) when there are highlight words.

### DIFF
--- a/Resources/Public/JavaScript/PageView/Utility.js
+++ b/Resources/Public/JavaScript/PageView/Utility.js
@@ -655,8 +655,8 @@ dlfUtils.scaleToImageSize = function (features, imageObj, width, height, optOffs
 dlfUtils.searchFeatureCollectionForCoordinates = function (featureCollection, coordinates) {
     var features = [];
     featureCollection.forEach(function (ft) {
-        if (ft.get('fulltext') !== undefined) {
-            if (ft.getId() === coordinates) {
+        if (ft.values_.fulltext !== undefined) {
+            if (ft.values_.fulltext === coordinates) {
                 features.push(ft);
             }
         }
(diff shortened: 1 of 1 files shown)


sebastian-meyer (approved):
I think the advantage of a broader search outweighs the disadvantage of false positives.

PR #1186, opened 2024-03-07
[BUGFIX] Add `languageUid` to `index_name` to make it unique

### LINKED ISSUE #1182: [BUG] Metadata `index_name` still gets changed when moving metadata in list
## Description

After #1147 has been resolved, it looked like everything worked. To recap: changing a metadata entry and saving it worked! But moving a metadata entry up or down in the list triggers again a renaming of the `index_name`:

`4: The value of the field "index_name" has been changed from "type" to "type0" as it is required to be unique.`

To be honest, I think we could ignore this because everything works as aspected as far as I saw it. But users could be irritated by that.

csidirop:
Fixes #1182 and the issues described in https://github.com/kitodo/kitodo-presentation/issues/1182#issuecomment-1981079193 .

### DIFF
--- a/Classes/Controller/Backend/NewTenantController.php
+++ b/Classes/Controller/Backend/NewTenantController.php
@@ -272,7 +272,7 @@ class NewTenantController extends AbstractController
                     $translatedRecord->setL18nParent($newRecord);
                     $translatedRecord->_setProperty('_languageUid', $siteLanguage->getLanguageId());
                     $translatedRecord->setLabel($this->getLLL('metadata.' . $indexName, $siteLanguage->getTypo3Language(), $metadataLabels));
-                    $translatedRecord->setIndexName($indexName."_".$siteLanguage->getLanguageId());
+                    $translatedRecord->setIndexName($indexName . "_" . $siteLanguage->getLanguageId());
                     $translatedRecord->setWrap($newRecord->getWrap());
 
                     $this->metadataRepository->add($translatedRecord);
--- a/Classes/Controller/Backend/NewTenantController.php
+++ b/Classes/Controller/Backend/NewTenantController.php
@@ -272,7 +272,7 @@ class NewTenantController extends AbstractController
                     $translatedRecord->setL18nParent($newRecord);
                     $translatedRecord->_setProperty('_languageUid', $siteLanguage->getLanguageId());
                     $translatedRecord->setLabel($this->getLLL('metadata.' . $indexName, $siteLanguage->getTypo3Language(), $metadataLabels));
-                    $translatedRecord->setIndexName($indexName);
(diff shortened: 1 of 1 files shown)


Classes/Controller/Backend/NewTenantController.php
```diff
@@ -272,7 +272,7 @@ public function addMetadataAction(): void
                     $translatedRecord->setL18nParent($newRecord);
                     $translatedRecord->_setProperty('_languageUid', $siteLanguage->getLanguageId());
                     $translatedRecord->setLabel($this->getLLL('metadata.' . $indexName, $siteLanguage->getTypo3Language(), $metadataLabels));
-                    $translatedRecord->setIndexName($indexName);
+                    $translatedRecord->setIndexName($indexName."_".$siteLanguage->getLanguageId());
```
sebastian-meyer:
```suggestion
                    $translatedRecord->setIndexName($indexName . '_' . $siteLanguage->getLanguageId());
```

sebastian-meyer (changes requested):
I am wondering if we would need to change the default translation method as well. What happens if we manually translate a metadata record in the backend? I assume the translated record would also get an exact copy of the original `index_name` triggering the same issues when moving the newly translated record around...

Maybe we need to work around this by changing the [`l10n_mode`](https://docs.typo3.org/m/typo3/reference-tca/11.5/en-us/Columns/Properties/L10nMode.html) from 'exclude' to 'prefixLangTitle' and set [`TCEMAIN.translateToMessage`](https://docs.typo3.org/m/typo3/reference-tsconfig/11.5/en-us/PageTsconfig/TceMain.html#pagetcemain-translatetomessage) to an appropriate prefix (like '%s_').

sebastian-meyer:
@csidirop Could you please verify that this also works for records manually translated in the backend? What does the `index_name` of those records look like?

sebastian-meyer:
@beatrycze-volk Is this potentially also related to not using the DataHandler in the backend (like in #1184)?

beatrycze-volk:
> @beatrycze-volk Is this potentially also related to not using the DataHandler in the backend (like in #1184)?

That is quiet possible. With `DataHandler` translated records were created automatically not like now manually.

sebastian-meyer:
So then this should be fixed with your re-introduction of the DataHandler as mentioned in #1184. Nice!

I'll leave this PR for now, because consistently using the DataHandler would be the better solution.

sebastian-meyer:
Reverting to using the DataHandler is more work than anticipated. Therefor I'll merge this workaround for now in order to release version 5.0.
I'll open a new issue for DataHandler usage.

sebastian-meyer (approved):

PR #1194, opened 2024-04-30
[FEATURE] Use MODS reader library

beatrycze-volk:


composer.lock
sebastian-meyer, line 1: Please do not delete `composer.lock`.

sebastian-meyer (changes requested):

sebastian-meyer (approved):

sebastian-meyer (approved):

PR #1201, opened 2024-05-17
[FEATURE] Add option for reindexing documents in chunks

### LINKED ISSUE #1165: [BUG] Memory leak in kitodo:reindex CLI job
## Description

Running the reindex jobs works with a small number of documents or very much memory, but fails with out of memory for huge collections. While the job is running, the used memory continuously grows. 10 GB of memory is not sufficient for less than 3000 documents.

## Reproduction

Just run the CLI job and watch how the memory usage grows.

## Expected Behavior

There is no reason why the required memory should grow with the number of indexed documents.
As soon as a document was indexed, there is no longer a need to store its metadata in RAM.

## Environment

 - Debian bullseye
 - MariaDB 10.5.21
 - TYPO3 10
 - PHP 7.4.33
 - code based on [Kitodo.Presentation 5.0.0-rc1](https://github.com/kitodo/kitodo-presentation/releases/tag/v5.0.0-rc1)

## Additional Context

See also issue #822. The fix added there obviously no longer works.

beatrycze-volk:


sebastian-meyer (approved):

jmechnich:
@beatrycze-volk using two-letter short options does not seem to be working, you might want to remove the -il and -is options.

PR #1211, opened 2024-05-23
[BUGFIX] Add `languageUid` to `index_name` to make it unique for structures too

csidirop:
Basically the same as #1186 but for `structures`


sebastian-meyer (approved):

sebastian-meyer:
Thanks, missed that!

PR #1220, opened 2024-05-29
[MAINTENANCE] Adjust reading of page uid in controller

beatrycze-volk:


Classes/Controller/AbstractController.php
```diff
@@ -91,6 +97,7 @@ public function injectDocumentRepository(DocumentRepository $documentRepository)
     protected function initialize(): void
     {
         $this->requestData = GeneralUtility::_GPmerged('tx_dlf');
+        $this->pageUid = (int) GeneralUtility::_GET('id');
```
sebastian-meyer, line 100: Does this also work when using URL enhancers (rewriting the `id` parameter to a "speaking" path segment)?

beatrycze-volk: I can't test it currently, as my DDEV instance is in the process of migration to PHP 8.2 and TYPO3 11. I was not yet able to configure routing correctly and need to use URLS as `https://sachsendigital.ddev.site/index.php?id=5`. I will mark it as draft and check it out after routing is configured.

sebastian-meyer: Thank you!

beatrycze-volk: I have tested it now. It works also with URL enhancers :)

sebastian-meyer: Great!

sebastian-meyer (approved):

PR #1221, opened 2024-05-29
[BUGFIX] Return in `SearchController` if no search parameters exist

beatrycze-volk:


Classes/Controller/SearchController.php
```diff
@@ -115,6 +115,11 @@ public function mainAction(): void
         $this->searchParams = $this->getParametersSafely('searchParameter');
         // if search was triggered by the ListView plugin, get the parameters from GET variables
         $listRequestData = GeneralUtility::_GPmerged('tx_dlf_listview');
+        // Quit without doing anything if no search parameters.
+        if(empty($this->searchParams) && empty($listRequestData)) {
```
sebastian-meyer:
```suggestion
        if (empty($this->searchParams) && empty($listRequestData)) {
```

sebastian-meyer (changes requested):

sebastian-meyer (approved):

PR #1236, opened 2024-05-31
[BUGFIX] Use empty array if wrap for given property doesn't exists

beatrycze-volk:


sebastian-meyer (approved):

sebastian-meyer (changes requested):
If I remember correctly, the wraps should be arrays, not strings. Or am I missing something?

beatrycze-volk:
> If I remember correctly, the wraps should be arrays, not strings. Or am I missing something?

I will look into it

beatrycze-volk:
@sebastian-meyer you were right, those are arrays :)

sebastian-meyer (approved):

PR #1239, opened 2024-05-31
[BUGFIX] Fix configuration for `forceAbsoluteUrl` property

beatrycze-volk:


sebastian-meyer:
What is your reasoning behind making this setting plugin-specific instead of global? If in an environment where absolute URLs are needed, they are needed everywhere, not just in one plugin.
I think the bug here is not that this setting is global, but that it is ignored by most plugins except `PageView` when generating URLs...

beatrycze-volk:
It was used as a plugin setting, not the global configuration. Because of that this value in settings array was always empty and always evaluated to false. I have moved it there, so the value can be actually accessed.

I can also solve it other way round: access this value from global configuration and not from settings array. Should it be done that way?

sebastian-meyer:
Yes, I think this would be the more consistent solution. Using absolute URLs should be a global setting. Thank you very much!

sebastian-meyer (approved):

PR #1240, opened 2024-05-31
[BUGFIX] Adjust reindexing command in given range

### LINKED ISSUE #1215: [BUG] Reindexing options -l and -b work not with -c (or long versions) or -a
### Problem

At the moment it is not possible to reindex a large collection with -l (--index-limit) and -b (--index-begin). It is only possible to reindex the whole dataset. 
If the reindexing options -l (--index-limit) and -b (--index-begin) are used, a simultaneously specified parameter -c (--coll) is ignored. 
If -a (--all) is used together with the parameters -l (--index-limit) and -b (--index-begin), these two parameters are ignored.

### Proposed Solution
It should also be possible to reindex a large collection using -l and -b. If the current behavior is desired, the documentation would have to be adapted. 

## Additional Context
https://github.com/kitodo/kitodo-presentation/blob/0dd0bfef744d25fb8aebc65190422f7f90011069/Classes/Command/ReindexCommand.php#L157-L186

beatrycze-volk:
Fix for https://github.com/kitodo/kitodo-presentation/issues/1215

### DIFF
--- a/Classes/Command/ReindexCommand.php
+++ b/Classes/Command/ReindexCommand.php
@@ -165,7 +165,6 @@ class ReindexCommand extends BaseCommand
                     ->setLimit((int) $input->getOption('index-limit'))
                     ->setOffset((int) $input->getOption('index-begin'))
                     ->execute();
-    
                 $io->writeln($input->getOption('index-limit') . ' documents starting from ' . $input->getOption('index-begin') . ' will be indexed.');
             } else {
                 // Get all documents.
--- a/Classes/Command/ReindexCommand.php
+++ b/Classes/Command/ReindexCommand.php
@@ -155,31 +155,44 @@ class ReindexCommand extends BaseCommand
         }
 
         if (!empty($input->getOption('all'))) {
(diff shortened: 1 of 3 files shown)


Classes/Command/ReindexCommand.php
```diff
@@ -155,31 +155,44 @@ protected function execute(InputInterface $input, OutputInterface $output): int
...
+                !empty($input->getOption('index-limit'))
+                && $input->getOption('index-begin') >= 0
+            ) {
+                // Get all documents for given limit and start.
+                $documents = $this->documentRepository->findAll()
+                    ->getQuery()
+                    ->setLimit((int) $input->getOption('index-limit'))
+                    ->setOffset((int) $input->getOption('index-begin'))
+                    ->execute();
+
```
sebastian-meyer:
```suggestion

```

sebastian-meyer (changes requested):

sebastian-meyer (approved):

PR #1248, opened 2024-06-04
[BUGFIX] Fix SearchInDocument middleware

beatrycze-volk:
It reverses the change made in https://github.com/kitodo/kitodo-presentation/pull/1088. 

`SearchInDocument` Middleware can not use `getQueryParams()` function because it gets parameters from the POST request.

```
   /**
     * Retrieve any parameters provided in the request body.
     *
     * If the request Content-Type is either application/x-www-form-urlencoded
     * or multipart/form-data, and the request method is POST, this method MUST
     * return the contents of $_POST.
     *
     * Otherwise, this method may return any results of deserializing
     * the request body content; as parsing returns structured content, the
     * potential types MUST be arrays or objects only. A null value indicates
     * the absence of body content.
     *
     * @return null|array|object The deserialized body parameters, if any.
     *     These will typically be an array or object.
     */
    public function getParsedBody();
```

I suspect that the change from mentioned PR has also disabled Suggester, as this request also makes POST AJAX request.

@sebastian-meyer could you test it? For me suggester doesn't work, but 'm not sure if it is really buggy or my DDEV instance is too slow


sebastian-meyer (approved):

sebastian-meyer:
The code looks fine, but since I am at BiblioCon I can't test it right now. I'll have a look after my vacations.

beatrycze-volk:
> The code looks fine, but since I am at BiblioCon I can't test it right now. I'll have a look after my vacations.

Did you have time to look into it?

sebastian-meyer (approved):

sebastian-meyer:
It doesn't work for me either, but that could be a problem of my environment as well, because the suggester didn't ever work here... ;o)

But since there is #454 which is part of an on-going re-work of the Solr integration by @frank-ulrich-weber, the suggester will be refactored, too. I suggest merging this PR (mainly for its maintenance changes) and then waiting for @frank-ulrich-weber to do his magic.

PR #1271, opened 2024-07-10
[BUGFIX] Add $field to the facetset otherwise the name of the field does not exist

christopher-timm:


sebastian-meyer:
I don't see the problem here. The field name is provided as `key` inside the array, so setting it as array-key as well would be duplicated information.
Just for my understanding: Do you get any errors, because the array-key is numerical?

christopher-timm:
In my case, the facets no longer worked and the translations also looked strange. As a query for the facets, the name of the field is written in front of the value (fieldname_faceting:value). Instead of fieldname_faceting, the corresponding number of the array is then used. This happens in the foreach loop in the SearchController in line 492. The variable $field is then only the number of the array.

sebastian-meyer (approved):

sebastian-meyer:
OK, now I see the problem. Thanks for the clarification!

PR #1283, opened 2024-07-22
[IMPROVEMENT] Improvements 3d viewer implementation

markusweigelt:
- Standardisation of the naming
- Extension of the global document type detection to render the embedded3dviewer plugin only if document contains model (only applies to KitodoPageView in TypoScript)
- Adjustment of the documentation
- Content component/plugin https://github.com/slub/dlf-3d-viewers/issues/39 of embedded 3d viewer

**Plugin selection**
image

**Plugin configuration**
image

**Rendering of plugin**
image


Classes/Controller/AbstractController.php
```diff
@@ -491,7 +491,7 @@ private function getDocumentByUid(int $documentId)
      *
      * @return AbstractDocument
      */
-    private function getDocumentByUrl(string $documentId)
+    protected function getDocumentByUrl(string $documentId)
```
sebastian-meyer, line 494: When changing visibility of a method, the docBlock has to be updated as well.

Classes/Controller/Embedded3dViewerController.php
```diff
@@ -0,0 +1,97 @@
...
+    }
+
+    /**
+     * Assign the model from document to view.
+     *
+     * @param AbstractDocument $document The document containing the model
+     */
+    public function assignModelFromDocument(AbstractDocument $document): void
+    {
+        if ($document->metadataArray['LOG_0001']['type'][0] == 'object') {
```
sebastian-meyer: We can't rely on the "main" structure always being `LOG_0001`. In order to fetch the toplevel ID, you can use `AbstrachtDocument->toplevelId`.
```suggestion
        if ($document->metadataArray[$this->document->toplevelId]['type'][0] == 'object') {
```

Classes/ExpressionLanguage/DocumentTypeFunctionProvider.php
```diff
@@ -132,25 +134,35 @@ function($arguments, $cPid)
...
+                        && in_array($metadata['type'][0], ['newspaper', 'ephemera', 'year', 'issue']))
                     ) {
-                        return $type;
+                        $type = $metadata['type'][0];
                     }
-                    $type = $metadata['type'][0];
                 }
+
+                // set 3d object type if object element exist at the first element of LOGICAL struct map
+                if ($this->document->getCurrentDocument()->metadataArray['LOG_0001']['type'][0] == 'object') {
```
sebastian-meyer: We already have the toplevel metadata array at this point (see line 151).
```suggestion
                if ($metadata['type'][0] == 'object') {
```

Classes/ExpressionLanguage/DocumentTypeFunctionProvider.php
```diff
@@ -132,25 +134,35 @@ function($arguments, $cPid)
...
+                    $type = TYPE_3D_OBJECT;
+                }
```
sebastian-meyer: General question: Why do we need this special handling for 3D objects? What's the problem with just working with the given type (`object` instead of `3d-object`)?

sebastian-meyer (changes requested):

Classes/Controller/Embedded3dViewerController.php
sebastian-meyer: If you just need the metadata of the toplevel element (as you do here, I believe), you could use `AbstractDocument::getToplevelMetadata()`.

Classes/ExpressionLanguage/DocumentTypeFunctionProvider.php
markusweigelt: Atm, I can't confirm if there are other non-3D object types, so I'll adapt this for the time being.

Classes/Controller/Embedded3dViewerController.php
```diff
@@ -0,0 +1,97 @@
...
+
+    }
+
+    /**
+     * Builds the embedded 3D viewer url.
+     *
+     * @param string $model The model url
+     * @return string The embedded 3D viewer url
+     */
+    public function buildEmbedded3dViewerUrl(string $model = ""): string
```
sebastian-meyer:
```suggestion
    public function buildEmbedded3dViewerUrl(string $model = ''): string
```
It's no big deal, but please only use `"..."` for strings if this is needed for variable expansion. Using `'...'` is a tiny bit faster and avoids accidental interpretation of strings as variables.

Classes/Controller/Embedded3dViewerController.php
```diff
@@ -0,0 +1,97 @@
...
+    }
+
+    /**
+     * Assign the model from document to view.
+     *
+     * @param AbstractDocument $document The document containing the model
+     */
+    public function assignModelFromDocument(AbstractDocument $document): void
+    {
+        if ($document->getToplevelMetadata()['type'][0] == 'object') {
```
sebastian-meyer:
```suggestion
        if ($document->getToplevelMetadata()['type'][0] === 'object') {
```
Please use strict comparisons whenever possible.

Classes/ExpressionLanguage/DocumentTypeFunctionProvider.php
```diff
@@ -27,6 +27,8 @@
 use TYPO3\CMS\Extbase\Configuration\ConfigurationManagerInterface;
 use TYPO3\CMS\Extbase\Object\ObjectManager;
 
+const TYPE_OBJECT = "object";
+
```
sebastian-meyer: From my point of view it doesn't make sense to work with a constant here, but hard-code the same value at other places (like in `Classes/Controller/Embedded3dViewerController.php`). Either make it completely configurable (e. g. in `ext_conf_template.txt`) or just use the string value (like we do for newspapers and their components).

Classes/ExpressionLanguage/DocumentTypeFunctionProvider.php
```diff
@@ -132,25 +134,26 @@ function($arguments, $cPid)
...
-                if (!empty($metadata['type'][0])) {
-                    // Calendar plugin does not support IIIF (yet). Abort for all newspaper related types.
-                    if (
-                        $this->document->getCurrentDocument() instanceof IiifManifest
-                        && array_search($metadata['type'][0], ['newspaper', 'ephemera', 'year', 'issue']) !== false
-                    ) {
-                        return $type;
-                    }
+                if (!empty($metadata['type'][0])
+                    && (!$this->isIiifManifestWithNewspaperRelatedType($metadata['type'][0]) || $metadata['type'][0] == TYPE_OBJECT)) {
```
sebastian-meyer:
```suggestion
                    && !$this->isIiifManifestWithNewspaperRelatedType($metadata['type'][0])) {
```
We don't need the additional clause. `$this->isIiifManifestWithNewspaperRelatedType()` already returns `false` for 3D objects (which is negated and therefore `true`).

sebastian-meyer (changes requested):

Classes/Controller/Embedded3dViewerController.php
```diff
@@ -0,0 +1,97 @@
...
+
+    /**
+     * Assign the model from document to view.
+     *
+     * @param AbstractDocument $document The document containing the model
+     */
+    public function assignModelFromDocument(AbstractDocument $document): void
+    {
+        if ($document->getToplevelMetadata()['type'][0] === 'object') {
+            $model = trim($document->getFileLocation($document->physicalStructureInfo[$document->physicalStructure[1]]['files']['DEFAULT']));
```
sebastian-meyer, line 93: This is fine for now, but we should make the fileGrp to use configurable in the future. For now, we only use `DEFAULT`, but that may not be the case for everyone.

sebastian-meyer (approved):

Classes/Controller/Embedded3dViewerController.php
markusweigelt: Thx! I have created an issue so we don't forget about it. https://github.com/kitodo/kitodo-presentation/issues/1286

PR #1296, opened 2024-08-09
[FEATURE] Setting for allow write into index

beatrycze-volk:


beatrycze-volk:
```
Issues
======
+ Solved 1
- Added 1
           

Complexity decreasing per file
==============================
+ Classes/Command/ReindexCommand.php  -6
+ Classes/Command/IndexCommand.php  -6
```

So all in all, the end complexity is decreased, but getting it to the expected threshold can be hard here.

Classes/Command/DeleteCommand.php
```diff
@@ -89,8 +89,30 @@ protected function execute(InputInterface $input, OutputInterface $output): int
 
         $this->initializeRepositories((int) $input->getOption('pid'));
 
+        $allowWrite = (int) $this->extConf['solr']['allowWrite'] === 1 ? true : false;
+
+        if ($allowWrite) {
+            return $this->executeDeleteCommand($input, $io);
+        } else {
+            $io->error('This system is not allowed to write into the SOLR Index.');
```
sebastian-meyer:
```suggestion
            $io->error('This system is not allowed to write to the Solr index.');
```

Classes/Command/IndexCommand.php
```diff
@@ -96,15 +96,38 @@ public function configure(): void
...
         $io->title($this->getDescription());
 
         $this->initializeRepositories((int) $input->getOption('pid'));
 
+        $allowWrite = (int) $this->extConf['solr']['allowWrite'] === 1 ? true : false;
+
+        if ($allowWrite) {
+            return $this->executeIndexCommand($input, $io);
+        } else {
+            $io->error('This system is not allowed to write into the SOLR Index.');
```
sebastian-meyer:
```suggestion
            $io->error('This system is not allowed to write to the Solr index.');
```

Classes/Command/ReindexCommand.php
```diff
@@ -112,15 +113,37 @@ public function configure(): void
...
         $io->title($this->getDescription());
 
         $this->initializeRepositories((int) $input->getOption('pid'));
 
+        $allowWrite = (int) $this->extConf['solr']['allowWrite'] === 1 ? true : false;
+
+        if ($allowWrite) {
+            return $this->executeReindexCommand($input, $io);
+        } else {
+            $io->error('This system is not allowed to write into the SOLR Index.');
```
sebastian-meyer:
```suggestion
            $io->error('This system is not allowed to write to the Solr index.');
```

Resources/Private/Language/de.locallang_labels.xlf
```diff
@@ -745,6 +745,10 @@
 				<target>Löschen von Solr Kern zulassen?: Soll beim Löschen eines Solr Kerns im TYPO3 Backend auch der entsprechende Index in Apache Solr gelöscht werden? (Standard ist "FALSE")</target>
 				<source>Allow Solr Core Deletion?: If a Solr Core is deleted in the TYPO3 Backend, should it be deleted in Apache Solr as well? (default is "FALSE")</source>
 			</trans-unit>
+            <trans-unit id="config.solr.allowWrite">
+                <target>Schreibzugriff auf SOLR? (Standard ist "TRUE")</target>
+                <source>Allow write to the SOLR? (default is "TRUE")</source>
```
sebastian-meyer:
```suggestion
                <target>Schreibzugriff auf Solr-Index? (Standard ist "TRUE")</target>
                <source>Allow write to the Solr index? (default is "TRUE")</source>
```

Resources/Private/Language/locallang_labels.xlf
```diff
@@ -560,6 +560,9 @@
 			<trans-unit id="config.solr.allowCoreDelete">
 				<source>Allow Solr Core Deletion?: If a Solr Core is deleted in the TYPO3 Backend, should it be deleted in Apache Solr as well? (default is "FALSE")</source>
 			</trans-unit>
+            <trans-unit id="config.solr.allowWrite">
+				<source>Allow write to the SOLR? (default is "TRUE")</source>
```
sebastian-meyer:
```suggestion
				<source>Allow write to the Solr index? (default is "TRUE")</source>
```

sebastian-meyer (changes requested):

sebastian-meyer (approved):

sebastian-meyer:
Don't worry about the Codacy issue! That's fine!

But could you please explain the reasoning behind this feature? Why would you want to have a read-only Solr index (but not database)? And what about other places where we write to the Solr index (like `Classes\Hooks\DataHandler.php` for example)?

beatrycze-volk:
> But could you please explain the reasoning behind this feature? Why would you want to have a read-only Solr index (but not database)? And what about other places where we write to the Solr index (like `Classes\Hooks\DataHandler.php` for example)?

I was asked about it by @ByteParty. There are going to be instances of Kitodo.Presentation where calling the indexing commands should not be allowed.

sebastian-meyer:
I am not sure if we should implement this in Kitodo.

Primarily because doing it right (i. e. preventing writing everywhere we currently interact with Solr) becomes very complicated. We would have to disallow writing in the `DataHandler` as well which essentially means not allowing any field configuration and data changes in the backend (because otherwise we would get inconsistencies between index and database).

But also, because this seems to be a very SLUB-specific issue and it would be much easier to just prevent the Solr update handler from being accessible by Kitodo. This could be done simply by configuration on the Solr side. Or am I missing something?

ByteParty:
It is only a matter of switching off the indexer cron. It is not about preventing all write operations in the SOLR. The indexer should be on its own TYPO3 instance. So there will be a TYPO3 instance in which editors can edit the data and a TYPO3 instance that keeps the index / DB up to date. Alternatively, the system design of Kitoto Presentation can be reconsidered and the indexer can be outsourced to an extra extension.
The background is the increased performance requirements due to the many AI bots.

sebastian-meyer:
I still don't understand why you need a new configuration flag to disable the indexer. Just don't use it if you don't want to... Since the indexer is a CLI command, it can't be called "accidentally" by bots.

ByteParty:
The indexer is not called by bots. The indexer is simply performance hungry. Yes, you could  not call the indexer, that's right. But with the flag, the system architecture is cleanly documented in Git. On the other side, an admin cannot start the indexer by mistake.

sebastian-meyer:
Closing this after speaking to @ByteParty: In order to make this merge-able more work would be necessary which is not worth it considering the "problem" can be solved organisationally.

PR #1297, opened 2024-08-14
[BUGFIX] Fix exception on not existing 3dviewer configuration path

markusweigelt:
Handles the exception where, with a fresh DFG-Viewer installation and no saved dlf extension configuration, the `3dviewer` configuration key does not exist.

Reported problem: https://github.com/slub/dfg-viewer/pull/295#issuecomment-2286547643

### DIFF
--- a/Classes/Middleware/Embedded3dViewer.php
+++ b/Classes/Middleware/Embedded3dViewer.php
@@ -19,6 +19,7 @@ use Psr\Http\Server\RequestHandlerInterface;
 use Psr\Log\LoggerAwareTrait;
 use TYPO3\CMS\Core\Configuration\ExtensionConfiguration;
 use TYPO3\CMS\Core\Configuration\Loader\YamlFileLoader;
+use TYPO3\CMS\Core\Exception;
 use TYPO3\CMS\Core\Http\HtmlResponse;
 use TYPO3\CMS\Core\Resource\Folder;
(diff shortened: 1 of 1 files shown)


sebastian-meyer (dismissed):
From a code perspective this is fine, but wouldn't it make more sense to assume a default value if the option is not set at all? (Or alternatively, the logged error should mention that simply re-saving configuration fixes this.)

markusweigelt:
If no configuration is set, the function returns an empty string as the viewer name, and then the built-in model viewer is loaded as default.

https://github.com/kitodo/kitodo-presentation/blob/93a201ed69d885c5710f2408d5149c65db0d6b04/Classes/Middleware/Embedded3dViewer.php#L82-L84

So, you don't have to do anything to display the built-in viewer and render the model. If the default behavior needs to be changed, the configuration must be saved in any case, but then the path will be available.

sebastian-meyer:
Then why log an error if everything works just as expected? I'd suggest to just gracefully ignore the unset configuration, if everything works with the default settings anyway.

markusweigelt:
Ok, that's a good point. Then I would simply change the log level to "debug" here, unless you think "warning" would be better? The information would then only appear in the log if the logger is configured accordingly.

sebastian-meyer:
Yes, "debug" would be the appropriate log level, I think. Thank you very much!

sebastian-meyer (approved):

PR #1316, opened 2024-08-28
[BUGFIX ] Translation of structure type doesn't work

### LINKED ISSUE #1314: [BUG] Translation of structure type doesn't work
## Description

Structure types are currently not translated. Instead of it the index name of structure is displayed. It is caused by explicit casting value of metadata from `array` to `string`. Then `parseType()` tries to retrieve from database the translation for the first element of `array`. But after cast the actual first element of `array` is the letter '_i_' instead of the word '_issue_'.

Code:
```
    private function parseType(int $i, array &$metadata) : void
    {
        $structure = $this->structureRepository->findOneByIndexName($metadata[$i]['type'][0]);
        if ($structure) {
            $metadata[$i]['type'][0] = $structure->getLabel();
        }
    }
```

## Expected Behavior

Structure types should be translated and displayed translated in metadata plugin (`MetadataController`). 

## Screenshots and Examples

translate

## Additional Context

Related to #1308

BFallert:
Related to #1314

### DIFF
--- a/Classes/Controller/MetadataController.php
+++ b/Classes/Controller/MetadataController.php
@@ -439,7 +439,7 @@ class MetadataController extends AbstractController
     {
         $structure = $this->structureRepository->findOneByIndexName($metadata[$i]['type']);
         if ($structure) {
-            $metadata[$i]['type'][0] = $structure->getLabel();
+            $metadata[$i]['type'] = $structure->getLabel();
         }
     }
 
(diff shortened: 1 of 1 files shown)


sebastian-meyer (approved):

beatrycze-volk:
Maybe additional explanation why this change should not be introduced:

https://github.com/kitodo/kitodo-presentation/blob/55f8976702dc7303d9c946421b3fb95dd30c7892/Resources/Private/Partials/Metadata/Entries.html#L22

Template assumes that passed value is the array, not string. I didn't test what the Fluid makes when gets string. Maybe there is some mechanism which converts it into array with one element, but anyway it is not good idea to depend on it.

Issue #1308, opened 2024-08-16
[BUG] Metadata values should be of type `array`, but are `string`


beatrycze-volk:
## Description

The value passed to the `foreach` loop is a `string`, but should be an `array`.

## Reproduction

Steps to reproduce the behaviour:

1. Install TYPO3 11.x with PHP 8.x
Click on link to display document page e.g. [https://dfgviewer-dist.ddev.site](https://dfgviewer-dist.ddev.site/viewer?tx_dlf%5Bdouble%5D=0&tx_dlf%5Bid%5D=https://digital.slub-dresden.de/data/kitodo/TheDarea_416971482-19100223/TheDarea_416971482-19100223_mets.xml)
See error

## Expected Behavior

The values should be stored as arrays or their usage should be adjusted to handle the strings. But the first option is more advisable as imploding and next exploding the array looks like not the most appropriate approach.

## Screenshots and Examples

<img width="880" alt="langerror" src="https://github.com/user-attachments/assets/673d8dc5-e0e3-4fe3-91aa-6811dd318b50">

## Environment

 - RDBMS version: MariaDB 10.3
- TYPO3 version: 11.5.37
- PHP version: 8.2

## Additional Context

This line introduces value type change:
```
    $metadata[$i][$name] = is_array($value)
        ? implode($this->settings['separator'], $value)
        : $value;
```

But usages of those values are not adjusted e.g:

```
    foreach ($metadata[$i][$name] as &$langValue) {
        $langValue = Helper::getLanguageName($langValue);
    }
```

sebastian-meyer:
There is a preliminary fix in #1339, but this only fixes non-hierarchical metadata. See this for details: https://github.com/kitodo/kitodo-presentation/blob/5240b16ea8557eea21f5a75b49fd4eade8ae044d/Classes/Controller/MetadataController.php#L172-L179

PR #1318, opened 2024-08-28
[BUGFIX ] Reading of collections from database doesn't work

### LINKED ISSUE #1317: [BUG] Reading of collections from database doesn't work
## Description

Reading of collections from database doesn't work in `MetadataController`.

## Expected Behavior

The collections should be read correctly from database and displayed to the frontend.

## Screenshots and Examples

Argument `$value` is a `string` not and `array`.

```
    private function parseCollections(int $i, $value, array &$metadata) : void
    {
        $j = 0;
        foreach ($value as $entry) {
            $collection = $this->collectionRepository->findOneByIndexName($entry);
            if ($collection) {
                $metadata[$i]['collection'][$j] = $collection->getLabel() ? : '';
                $j++;
            }
        }
    }
```

## Additional Context

Related to #1314 and #1308

BFallert:
Related to #1317

### DIFF
--- a/Classes/Controller/MetadataController.php
+++ b/Classes/Controller/MetadataController.php
@@ -460,7 +460,7 @@ class MetadataController extends AbstractController
         foreach ($value as $entry) {
             $collection = $this->collectionRepository->findOneByIndexName($entry);
             if ($collection) {
-                $metadata[$i]['collection'][$j] = $collection->getLabel() ? : '';
+                $metadata[$i]['collection'] = $collection->getLabel() ? : '';
                 $j++;
             }
         }


Classes/Controller/MetadataController.php
```diff
@@ -460,7 +460,7 @@ private function parseCollections(int $i, $value, array &$metadata) : void
         foreach ($value as $entry) {
             $collection = $this->collectionRepository->findOneByIndexName($entry);
             if ($collection) {
-                $metadata[$i]['collection'][$j] = $collection->getLabel() ? : '';
+                $metadata[$i]['collection'] = $collection->getLabel() ? : '';
```
beatrycze-volk, line 463: In this case only the last read collection will be displayed to the user. Document can be part of more than one collection and each of them should be displayed in metadata list.

BFallert: Without the adjustments only the first letter of the label is inserted, the rest is then the remaining part of the index identifier of the collection, this is probably not intended. It may be necessary to make a further correction if there are several collections, but the existing version is incorrect in any case.

sebastian-meyer: We shouldn't merge a fix which doesn't actually fix the problem except for a single case. ;o)

beatrycze-volk: The real error should be fixed. This only hides the real problem and when the real problem will be corrected this will not work anymore.

sebastian-meyer:
This issue should be fixed by #1339. Could you please check and confirm, @BFallert?

sebastian-meyer:
Closing this, since it has to be fixed with #1308.

PR #1319, opened 2024-08-28
[FEATURE] Add MIME type filter to ensure only supported image files are downloaded

fschoelzel:
The filter is necessary because the method would otherwise let the user download any file present in mets:fileGrp DEFAULT.

### DIFF
--- a/Classes/Controller/ToolboxController.php
+++ b/Classes/Controller/ToolboxController.php
@@ -296,8 +296,7 @@ class ToolboxController extends AbstractController
         "image/x-portable-anymap",
         "image/x-portable-bitmap",
         "image/x-portable-graymap",
-        "image/x-portable-pixmap",
-        "application/pdf"
+        "image/x-portable-pixmap"
     ];
 
     /**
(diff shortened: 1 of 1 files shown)


fschoelzel:
As it says Image Download only and there is an additional PDF Download, the mimetype "application/pdf" can probably be removed in the image constants list.

Classes/Controller/ToolboxController.php
```diff
@@ -268,6 +268,38 @@ public function renderScoreTool()
...
+        "image/x-xpixmap",
+        "image/jp2",
+        "image/jpx",
+        "image/jpm",
+        "image/mj2",
+        "image/x-portable-anymap",
+        "image/x-portable-bitmap",
+        "image/x-portable-graymap",
+        "image/x-portable-pixmap",
+        "application/pdf"
```
markusweigelt: Isn't this mimetype a bit misleading, cause it`s no image mime type? Or should the name of the list therefore be adapted?

fschoelzel: If you mean "application/pdf". Actually yes, it is. See my previous comment. It should accept most of the image mime-types. But if some of them are misleading, they should be removed. I can update the list.

sebastian-meyer (approved):

PR #1330, opened 2024-09-11
[MAINTENANCE] Restrict GitHub CI to use Solr 9.6 due to issues with 9.7

thomaslow:
There is an issue with Solr 9.7, see https://github.com/dbmdz/solr-ocrhighlighting/issues/457. Until this problem is fixed, I suggest to restrict automated GitHub CI tests to Solr 9.6.

PR #1334, opened 2024-09-13
[BUGFIX] Call `sanitizeSettings` function in `loadDocument` of `AbstractController`

### LINKED ISSUE #1329: [BUG] setCPid() must be of the type int, null given
## Description

_-- This Issue might be in the wrong Extension --_

The `settings` array of the `Slub/Dfgviewer/Controller/UriController` is empty. So `$doc->cPid` is set to `null` instead of `0` if the `storagePid` is empty. (in this case the whole array is)
Without the cast to (int) it throws an error.

## Reproduction

1. reload the Site

## Screenshots and Examples

error-cpid-1
error-cpid-2

## Environment

 - OS version: Linux Mint 21 Cinnamon
 - RDBMS version: MariaDB 10.1
 - TYPO3 version: 11.5.38
 - PHP version: 7.4.33
 - Extensions: kitodo/presentation, slub/dfgviewer, slub/slub-digitalcollections

## Additional Context
occurred after merging this PR:
#1324

beatrycze-volk:
I have forgotten push last commit from branch in https://github.com/kitodo/kitodo-presentation/pull/1324

Closes https://github.com/kitodo/kitodo-presentation/issues/1329


beatrycze-volk:
4 tests with failures will be gone after https://github.com/kitodo/kitodo-presentation/pull/1330 is merged

8 tests with errors are coming from `OaiPmhTest` class and are probably related to setting of default values. There is possibility that those tests were actually depending on missing settings and now when those settings are there it cause problem. I will look into it,

beatrycze-volk:
Logger can be accessed in constructor if this feature will be implemented: https://docs.typo3.org/m/typo3/reference-coreapi/11.5/en-us/ApiOverview/Logging/Logger/Index.html#instantiation

sebastian-meyer (approved):

PR #1338, opened 2024-09-18
[FEATURE] Implement MIME type filtering for files in Helper class

fschoelzel:
- switched from simple mime-type list to provided MimeTypeCollection list from TYPO3 Core
- add filter method to helper class
- custom key and category Parameter
- add filter for images in ToolboxController and PageViewController
- prevent download of files which are in mets:fileGrp USE="DEFAULT"

### DIFF
--- a/Classes/Controller/PageViewController.php
+++ b/Classes/Controller/PageViewController.php
@@ -632,23 +632,20 @@ class PageViewController extends AbstractController
         $fileGrpsImages = GeneralUtility::trimExplode(',', $this->extConf['files']['fileGrpImages']);
 
         foreach ($fileGrpsImages as $fileGrpImages) {
-            $files = $this->getFiles($page, $specificDoc);
-            if (!empty($files)) {
-                // Get image link.
-                $file = $this->getFileInfo($files, $fileGrpImages, $specificDoc);
-                if ($file) {
-                    if (Helper::filterFilesByMimeType($file, 'image', 'mimeType')) {
-                        $image['url'] = $file['location'];
-                        $image['mimetype'] = $file['mimeType'];
-                    }
-
-                    // Only deliver static images via the internal PageViewProxy.
-                    // (For IIP and IIIF, the viewer needs to build and access a separate metadata URL, see `getMetadataURL` in `OLSources.js`.)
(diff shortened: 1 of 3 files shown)


sebastian-meyer (commented):
While this is great for the image download tool, will this still work for IIIF and IIP images? Those won't have an `image/*` MIME type, but something like `application/*`. The changes in `PageView` controller suggest that those images won't be displayed any more... Or am I missing something?

Classes/Controller/PageViewController.php
```diff
@@ -634,15 +635,17 @@ protected function getImage(int $page, MetsDocument $specificDoc = null): array
...
-                    // (For IIP and IIIF, the viewer needs to build and access a separate metadata URL, see `getMetadataURL` in `OLSources.js`.)
-                    if ($this->settings['useInternalProxy'] && !str_contains(strtolower($image['mimetype']), 'application')) {
-                        $this->configureProxyUrl($image['url']);
+                    if (Helper::filterFilesByMimeType($file, 'image', 'mimeType')) {
+                        $image['url'] = $file['location'];
+                        $image['mimetype'] = $file['mimeType'];
+
+                        // Only deliver static images via the internal PageViewProxy.
+                        // (For IIP and IIIF, the viewer needs to build and access a separate metadata URL, see `getMetadataURL` in `OLSources.js`.)
+                        if ($this->settings['useInternalProxy'] && !str_contains(strtolower($image['mimetype']), 'application')) {
```
sebastian-meyer: In line 638 you filter out anything, that hasn't an `image/*` MIME type, but in line 644 you still expect to find images with `application/*` MIME type? ;)

fschoelzel: You're Right. I actually missed that. 
I'll look over it when I'm back again.

fschoelzel: @sebastian-meyer I fixed the wrong inclusion of `configureProxyUrl` inside the if statement.

fschoelzel:
I simplified the `getImage` Method and removed the redundancy.
Added Custom Dlf Mimetypes to the Helper Method.

sebastian-meyer (approved):

PR #1339, opened 2024-09-19
[BUGFIX] Change metadata to keep arrays instead of strings

### LINKED ISSUE #1317: [BUG] Reading of collections from database doesn't work

beatrycze-volk:
This change restores correct displaying of metadata.

Closes https://github.com/kitodo/kitodo-presentation/issues/1317


Classes/Controller/MetadataController.php
```diff
@@ -166,15 +166,17 @@ protected function printMetadata(array $metadata): void
...
-                    if ($metadata[$i][$name] === 'Array') {
+                    // PHPStan error
+                    // I don't understand what this code does, so I take it away until author can fix it
+                    /*if ($metadata[$i][$name][0] === 'Array') {
                         $metadata[$i][$name] = [];
                         foreach ($value as $subKey => $subValue) {
                             $metadata[$i][$name][$subKey] = $subValue;
                         }
-                    }
+                    }*/
```
sebastian-meyer, line 179: I think this handles the two-level metadata. So, if a metadata value is an array, then this is supposed to hold the sub-elements as key/value pairs.
But in order to work correctly, the sub-elements would have to be handled differently then the code currently suggests.

Pinging @chrizzor to please have a look at this very important bug fix!

sebastian-meyer (changes requested):
Apart from the fix for two-level metadata fields by @chrizzor this looks fine!

sebastian-meyer:
I am merging this to at least have a fix for "normal" metadata in 5.x branch. But I'll keep #1308 open, because it still won't work for two-level metadata (see my comment above).

Thank you very much for this very important fix, @beatrycze-volk!

sebastian-meyer (approved):

PR #1344, opened 2024-09-24
[BUGFIX] Fix exception when showing metadata of a document

### LINKED ISSUE #1343: [BUG] Exception when showing metadata of a document
## Description

`Argument 1 passed to Kitodo\Dlf\Common\MetsDocument::getLogicalStructure() must be of the type string, array given, called in /var/www/html/kitodo-presentation/Classes/Controller/MetadataController.php on line 313`

Probably related to #1339 and #1317.

## Reproduction

1. Select a document
2. Open page with metadata widget
3. Get Exception

## Expected Behavior

Metadata is shown without exception.

## Environment

- OS version: Linux Docker (docker.io/php7.4-apache)
- RDBMS version: docker.io/library/mariadb:10.11
- Apache Solr version: 9.6
- TYPO3 version: 11.5.39
- PHP version: 7.4

thomaslow:
Fixes #1343


sebastian-meyer (approved):
This is fine, but it's only a temporary fix for a deeper issue: #1308

PR #1337, opened 2024-09-17
[MAINTENANCE] Add support for Typo3 v12 (fixes #900)

thomaslow:
### Changes

The following changes have been implemented:

- Update composer.json for Typo3 v12
- Add missing dependency for `typo3/cms-scheduler`
- Add version restriction for phpunit v9 (phpunit v10 has breaking changes)
- Upgrade dependency `typo3/testing-framework` to v7, which is compatible with Typo3 v11 and v12
- Fix basic problems with extension configuration (see Typo3 [Feature #96733](https://docs.typo3.org/c/typo3/cms-core/main/en-us/Changelog/12.0/Feature-96733-NewBackendModuleRegistrationAPI.html))
- Replace `TYPO3_MODE` references (see Typo3 [Deprecation #92947](https://docs.typo3.org/c/typo3/cms-core/main/en-us/Changelog/11.0/Deprecation-92947-DeprecateTYPO3_MODEAndTYPO3_REQUESTTYPEConstants.html))
- Remove phpstan ignore rule `TYPO3_MODE not found`, which is not required any more
- Fix flash message rendering (see Typo3 [Depcrecation #97787](https://docs.typo3.org/c/typo3/cms-core/main/en-us/Changelog/12.0/Deprecation-97787-SeveritiesOfFlashMessagesAndReportsDeprecated.html))
- Fix `SolrSearch` implementation by extending from `Query` class (see Typo3 [Breaking #96044](https://docs.typo3.org/c/typo3/cms-core/main/en-us/Changelog/12.0/Breaking-96044-HardenMethodSignatureOfLogicalAndAndLogicalOr.html))
- Avoid initialization via constructor in `AbstractController` (see Typo3 [Deprecation #99615](https://docs.typo3.org/c/typo3/cms-core/main/en-us/Changelog/12.2/Deprecation-99615-GeneralUtilityGPMerged.html]))
- Fix NewTenantController by using dependency injection (see Typo3 [Manual](https://docs.typo3.org/m/typo3/reference-coreapi/11.5/en-us/ApiOverview/DependencyInjection/Index.html#when-to-use-dependency-injection-in-typo3))
- Return ResponseInterfaces in controller actions (see Typo3 [Deprecation #92784](https://docs.typo3.org/c/typo3/cms-core/main/en-us/Changelog/11.0/Deprecation-92784-ExtbaseControllerActionsMustReturnResponseInterface.html))
[...]

PR #1353, opened 2024-09-26
[MAINTENANCE] Improve reading of file groups

beatrycze-volk:


sebastian-meyer (approved):

sebastian-meyer:
Tests fail because of the issue describe here: https://github.com/kitodo/kitodo-presentation/pull/1337#issuecomment-2399994527

We need to fix that first, before I can merge this one.

PR #1357, opened 2024-10-11
[MAINTENANCE] Organize plugins in `tt_content.php` file

beatrycze-volk:


beatrycze-volk:
False positive:
```
Configuration/TCA/Overrides/tt_content.php#L35

Direct use of $GLOBALS Superglobal detected.
```

It needs to be used in this file.

sebastian-meyer (approved):

PR #1360, opened 2024-10-17
[MAINTENANCE] Fix some typos in configuration, documentation and code comments

stweil:
Most of them were found using `codespell` and `typos`.

### DIFF
--- a/Classes/Common/MetsDocument.php
+++ b/Classes/Common/MetsDocument.php
@@ -1508,7 +1508,7 @@ final class MetsDocument extends AbstractDocument
                 }
             }
 
-            // Get track info wtih begin end extent time for later assignment with musical
+            // Get track info with begin end extent time for later assignment with musical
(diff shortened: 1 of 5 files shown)


Classes/Common/MetsDocument.php
```diff
@@ -1508,7 +1508,7 @@ private function getPhysicalElements(array $elementNodes, array $fileUse): array
                 }
             }
 
-            // Get track info wtih begin end extent time for later assignment with musical
+            // Get track info with begin end extent time for later assignment with musical
```
stweil: I am not sure here. Perhaps "and extent time" was intended.

sebastian-meyer: Yes, it should be "and" instead of "end".
```suggestion
            // Get track info with begin and extent time for later assignment with musical
```

sebastian-meyer (changes requested):

sebastian-meyer (approved):

PR #1335, opened 2024-09-13
[MAINTENANCE] Upgrade solr-ocrhighlighting to v0.9.1 for Solr 9.7

thomaslow:
Fixes https://github.com/dbmdz/solr-ocrhighlighting/issues/457

PR #1361, opened 2024-10-21
[MAINTENANCE] Fix requirements

stweil:
The newer version of MariaDB works with the DFG Viewer and TYPO3 11 in an installation on Debian trixie.

### DIFF
--- a/README.md
+++ b/README.md
@@ -19,7 +19,7 @@ Currently **TYPO3 11.5 ELTS** and **TYPO3 12.4 LTS** is supported with the follo
 | TYPO3                   | 11.5.37+               | 12.4.16+               |
 | PHP                     | 7.4.x - 8.3.x          | 8.1.x - 8.3.x          |
 | MySQL                   | 5.7.9 - 8.0.x          | 5.7.9 - 8.0.x          |
-| MariaDB                 | 10.2.7 - 10.11.x       | 10.2.7 - 10.11.x       |
+| MariaDB                 | 10.2.7 - 11.4.x        | 10.2.7 - 11.4.x        |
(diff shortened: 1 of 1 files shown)


stweil:
I wonder whether MySQL is still relevant. Maybe it can be dropped.

sebastian-meyer:
The requirements are wrong in more ways than just the DB versions...

In general, the TYPO3 requirements for [11.x](https://get.typo3.org/version/11#system-requirements) and [12.x](https://get.typo3.org/version/12#system-requirements) should be respected. So maybe we should just link those instead of having our own requirements table. The only dependencies for which we have our own requirements are Apache Solr and the solr_ocrhighlighting plugin - and even those are wrong because we support Solr 9 since #1335.

Do you want to rework this or should I do it?

stweil:
I noticed the TYPO3 requirements and at first intended to apply them, but then decided against it. Those requirements are very narrow, and they only say that they mark the tested versions: "TYPO3 has been tested to work on the platforms listed above. Further setups are conceivable". I think it would be good to support a certain range of software versions which are provided by major Linux distributions. Ideally we can then simply say that for example Debian stable and the latest Ubuntu LTS version are supported, and maybe Debian testing (which triggered my PR) and oldstable are supported, too.

But adding links to the TYPO3 requirements sounds like a good idea. I'd just suggest to add a remark that these requirements are not necessarily hard limits and add our own experiences, too.

And of course wrong or outdated requirements should be fixed. You have more experience for this, so please feel free to update my PR or replace it by a new one.

sebastian-meyer:
I've updated requirements documentation with f4f53a8de218348b4429f959a135de66ad0dfc37.

PR #1363, opened 2024-10-22
[BUGFIX] Fix exception 'data structure identifier must be set'

thomaslow:
Tried to investigate a problem but couldn't test Kitodo plugins due to an exception, which is raised as soon as I add a Kitodo plugin to a page:

> (1/1) #1480765571 RuntimeException
> Data structure identifier must be set, typically by executing TcaFlexPrepare data provider before

Applying the suggested fix of the exception description [#1480765571](https://docs.typo3.org/m/typo3/reference-exceptions/main/en-us/Exceptions/1480765571.html) seems to solve the problem.

### DIFF
--- a/Configuration/TCA/Overrides/tt_content.php
+++ b/Configuration/TCA/Overrides/tt_content.php
@@ -17,7 +17,7 @@ if (!defined('TYPO3')) {
 $excludeList = 'layout,select_key,pages,recursive';
 $addList = 'pi_flexform';
 
-$flexFormsDirectory = 'EXT:dlf/Configuration/FlexForms/';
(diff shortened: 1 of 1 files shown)


sebastian-meyer (changes requested):
Wouldn't it be easier to just prepend `FILE:` to the `$flexFormsDirectory` variable at line 20?

thomaslow:
> Wouldn't it be easier to just prepend `FILE:` to the `$flexFormsDirectory` variable at line 20?

Of course, that works, too. I thought adding the word "file" to a variable called "directory" felt wrong somehow.

sebastian-meyer (approved):
Thank you very much!

PR #1368, opened 2024-10-29
[MAINTENANCE] Improvements regarding TYPO3 12 update

markusweigelt:
- Fix components used by DFG-Viewer
- Tested with TYPO3 11 & 12


Classes/ExpressionLanguage/DocumentTypeFunctionProvider.php
```diff
@@ -162,6 +161,7 @@ protected function loadDocument(array $requestData, int $pid): void
         if (!empty($requestData['id'])) {
             $this->initializeRepositories($pid);
             $doc = null;
+            $this->document = null;
```
sebastian-meyer, line 164: Is this really necessary? `$this->document` is declared as a class property and should be `null` anyways as long as it is not explicitly set. Could you explain your reasoning behind this change?

Classes/Middleware/Embedded3dViewer.php
```diff
@@ -210,7 +211,7 @@ public function renderDefaultViewer($model): HtmlResponse
     {
         /** @var ResourceFactory $resourceFactory */
         $resourceFactory = GeneralUtility::makeInstance(ResourceFactory::class);
-        $html = $resourceFactory->retrieveFileOrFolderObject('EXT:dlf/Resources/Private/Templates/Embedded3dViewer/Standalone.html')->getContents();
+        $html = $resourceFactory->retrieveFileOrFolderObject('EXT:dlf/Resources/Public/Html/Embedded3dViewerStandalone.html')->getContents();
```
sebastian-meyer, line 214: Why did you move this template to `public` folder? It is only handled on the server, so it doesn't need to be accessible by clients. Or am I missing something?

Configuration/TypoScript/setup.typoscript
```diff
@@ -66,10 +66,19 @@ page {
...
   }
 }
 
+[typo3.branch=="11.5"]
+  page.inlineSettings {
+    dlfAssets = /typo3conf/ext/dlf/Resources/Public/
+  }
+[else]
+  page.inlineSettings {
+    dlfAssets = /_assets/007831ab53eafe3f6be2b9b1baf2a8d6/
```
sebastian-meyer, line 78: According to [documentation](https://docs.typo3.org/permalink/t3coreapi:Administration/Upgrade/MigrateToComposer/AssetMigration#migration) we can't rely on the asset folder hash being the same for every installation. Is there no better way to access the assets without the need to hard-code the hash here?

Classes/Middleware/Embedded3dViewer.php
markusweigelt: This is because changes in TYPO3 12 modify the behavior of handling resources. `ResourceFactory->retrieveFileOrFolderObject` now only handles public files.

> Tried to access a private resource file ... from fallback compatibility storage. This storage only handles public files.

I looked into other TYPO3 extensions, and a public HTML folder seems to be common.

sebastian-meyer: OK, thanks for the explanation!

Configuration/TypoScript/setup.typoscript
markusweigelt: Atm the hash is only contains the extension key.
https://github.com/TYPO3/typo3/blob/2bd8c68ddb8056482f9948eec4fab995eba3a62d/typo3/sysext/core/Classes/Utility/PathUtility.php#L71

Composer path is removed before hash generation. https://github.com/TYPO3/typo3/blob/2bd8c68ddb8056482f9948eec4fab995eba3a62d/typo3/sysext/core/Classes/Utility/PathUtility.php#L68C53-L68C96

Classes/ExpressionLanguage/DocumentTypeFunctionProvider.php
markusweigelt: The reason was a null pointer exception or an undeclared `$this->document` at the following point: https://github.com/kitodo/kitodo-presentation/blob/e0b76f6ecd88771cad79a0d95b19dc565a787e02/Classes/ExpressionLanguage/DocumentTypeFunctionProvider.php#L189

Classes/Middleware/Embedded3dViewer.php
markusweigelt: For security reasons, we could potentially add a .htaccess file with deny rules, but I don’t think it’s necessary, and I didn’t find any examples of such a file in the public HTML folder in the other TYPO3 extensions too.

sebastian-meyer (approved):

PR #1372, opened 2024-11-07
[BUGFIX] Fix passing page id in `ToolboxController`

beatrycze-volk:


sebastian-meyer (approved):

fschoelzel:
@sebastian-meyer 
Is it possible to have this fix merged into v5.0.x Versions?
Or is it a breaking change because of the changed input Field in the Template?
In this case we can keep the ArrayKey and just fix the Value, so no Sitepackages need to be adjusted.

I urgently need this fix, because the DDB Zeitungsportal SearchInDocument Function won't work without it.

PR #1376, opened 2024-11-18
[MAINTENANCE] Change log level to INFO

markusweigelt:
Using default should not be a warning because it signals a problem and distracts from actual issues.

The following warning will be shown in the log.
```
Setting "displayIiifDescription" not set, using default value "1". Probably FlexForm for controller "Kitodo\Dlf\Controller\MetadataController" is not read. 
```
I think it's useful to inform about configuration decisions without signaling a problem. I'll adjust the message accordingly.

### DIFF
--- a/Classes/Controller/AbstractController.php
+++ b/Classes/Controller/AbstractController.php
@@ -336,7 +336,7 @@ abstract class AbstractController extends ActionController implements LoggerAwar
     {
         if (!array_key_exists($setting, $this->settings) || empty($this->settings[$setting])) {
             $this->settings[$setting] = $value;
-            $this->logger->info('Setting "' . $setting . '" not set, using default value "' . $value . ' in "' . get_class($this) . '".');
+            $this->logger->info('Setting "' . $setting . '" not set, using default value "' . $value . '" in ' . get_class($this) . '.');
         } else {
             $this->settings[$setting] = (int) $this->settings[$setting];
         }
--- a/Classes/Controller/AbstractController.php
+++ b/Classes/Controller/AbstractController.php
@@ -336,7 +336,7 @@ abstract class AbstractController extends ActionController implements LoggerAwar
     {
         if (!array_key_exists($setting, $this->settings) || empty($this->settings[$setting])) {
             $this->settings[$setting] = $value;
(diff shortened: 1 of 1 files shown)


beatrycze-volk:
Actually it is a real problem, as FlexForm should be configured and used.

markusweigelt:
> Actually it is a real problem, as FlexForm should be configured and used.

I think then there is no need for default values here but an error message that this FlexForm must be configured.

beatrycze-volk:
> > Actually it is a real problem, as FlexForm should be configured and used.
> 
> I think then there is no need for default values here but an error message that this FlexForm must be configured.

But then without those values the whole page crashes.

sebastian-meyer:
I am with @markusweigelt here: Falling back to default values should not be considered an error. If it was we shouldn't provide default values, but immediately throw an exception when mandatory settings are not available.

sebastian-meyer:
Making sure that Flexform settings are read at all (if provided) is a different issue which should not be adressed by warning about the usage of default values. Also, I think not using Flexform at all, but instead providing settings in Typoscript is perfectly fine.

So, in essence: The plugins should make sure that mandatory settings are provided (and throw an error otherwise), but they should not need to worry about where those settings come from (Flexform, Typoscript, Extension config).

Classes/Controller/AbstractController.php
```diff
@@ -336,7 +336,7 @@ protected function setDefaultIntSetting(string $setting, int $value): void
     {
         if (!array_key_exists($setting, $this->settings) || empty($this->settings[$setting])) {
             $this->settings[$setting] = $value;
-            $this->logger->warning('Setting "' . $setting . '" not set, using default value "' . $value . '". Probably FlexForm for controller "' . get_class($this) . '" is not read.');
+            $this->logger->info('Setting "' . $setting . '" not set, using default value "' . $value . ' in "' . get_class($this) . '".');
```
```suggestion
            $this->logger->info('Setting "' . $setting . '" not set, using default value "' . $value . '" in ' . get_class($this) . '.');
```

sebastian-meyer (changes requested):

sebastian-meyer (approved):

PR #1380, opened 2024-11-21
[FEATURE] Seamless document browsing

### LINKED ISSUE #760: [FEATURE] Allow seamless browsing between successive issues
### Problem
When working with newspapers or other publications with serial issues, browsing currently stops at the first and last page of an issue. It is necessary to switch to the proceeding or following issues explicitly to view the pages there.

### Proposed Solution
It would be easier if seamless browsing would be supported for all kinds of serial publications, but especially for newspapers (also in the DFG viewer).

michaelkubina:
This PR adds a feature to the navigation controller, that allows navigating to the prev/next document (volume/issue/etc) relative to the currently one viewed. It implements backtracking, so that it allows for overflowing to the next leafe node if necesseary, e.g. from the last newspaper issue of the current year to the first issue of the next year (and back). Since the backtracking is recursive, this allows backtracking any depths which will very well be necessary for complex document trees from Kalliope (inventories with subinventories with subinventories etc.).

This addresses the following issue: #760

### DIFF
--- a/Classes/Domain/Repository/DocumentRepository.php
+++ b/Classes/Domain/Repository/DocumentRepository.php
@@ -624,70 +624,51 @@ class DocumentRepository extends Repository
     }
 
     /**
-     * Find all documents with from Solr
-     *
-     * @access private
-     *
-     * @param array|QueryResultInterface $collections
-     * @param array $settings
-     * @param array $searchParams
-     * @param QueryResult $listedMetadata
-     *
-     * @return SolrSearch
-     */
-    private function findSolr($collections, $settings, $searchParams, $listedMetadata = null, $indexedMetadata = null): SolrSearch
-    {
-        // set settings global inside this repository
-        // (may be necessary when SolrSearch calls back)
-        $this->settings = $settings;
-
-        $search = new SolrSearch($this, $collections, $settings, $searchParams, $listedMetadata, $indexedMetadata);
-        $search->prepare();
-        return $search;
-    }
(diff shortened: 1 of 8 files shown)


Classes/Domain/Repository/DocumentRepository.php
```diff
@@ -645,4 +645,168 @@ private function findSolr($collections, $settings, $searchParams, $listedMetadat
...
+     *
+     * @param int $uid
+     *
+     * @return int|null
+     */
+    public function getPreviousDocumentUid($uid)
+    {
+        $currentDocument = $this->findOneByUid($uid);
+        $currentVolume = '';
+        $parentId = $currentDocument->getPartof();
```
beatrycze-volk: This change will only work in case when documents are stored in database, but in case of DFG Viewer they are passed as XML files. It causes error:
`Call to a member function getPartof() on null`

It is necessary to make a null check already on `$currentDocument` variable.

beatrycze-volk (changes requested):

Classes/Domain/Repository/DocumentRepository.php
sebastian-meyer: @michaelkubina Please apply your fix from #1381 for this pull request as well.

sebastian-meyer (changes requested):

sebastian-meyer (approved):
Thank you!

PR #1383, opened 2024-11-26
[BUGFIX] Adjust null checks so they also check for non existent key arrays

beatrycze-volk:


sebastian-meyer (approved):

stweil:
I just had to apply the same fixes to my local installation of release 5.0.7.

PR #1394, opened 2024-12-04
[FEATURE] Implement IntlDateFormatter for SolrCoreStatus timestamps

fschoelzel:
Due to the deprecation of `strftime()` in PHP 8.1, we are transitioning to the `IntlDateFormatter` class. This change ensures compatibility with TYPO3 v12 for the SolrCoreStatus view in the backend.


fschoelzel:
When Kitodo doesn't rely on TYPO3 v11 anymore, we can switch to  TYPO3 Core `DateFormatter` class introduced in TYPO3 v12.
See Pull Request #1392

sebastian-meyer (approved):

PR #1395, opened 2024-12-04
[FEATURE] 3D viewer selection

markusweigelt:
- Adjusts the `modeldownloadtool` to additionally retrieve the model url from the `tx_dlf[model]` parameter.
- It provides the `viewerselectiontool` to display all viewers that support the model format for rendering (therefore, the model URL is extracted from the METS file or the model parameter is used).

image

- documentation of both tools


Classes/Controller/ToolboxController.php
```diff
@@ -362,18 +395,55 @@ private function renderImageManipulationTool(): void
      */
     private function renderModelDownloadTool(): void
     {
-        if (
-            $this->isDocMissingOrEmpty()
-            || empty($this->settings['fileGrpsModelDownload'])
-        ) {
-            // Quit without doing anything if required variables are not set.
+        $modelUrl = $this->getModelUrl();
+        if ($modelUrl) {
```
sebastian-meyer: Shouldn't this check for NO `$modelUrl` instead?

```suggestion
        if ($modelUrl === '') {
```

sebastian-meyer (changes requested):

Classes/Controller/ToolboxController.php
markusweigelt: Great thx!

Documentation/Plugins/Index.rst
```diff
@@ -1024,6 +1024,60 @@ The fulltext is fetched and rendered by JavaScript into the `<div id="tx-dlf-ful
...
+ :header-rows: 1
+
+ - :Property:
+       Property
+   :Data Type:
+       Data type
+   :Values:
+       Values
+
+ - :Property:
```
beatrycze-volk: Why do you set up this property here?

`fileGrps...` config is configured in `ext_conf_template.txt` (https://github.com/kitodo/kitodo-presentation/blob/4b3a130ff4794170b0fb4fe1970c3f996309b2df/ext_conf_template.txt#L22)

Please move it there.

Documentation/Plugins/Index.rst
```diff
@@ -1024,6 +1024,60 @@ The fulltext is fetched and rendered by JavaScript into the `<div id="tx-dlf-ful
...
+       fileGrpsModelDownload
```
beatrycze-volk: The same as above.

Classes/Controller/ToolboxController.php
```diff
@@ -130,6 +139,30 @@ private function renderToolByName(string $tool): void
...
+     * @access private
+     *
+     * @return string
+     */
+    private function getModelUrl(): string
+    {
+        $modelUrl = '';
+        if (!empty($this->requestData['model'])) {
+            $modelUrl = $this->requestData['model'];
+        } elseif (!($this->isDocMissingOrEmpty() || empty($this->settings['fileGrpsModelDownload']))) {
```
beatrycze-volk: Access to properties from `ext_conf_template.txt`:
Example: `GeneralUtility::trimExplode(',', $this->extConf['files']['fileGrpThumbs']);`

beatrycze-volk (changes requested):

sebastian-meyer:
Any news on this one, @markusweigelt?
I'd love to merge this as soon as you fixed the small issues mentioned by @beatrycze-volk.

markusweigelt:
@sebastian-meyer Currently i had no time to add/discuss these changes and to test. I will tackle this soon.

sebastian-meyer:
Thanks for the head's up!

Classes/Controller/ToolboxController.php
markusweigelt: Thx for this! I adjusted my implementation accordingly, removed some misleading comments, and eliminated `fileGrpsImageDownload` from `tx_dlf_pdfdownloadtool`

Classes/Controller/ToolboxController.php
```diff
@@ -130,6 +139,30 @@ private function renderToolByName(string $tool): void
...
+     */
+    private function getModelUrl(): string
+    {
+        $modelUrl = '';
+        if (!empty($this->requestData['model'])) {
+            $modelUrl = $this->requestData['model'];
+        } elseif (!($this->isDocMissingOrEmpty() || empty($this->extConf['files']['useGroupsModel']))) {
+            $this->setPage();
+            if (isset($this->requestData['page'])) {
+                $file = $this->getFile($this->requestData['page'], GeneralUtility::trimExplode(',', $this->extConf['files']['useGroupsModel']));
```
beatrycze-volk: It would be nice to use consistently  `$this->useGroupsConfiguration`. so for this case it would be something like `$this->useGroupsConfiguration->getModel()`. 

The function for it should be implemented inside `UseGroupsConfiguration` and look like this

```
    public function getModel(): array
    {
        return $this->getByType('Model');
    }
```

But for now you can leave it as it is.

sebastian-meyer (dismissed):

sebastian-meyer:
Shall I wait for your review, @beatrycze-volk?

beatrycze-volk (approved):

Classes/Controller/ToolboxController.php
markusweigelt: Now adjusted as you suggested.

beatrycze-volk (approved):
Thanks!

sebastian-meyer (approved):

PR #1397, opened 2024-12-05
[BUGFIX] Fix highlighting for words 5.x

beatrycze-volk:
`includes()` is case sensitive


beatrycze-volk:
Test failure can't be related to change as it is change in JS.

sebastian-meyer (approved):

PR #1398, opened 2024-12-05
[BUGFIX] Fix GEO JSON in 4.x

beatrycze-volk:
`type` should be first, `coordinates` should be second


beatrycze-volk:
Test failure not relate to the change.

sebastian-meyer (approved):

PR #1401, opened 2024-12-10
[FEATURE] Implement document cache manager

beatrycze-volk:
It adds more flexiblity to cache management in the application

### DIFF
--- a/Classes/Common/DocumentCacheManager.php
+++ b/Classes/Common/DocumentCacheManager.php
@@ -1,0 +1,110 @@
+<?php
+
+/**
+ * (c) Kitodo. Key to digital objects e.V. <contact@kitodo.org>
+ *
+ * This file is part of the Kitodo and TYPO3 projects.
+ *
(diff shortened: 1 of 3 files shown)


sebastian-meyer (approved):

Classes/Common/DocumentCacheManager.php
```diff
@@ -0,0 +1,110 @@
...
+     *
+     * @access public
+     *
+     * @param string $location
+     *
+     * @return AbstractDocument|false
+     */
+    public function get(string $location)
+    {
+        return $this->cache->get($this->getIdentifier($location));
```
stweil, line 53: Maybe here is reason for issue #1515. If the cache contains an instance of `MetsDocument`, the function returns a `MetsDocument` without setting `$useGroupsConfiguration` which is part of the base class `AbstractDocument`. I expect that other variables from `AbstractDocument` are missing, too.

Like this the caching makes Kitodo.Presentation unusable.

PR #1406, opened 2024-12-12
[FEATURE] Add caching to pagegrid

michaelkubina:
This PR adds a cache specifically for the pagegrid and optimizes the for-loop.

Background: each call to the PageGrid-Controller requires the whole `structMap PHYSICAL` to be parsed alongside the corresponding `fileGrp` for the thumbnail URL. This is a very expensive operation with linear time complexety. Since we always load the whole `$entryArray` with each call and pass it later on to the pagination, this results in very long response times as it needs to be processed again and again. The number of pages displayed in the pagegrid (via settings) has no effect on execution time as it only affects the pagination itself.

On a manuscript with about 1800 (https://resolver.sub.uni-hamburg.de/kitodo/PPN1842517988) pages the execution time for the pagegrid alone takes about 10+ seconds. After adding the cache it went down to about 50ms (on our development instance) once there was a cached item.

With this PR and the #1405 optimization the whole response time for said document went down from nearly 25 seconds to about 2,5 seconds on our development instance. On a typical book with about 200 pages it went down from about 3,5 seconds to about 600ms.

Sidenote: At the SUB-HH the pagegrid is directly integrated into the details-page as a initially invisible container. once it is opened we present all pages without the pagination with lazy-loading of the thumbnails. This way we always allow for a whole preview of the document and see it as a navigation feature that should not require a page-reload when opened. Typically document have around 200-300 pages, so that we accept edge-cases like these large documents to be still quite slower than others. Even if it the pagegrid is used as a standalone page, this optimization improves performance significantly.


sebastian-meyer (commented):
This looks fine, but @beatrycze-volk just introduced a dedicated `DocumentCacheManager` in #1401. Maybe we should incorporate this into the new cache manager instead of doing our own caching in `PageGridController`?

beatrycze-volk:
Yes, this changes can be added to `DocumentCacheManager` class.

sebastian-meyer (changes requested):
Please incorporate this into the new cache manager (see #1401).

sebastian-meyer:
Any news on this, @michaelkubina? I'd love to merge this for the 6.0 release.

michaelkubina:
@sebastian-meyer i had a tighter schedule due to some courses...i will look into it the next coming days

sebastian-meyer:
No worries! I didn't want to stress you, but just to make sure it's not forgotten. ;o)

michaelkubina:
Okay, i have looked at it now...sorry for the delay. I see a problem therein, that the `DocumentCacheManager` uses `AbstractDocument` as parameters and return values. The paggrid is just a simple array seperate from the document, so i would not be able to simply utilize the `DocumentCacheManager` to store it as it is.

To make it somehow work, as i see it, it would be necessary to rewrite the `DocumentCacheManager` to some kind of more general CacheManager, that could not only store AbstractDocument, but also PageGrid or possibly even the SolrResponses (or any other caches that we could need in the future).

Or am i missing something?

beatrycze-volk (approved):
Yes, probably it would demand some changes in `DocumentCacheManager` to allow it to store more general data.

I would say, we merge it for now in this form and eventually make those changes in the future.

sebastian-meyer:
Alright! Thanks for investigating!

Classes/Controller/PageGridController.php
```diff
@@ -44,19 +45,32 @@ public function mainAction(): ResponseInterface
...
+            }
+            $cache->set($cacheKey, $entryArray);
         }
 
         // Get current page from request data because the parameter is shared between plugins
         $currentPage = $this->requestData['page'] ?? 1;
 
+        // Mark currently active page
+        $entryArray[$currentPage - 1]['state'] = 'cur';
+
```
sebastian-meyer:
```suggestion
```

sebastian-meyer (changes requested):

sebastian-meyer (approved):

PR #1411, opened 2024-12-19
[MAINTENANCE] Filter array for empty search parameters

beatrycze-volk:


beatrycze-volk:
PHPStan error is a bit questionable: https://github.com/phpstan/phpstan/issues/2266

I will think about it.

sebastian-meyer:
I'd be fine with adding this to the ignored errors in `.github/phpstan.neon`.

sebastian-meyer (approved):

PR #1415, opened 2024-12-19
[MAINTENANCE] Simplify attributes' reading

beatrycze-volk:


Classes/Common/MetsDocument.php
```diff
@@ -1873,4 +1878,18 @@ protected function magicGetNumMeasures(): int
         $this->magicGetMusicalStructure();
         return $this->numMeasures;
     }
+
+    /**
+     * Get node attribute as string.
+     *
```
```suggestion
     *
```

sebastian-meyer (changes requested):

sebastian-meyer (approved):

PR #1417, opened 2024-12-26
[BUGFIX] Fix regression for METS files without physical structure

stweil:


stweil:
Without this fix, anchor or year METS files of periodicals cause a runtime error because they don't have a physical structure. The access to index 1 of the empty array fails (`$this->physicalStructure[1]`).

stweil:
Related issue: https://github.com/slub/dfg-viewer/issues/327. The modifications were needed to fix runtime errors with anchor and year METS files of newspapers. Those METS files don't contain a physical structure, and there may be missing years and missing months. The additional tests catch these special cases.

Even with this pull request, the calendar views still don't work with TYPO3 12, partially because the URL argument tx_dlf[id] is missing (it is replaced by tx_dlf_calendar[id]). The log file shows errors like this one:
```
Fri, 27 Dec 2024 19:59:43 +0000 [ERROR] request="02ad6d0794050" component="Kitodo.Dlf.Controller.AbstractController": Invalid ID "0" or PID "3" for document loading 
```

sebastian-meyer (approved):

PR #1419, opened 2024-12-30
[BUGFIX] MIME Type Filter

fschoelzel:
Fix Mime Type Filter using File Extension instead of mimetype description out of METS/MODS.
Because of inconsistent use of MIME Type description in METS mets:fileSec. 
For Example: `image/jpg` - which is not an allowed MIME Type according to iana [(www.iana.org)](https://www.iana.org/assignments/media-types/media-types.xhtml#image)

see: [https://api.deutsche-digitale-bibliothek.de/items/72OLPWCWDFDG7AVAYDDVJGMYRMZWBHDR/source/record](https://api.deutsche-digitale-bibliothek.de/items/72OLPWCWDFDG7AVAYDDVJGMYRMZWBHDR/source/record)

### DIFF
--- a/Classes/Common/Helper.php
+++ b/Classes/Common/Helper.php
@@ -26,7 +26,6 @@ use TYPO3\CMS\Core\Messaging\FlashMessage;
 use TYPO3\CMS\Core\Messaging\FlashMessageService;
 use TYPO3\CMS\Core\Messaging\FlashMessageQueue;
 use TYPO3\CMS\Core\Resource\MimeTypeCollection;
-use TYPO3\CMS\Core\Resource\MimeTypeDetector;
 use TYPO3\CMS\Core\Utility\ArrayUtility;
 use TYPO3\CMS\Core\Utility\GeneralUtility;
 use TYPO3\CMS\Core\Utility\MathUtility;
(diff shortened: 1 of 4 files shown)


Classes/Common/Helper.php
```diff
@@ -979,13 +983,22 @@ function ($mimeType) use ($categories) {
...
-
-        // Filter custom MIME types based on provided keys
+        // Filter custom dlf MIME types based on provided keys
         $filteredDlfMimeTypes = array_intersect_key($dlfMimeTypeArray, array_flip($dlfMimeTypes));
 
-        if (is_array($file) && isset($file[$mimeTypeKey])) {
-            return in_array($file[$mimeTypeKey], $mimeTypes) || in_array($file[$mimeTypeKey], $filteredDlfMimeTypes);
+        // Filter file based on its MIME type
+        if (!empty($fileUrl)) {
+            $fileExtension = pathinfo($fileUrl, PATHINFO_EXTENSION);
```
beatrycze-volk: Maybe using https://api.typo3.org/11.5/classes/TYPO3-CMS-Core-Utility-PathUtility.html would remove Codacy error.

beatrycze-volk (changes requested):

sebastian-meyer:
Technically this is fine, but I am not sure if it's a good idea to allow invalid MIME types (by ignoring them). People should better fix their METS files instead...
Also, image servers like Cantaloupe don't necessarily provide file extensions, because the file type can be specified by parameter. So some image server links won't have file extensions at all while others may have misleading extension (like "jp2", although the actual graphic is png-encoded).

Are invalid MIME types a common problem in METS files? Otherwise I'd prefer asking the "Bibliothek der Friedrich-Ebert-Stiftung" to fix their files.

sebastian-meyer:
Any news on this, @fschoelzel?

fschoelzel:
> Technically this is fine, but I am not sure if it's a good idea to allow invalid MIME types (by ignoring them). People should better fix their METS files instead... Also, image servers like Cantaloupe don't necessarily provide file extensions, because the file type can be specified by parameter. So some image server links won't have file extensions at all while others may have misleading extension (like "jp2", although the actual graphic is png-encoded).
> 
> Are invalid MIME types a common problem in METS files? Otherwise I'd prefer asking the "Bibliothek der Friedrich-Ebert-Stiftung" to fix their files.

As of my short research, the `<mets:file ID="xxx" MIMETYPE="image/jpg">` is very often. 
(f.e. https://api.deutsche-digitale-bibliothek.de/2/items/I7O7SOUZMYOUIZ2NF6AHQV4ONJ3Y2XDB/source/record)
Thanks for the hints, i didn't thought of this cases. I will change it back to check for the MIMETYPE out of the METS Files.
And, because it is very often, add an exception for `image/jpg` and hope for correct METS Files in the Future.

What do you think @sebastian-meyer ?

sebastian-meyer:
Yes, that's reasonable! Thank you very much!

sebastian-meyer (approved):

sebastian-meyer:
Thank you!

PR #1435, opened 2025-01-22
[MAINTENANCE] Clear document cache if force reload is true

beatrycze-volk:
Additional level for ensuring that the document is really refreshed


beatrycze-volk:
Changes can't cause those test failures, so reason must be some change in test suite itself.

sebastian-meyer (approved):

PR #1452, opened 2025-02-01
[BUGFIX] Remove old code relict which creates wrong log entries

stweil:
The conditional statement is always true which does not make sense and which can result in a log message "Incomplete plugin configuration".


stweil:
See [related discussion](https://github.com/kitodo/kitodo-presentation/commit/a6773b3d9f38aaf185afbabfef298830256f8950#r151938996).

sebastian-meyer (approved):

PR #1455, opened 2025-02-01
[MAINTENANCE] Run tests for all working PHP versions (including PHP 8.4)

stweil:
~This fixes two vulnerabilities which were reported by Dependabot.~

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -5,6 +5,7 @@ on:
     branches: [ "main" ]
   pull_request:
     branches: [ "main" ]
+  workflow_dispatch:
 
 jobs:
   test:
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -12,7 +12,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        variants: [ {typo3: 11.5, php: 8.1}, {typo3: 12.4, php: 8.1} ]
+        variants: [ {typo3: 11.5, php: 8.1}, {typo3: 12.4, php: 8.1}, {typo3: 12.4, php: 8.4}  ]
     steps:
       - name: Checkout code
         uses: actions/checkout@v4
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -13,7 +13,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        variants: [ {typo3: 11.5, php: 8.1}, {typo3: 12.4, php: 8.1}, {typo3: 11.5, php:8.3}, {typo3: 12.4, php: 8.4}  ]
+        variants: [ {typo3: 11.5, php: 8.1}, {typo3: 12.4, php: 8.1}, {typo3: 11.5, php: 8.3}, {typo3: 12.4, php: 8.4}  ]
     steps:
       - name: Checkout code
         uses: actions/checkout@v4
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -13,7 +13,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        variants: [ {typo3: 11.5, php: 8.1}, {typo3: 12.4, php: 8.1}, {typo3: 12.4, php: 8.4}  ]
+        variants: [ {typo3: 11.5, php: 8.1}, {typo3: 12.4, php: 8.1}, {typo3: 11.5, php:8.3}, {typo3: 12.4, php: 8.4}  ]
     steps:
       - name: Checkout code
         uses: actions/checkout@v4
--- a/composer.json
+++ b/composer.json
@@ -25,7 +25,7 @@
     "docs": "https://docs.typo3.org/p/kitodo/presentation/main/en-us/"
   },
   "require": {
-    "php": "8.1 - 8.3",
+    "php": "8.1 - 8.4",
     "ext-curl": "*",
     "ext-dom": "*",
     "ext-json": "*",
--- a/ext_emconf.php
+++ b/ext_emconf.php
@@ -17,7 +17,7 @@ $EM_CONF[$_EXTKEY] = [
     'category' => 'misc',
     'constraints' => [
         'depends' => [
-            'php' => '8.1.0-8.3.99',
+            'php' => '8.1.0-8.4.99',
             'typo3' => '11.5.0-12.4.99'
         ],
         'conflicts' => [],
--- a/Build/Test/docker-compose.yml
+++ b/Build/Test/docker-compose.yml
@@ -16,7 +16,7 @@ services:
       - /var/lib/mysql/:rw,noexec,nosuid
 
   web:
-    image: docker.io/typo3/core-testing-${DOCKER_PHP_IMAGE}:latest
+    image: ghcr.io/typo3/core-testing-${DOCKER_PHP_IMAGE}:latest
     user: "${HOST_UID}:${HOST_GID}"
     stop_grace_period: 1s
     volumes:
@@ -58,7 +58,7 @@ services:
       "
 
   composer_install:
-    image: docker.io/typo3/core-testing-${DOCKER_PHP_IMAGE}:latest
+    image: ghcr.io/typo3/core-testing-${DOCKER_PHP_IMAGE}:latest
     user: "${HOST_UID}:${HOST_GID}"
     volumes:
       - ${DLF_ROOT}:${DLF_ROOT}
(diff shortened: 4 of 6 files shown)


sebastian-meyer:
Hm, I don't like the big span of supported PHP versions, because we do not test all of them. Instead of making the support even wider, I am in favor of narrowing it down to those versions which are supported by both TYPO3 v11 and v12 (i. e. 8.1, 8.2 and 8.3). What do you think?

sebastian-meyer:
Also, updating `composer.lock` for PHP 8.4 breaks tests, because those run on PHP 8.1.

stweil:
Dropping support for PHP 7.4 and PHP 8.0 is a good idea. ~I still have to find the reason for the CI failure with PHP 8.1.~

stweil:
> Also, updating `composer.lock` for PHP 8.4 breaks tests, because those run on PHP 8.1.

It looks like exactly this CI test failure [occurs from time to time](https://github.com/kitodo/kitodo-presentation/actions/runs/13083934605/job/36512453949), so it is unrelated to my modifications.

stweil:
> Dropping support for PHP 7.4 and PHP 8.0 is a good idea.

This is done now. So two PHP releases were dropped, PHP 8.4 was added => the span of PHP releases decreased.

The CI test with PHP 8.4 works locally and can be added here as soon as a [pending issue](https://git.typo3.org/typo3/CI/testing-infrastructure/-/issues/6) is solved.

Update: The answer to this TYPO3 issue was "Hey. We don't upload to docker hub anymore due to their changed policies and limits. Please find current and maintained images at https://github.com/orgs/TYPO3/packages." This is done in a commit of this pull request now.

sebastian-meyer:
Thanks for removing support for old PHP versions! It's much cleaner now!

But I still don't think we should add support for PHP 8.4. I am in favor of sticking to those PHP versions which are supported by both TYPO3 versions, v11 and v12. (TYPO3 v11 doesn't support PHP 8.4.) It makes development and testing much easier.

stweil:
PHP 8.4 is the latest stable release and standard for me on my macBook. I recently switched to a native installation of Kitodo.Presentation there because it makes source code debugging much easier for me. But I can do this using my local fork, of course.

.github/workflows/tests.yml
```diff
@@ -12,7 +12,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        variants: [ {typo3: 11.5, php: 7.4}, {typo3: 12.4, php: 8.1} ]
+        variants: [ {typo3: 11.5, php: 8.1}, {typo3: 12.4, php: 8.1}, {typo3: 12.4, php: 8.3}  ]
```
sebastian-meyer: Why not run tests for all supported combinations?
```suggestion
        variants: [ {typo3: 11.5, php: 8.1}, {typo3: 11.5, php: 8.2}, {typo3: 11.5, php: 8.3}, {typo3: 12.4, php: 8.1}, {typo3: 12.4, php: 8.2}, {typo3: 12.4, php: 8.3} ]
```

composer.json
```diff
@@ -25,7 +25,7 @@
     "docs": "https://docs.typo3.org/p/kitodo/presentation/main/en-us/"
   },
   "require": {
-    "php": "7.4 - 8.3",
+    "php": "8.1 - 8.4",
```
sebastian-meyer, line 28:
```suggestion
    "php": "8.1 - 8.3",
```

ext_emconf.php
```diff
@@ -17,7 +17,7 @@
     'category' => 'misc',
     'constraints' => [
         'depends' => [
-            'php' => '7.4.0-8.3.99',
+            'php' => '7.4.0-8.4.99',
```
sebastian-meyer:
```suggestion
            'php' => '8.1.0-8.3.99',
```

composer.lock
sebastian-meyer, line 1: Please re-run `composer update` with PHP 8.1.

sebastian-meyer (changes requested):
Your work towards PHP v8.4 support is very much appreciated and it's great if it works already! But "official" support should be postponed for the next major version (supporting TYPO3 v12 and v13).

composer.lock
stweil: > Please re-run `composer update` with PHP 8.1.

I created a new pull request #1457 which only updates composer.lock with PHP 8.1.

stweil:
> Your work towards PHP v8.4 support is very much appreciated.

Up to now the only "work" was removing the current limitation which only supports up to PHP 8.3.

.github/workflows/tests.yml
stweil: > Why not run tests for all supported combinations?

Yes, this should be supported for manual CI runs.

I don't expect that it's worth running all combinations each time. Each CI run costs resources. Covering the corner cases (oldest and newest supported releases of TYPO3 and PHP) is hopefully sufficient to detect all possible problems.

sebastian-meyer (dismissed):

sebastian-meyer:
This is fine! But you marked it as draft.
Just let me know when I can go ahead and merge this pull request!

composer.lock
sebastian-meyer: Thanks! But then you have to remove all changes from this pull request to avoid merge conflicts.

stweil:
I marked this PR as draft because it addressed support for PHP 8.4.

stweil:
The new PR #1458 removes old PHP releases and superseeds similar changes that were part of the PR here.

Increasing the test matrix ~will be addressed in another PR~ was done now, too.

stweil:
The pull requests #1505 and #1539 implement parts of this pull requests which can already be applied now.

stweil:
The pull request now added a working test for TYPO3 v12 with PHP 8.4. Therefore I think it is ready for getting merged.

.github/workflows/tests.yml
```diff
@@ -5,14 +5,15 @@ on:
...
+  workflow_dispatch:
 
 jobs:
   test:
     name: TYPO3
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        variants: [ {typo3: 11.5, php: 8.1}, {typo3: 12.4, php: 8.1} ]
+        variants: [ {typo3: 11.5, php: 8.1}, {typo3: 12.4, php: 8.1}, {typo3: 12.4, php: 8.4}  ]
```
sebastian-meyer:
```suggestion
        variants: [ {typo3: 11.5, php: 8.1}, {typo3: 12.4, php: 8.1}, {typo3: 11.5, php:8.3}, {typo3: 12.4, php: 8.4}  ]
```

composer.json
sebastian-meyer: Why did you change this back to `8.4`? Official support should be for the same PHP versions as TYPO3 v11 and v12, so PHP 8.1, 8.2 and 8.4.
You can choose to use another PHP version, but that is up to you und should not be reflected in the official `composer.json`. See https://getcomposer.org/doc/06-config.md#platform for how to set Composer to use a different PHP version, but resolve dependencies anyway.

ext_emconf.php
```diff
@@ -17,7 +17,7 @@
     'category' => 'misc',
     'constraints' => [
         'depends' => [
-            'php' => '8.1.0-8.3.99',
+            'php' => '8.1.0-8.4.99',
```
sebastian-meyer, line 20:
```suggestion
            'php' => '8.1.0-8.3.99',
```
Again, we do not officially support PHP 8.4 as long as we still support TYPO3v11. Please adjust this locally if need arises.

sebastian-meyer (changes requested):

sebastian-meyer:
Please don't get me wrong: I am all for making necessary adjustments to allow PHP 8.4 to be used (like declaring nullable parameters) and your work doing so is very much appreciated! I just don't want to add "official" support as long as we still support TYPO3v11 which does not support PHP 8.4, because that would make testing even more complicated in order to account for yet another PHP version.

ext_emconf.php
stweil: TYPO3 v11 will still limit the PHP releases to 8.3, even if the dlf extension raises the limit. So for TYPO3 v11 nothing changes. The important point is that the new limit works with TYPO3 v12.

sebastian-meyer: OK, fine. Thanks for the reminder!

.github/workflows/tests.yml
```diff
@@ -5,14 +5,15 @@ on:
...
+  workflow_dispatch:
 
 jobs:
   test:
     name: TYPO3
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        variants: [ {typo3: 11.5, php: 8.1}, {typo3: 12.4, php: 8.1} ]
+        variants: [ {typo3: 11.5, php: 8.1}, {typo3: 12.4, php: 8.1}, {typo3: 11.5, php:8.3}, {typo3: 12.4, php: 8.4}  ]
```
sebastian-meyer:
```suggestion
        variants: [ {typo3: 11.5, php: 8.1}, {typo3: 12.4, php: 8.1}, {typo3: 11.5, php: 8.3}, {typo3: 12.4, php: 8.4}  ]
```
Sorry, I messed this up. ;o)

sebastian-meyer (changes requested):

sebastian-meyer (approved):

PR #1456, opened 2025-02-01
[CONFIGURATION] Do not run the GitHub action deploy job in forks

stweil:
Forks cannot deploy the documentation, so this action always failed for the main branch of forks.

### DIFF
--- a/.github/workflows/documentation.yml
+++ b/.github/workflows/documentation.yml
@@ -33,7 +33,7 @@ jobs:
       # Add a dependency to the build job
       needs: build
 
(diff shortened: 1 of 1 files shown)


stweil:
Maybe forks should not run the "documentation" action at all. Should I modify the PR to restrict the whole workflow to the kitodo owner?

sebastian-meyer (approved):

sebastian-meyer:
I think your changes are fine. The workflow should be available on forks in order to at least be able to build documentation - even if deployment fails.

PR #1458, opened 2025-02-02
[MAINTENANCE] Drop support for PHP 7.4 and 8.0

stweil:
Both PHP releases had their end of life long ago.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -12,7 +12,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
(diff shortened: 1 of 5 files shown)


stweil:
The CI list above says "TYPO3 (11.5, 7.4) Expected — Waiting for status to be reported". Where does this come from?

sebastian-meyer (approved):

sebastian-meyer:
> The CI list above says "TYPO3 (11.5, 7.4) Expected — Waiting for status to be reported". Where does this come from?

That's because I've marked those tests as mandatory. I'll fix this after merging.

PR #1473, opened 2025-02-05
[MAINTENANCE] Format XML output of OAI-PMH

stweil:
The current XML output is generated with a Fluid template. Therefore it is badly formatted with lots of empty lines and unnecessary whitespace.

### DIFF
--- a/Classes/Controller/OaiPmhController.php
+++ b/Classes/Controller/OaiPmhController.php
@@ -307,7 +307,7 @@ class OaiPmhController extends AbstractController
         $xmlOutput = $this->view->render();
 
         // Format the XML.
-        $dom = new \DOMDocument();
+        $dom = new DOMDocument();
         $dom->preserveWhiteSpace = false;
         // Here we could also choose `false` for a minimized XML.
         $dom->formatOutput = true;
--- a/Classes/Controller/OaiPmhController.php
+++ b/Classes/Controller/OaiPmhController.php
@@ -11,6 +11,7 @@
 
 namespace Kitodo\Dlf\Controller;
 
+use DOMDocument;
 use TYPO3\CMS\Core\Utility\GeneralUtility;
 use Kitodo\Dlf\Common\Solr\Solr;
 use Kitodo\Dlf\Domain\Model\Token;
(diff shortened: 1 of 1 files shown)


stweil:
Here are some numbers for `http://HOST/oai?verb=ListIdentifiers&metadataPrefix=mets"|wc`:
```
# minimized XML
       3      13    1766
# formatted XML (new code)
      49      59    2046
# unformatted XML (old code)
     172      59    6172
```

So this pull request reduces the size of the answer for this specific OAI-PMH request from 6172 bytes to 2046 bytes.
A minimized XML would reduce the size a little bit more and could be added as a configuration option in the future.

Classes/Controller/OaiPmhController.php
```diff
@@ -302,7 +303,19 @@ public function mainAction(): ResponseInterface
         $this->view->assign('parameters', $this->parameters);
         $this->view->assign('error', $this->error);
 
-        return $this->htmlResponse();
+        // Generate the XML output.
+        $xmlOutput = $this->view->render();
+
+        // Format the XML.
+        $dom = new \DOMDocument();
```
sebastian-meyer:
```suggestion
        $dom = new DOMDocument();
```

sebastian-meyer (changes requested):

sebastian-meyer (approved):

Classes/Controller/OaiPmhController.php
stweil: Note that the code already uses `new \DOMDocument();` in function `getMetsData`. This is more safe and efficient than `new DOMDocument();` which should only be used if this function were overridden in the current namespace.

sebastian-meyer: Codacy complained because you declared `use DOMDocument;` at the top of the file, but then added the namespace in this line anyways. The use-statement already makes clear that there is no such function/object in current namespace and the global version should be used.
The reason why `getMetsData()` uses `new \DOMDocument();` is, that before your changes we didn't have an use-statement for `DOMDocument`. Strictly speaking we should remove the backslash from `getMetsData` as well, now.

stweil: Codacy complained about my new code before I added the `use` statement. I added it to fix the Codacy warning.

sebastian-meyer: Both is true: Codacy generally discourages using any namespaces in the code in favor of declaring all used namespaced functions and objects per use-statement.
So Codacy first complained about the missing use-statement and then complained about a namespace in code for which a use-statement exists.

PR #1482, opened 2025-02-08
[BUGFIX] Add missing label for tx_dlf_tokens

stweil:
This fixes the currently empty record name when a new Kitodo.Presentation record is added in the TYPO3 backend.


stweil:
See related discussion #1469.

sebastian-meyer (approved):

sebastian-meyer:
This is fine for now, but ultimately we should not display those records in the backend at all.

Issue #1480, opened 2025-02-07
[BUG] Fix creation and validation for OAI-PMH setSpec values


stweil:
## Description

According to [OAI-PMH](https://www.openarchives.org/OAI/openarchivesprotocol.html#Set), a set has "a setSpec -- a colon [:] separated list indicating the path from the root of the set hierarchy to the respective node. Each element in the list is a string consisting of any valid [URI unreserved characters](http://www.ietf.org/rfc/rfc2396.txt), which must not contain any colons [:]. Since a setSpec forms a unique identifier for the set within the repository, it must be unique for each set. Flat set organizations have only sets with setSpec that do not contain any colons [:]."

Kitodo.Presentation can create sets automatically while indexing (for example with `kitodo:index`). The [current code](https://github.com/kitodo/kitodo-presentation/blob/4e77a8f8b734d13ff4937dc4c7a9a5ca71e10865/Classes/Command/BaseCommand.php#L342) gets `setSpec` from the collection name in the indexed METS file, but converts this name to lowercase, removes some characters and replaces others. It does not support underscores or colons.

It is also possible to change `setSpec` in the TYPO3 backend. There a [filter](https://github.com/kitodo/kitodo-presentation/blob/f0b6815223eb1d9a3afe02032f4b28bcd4be4dab/Configuration/TCA/tx_dlf_collections.php#L140C1-L140C60) is used which removes blanks and which allows alphanumeric characters, underscore and dash.

## Reproduction

Import METS file with DINI compatible collection names like _open_access_, _ddc:070_ or _doc-type:Periodical_.

## Expected Behavior

Ideally any `setSpec` which conforms to OAI-PMH should be supported while indexing. It should also be possible to change it to any other form which complies with the protocol specification.

Ideally any value for `setSpec` which does not comply with OAI-PMH should be refused.

stweil:
@sebastian-meyer, I suggest to allow uppercase, underscore and colon while indexing in a first step. Like this indexing could create the `setSpec` values which are required for DINI, although more URI unreserved characters would still be unsupported.

The validation for `setSpec` values in the TYPO3 backend needs a custom validation class. Would it be acceptable to simply drop the `alphanum_x` filter in a first step? This reduces the validation, and users could enter any valid (but also invalid) `setSpec`.

Note that these suggestions would allow entering a hierarchical `setSpec`, but full support for a set hierarchy requires additional code changes.

sebastian-meyer:
[...]

PR #1483, opened 2025-02-08
[BUGFIX] Allow OAI-PMH set names which are required for DINI

stweil:
See related issue #1480.

### DIFF
--- a/Classes/Command/BaseCommand.php
+++ b/Classes/Command/BaseCommand.php
@@ -339,7 +339,26 @@ class BaseCommand extends Command
                 $documentCollection = GeneralUtility::makeInstance(Collection::class);
                 $documentCollection->setIndexName($collection);
                 $documentCollection->setLabel($collection);
-                $documentCollection->setOaiName((!empty($this->extConf['general']['publishNewCollections']) ? Helper::getCleanString($collection) : ''));
+                $setSpec = '';
+                if (!empty($this->extConf['general']['publishNewCollections'])) {
+                    // setSpec only allows unreserved characters (rfc2396).
+                    // alnum | "-" | "_" | "." | "!" | "~" | "*" | "'" | "(" | ")"
+                    // Here we are a little bit more restrictive because
+                    // some more unreserved characters are not allowed.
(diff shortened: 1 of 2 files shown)


sebastian-meyer (approved):

sebastian-meyer:
Please see my comment in #1480: Allowing colons but not supporting hierarchical sets breaks OAI-PMH compliance.

Classes/Command/BaseCommand.php
```diff
@@ -339,7 +339,26 @@ private function addCollections(Document &$document, array $collections): void
...
+                    $setSpec = preg_replace('/[\s]/', '-', $setSpec);
+                    // Remove multiple dashes.
+                    $setSpec = preg_replace('/[-]{2,}/', '-', $setSpec);
+                    // Remove undesired characters.
+                    $setSpec = preg_replace('/[^\w:-]/', '', $setSpec);
+                    // A hierarchical setSpec consists of two or more
+                    // normal setSpec which are separated by colons.
+                    // Remove colons which don't separate hierarchical setSpec entries.
+                    $setSpec = trim($setSpec, ':');
+                    $setSpec = preg_replace('/:{2,}/', ':', $setSpec);
```
sebastian-meyer, line 359: Maybe don't allow colons for now until we support hierarchical sets?

stweil: Support for hierarchical sets means that an OAI-PMH request with parameter `set=SETSPEC` should not only return entries from set "SETSPEC", but also entries from all sets "SETSPEC:*", right? That should not be very difficult to implement. Did I miss more cases?

stweil:
> Allowing colons but not supporting hierarchical sets breaks OAI-PMH compliance.

I think it would reduce the current breakage. Hierarchical sets are not an option feature but required by OAI-PMH. Allowing to enter a hierarchical set name would enhance the currently functionality without breaking any existing feature.

Classes/Command/BaseCommand.php
sebastian-meyer: Yes, that's my understanding of the specification as well.

PR #1493, opened 2025-02-19
[BUGFIX] Initialize new document even if record id is not there

beatrycze-volk:
Fix for error:
`Call to a member function setLocation() on null`

Currently document is created only if document record id exists. If not than document is not created but the location is set up anyway. Added creation o document even if there is no record id.

### DIFF
--- a/Classes/Controller/AbstractController.php
+++ b/Classes/Controller/AbstractController.php
@@ -538,14 +538,13 @@ abstract class AbstractController extends ActionController implements LoggerAwar
         $doc = AbstractDocument::getInstance($documentId, $this->settings, true);
 
         if ($doc !== null) {
+            $this->document = GeneralUtility::makeInstance(Document::class);
+
             if ($doc->recordId) {
                 // find document from repository by recordId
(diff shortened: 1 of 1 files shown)


sebastian-meyer:
We need at least one unique identifier for every document in order to be able to identify if a document has already been indexed before. So if the same document gets indexed again we want to update the existing record instead of creating a new one. As far as I am aware, currently the record identifier is used as such a primary identifier.

So if we allow documents without a record identifier to be indexed, we will create a lot of duplicates when re-indexing those documents. We can't rely on the file location for this, because file names could change.

Have you tested if re-indexing and assigning volumes to their multi-volume parents works correctly with your proposed changes?

beatrycze-volk:
But this part of the code is not relevant for indexing, only for loading documents which are passed by URL (so mostly DFG Viewer). It was working correctly in older versions. Indexer is not using `AbstractController` for getting and preparing documents to indexing.

sebastian-meyer:
You are right, my bad!

sebastian-meyer (approved):

PR #1499, opened 2025-02-26
[BUGFIX] Implement display of single collection in `CollectionController`

beatrycze-volk:
`Collection` plugin should display a single collection view if `showSingle` setting is checked.

- setting name adjusted (it was contradicting its description)
- redirect to `pid` of `Collection` plugin instead of `ListView` plugin (it contains list of documents)
- handle two options:
    1. there is one configured collection and then display overview of this collection
    2. there is none configured and then display overview of collection passed in the search parameters

### DIFF
--- a/Classes/Controller/CollectionController.php
+++ b/Classes/Controller/CollectionController.php
@@ -86,17 +86,21 @@ class CollectionController extends AbstractController
 
         // Sort collections according to order in plugin flexform configuration
         if ($this->settings['collections']) {
-            $sortedCollections = [];
             foreach (GeneralUtility::intExplode(',', $this->settings['collections']) as $uid) {
-                $sortedCollections[$uid] = $this->collectionRepository->findByUid($uid);
+                $collections[$uid] = $this->collectionRepository->findByUid($uid);
             }
-            $collections = $sortedCollections;
         } else {
             $collections = $this->collectionRepository->findAll();
         }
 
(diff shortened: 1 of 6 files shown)


sebastian-meyer:
You are right: the description of `dont_show_single` is wrong and contradicts what the setting actually does.

But I am not sure if changing the setting to match its description is the right way of fixing this. I'd rather change the description and keep the setting as it is.

The reason is backwards-compatibility: When we change the whole setting (and even rename it), everybody has to change their configuration in order to keep the old behavior when updating the extension.

What do you think?

beatrycze-volk:
> You are right: the description of `dont_show_single` is wrong and contradicts what the setting actually does.
> 
> But I am not sure if changing the setting to match its description is the right way of fixing this. I'd rather change the description and keep the setting as it is.
> 
> The reason is backwards-compatibility: When we change the whole setting (and even rename it), everybody has to change their configuration in order to keep the old behavior when updating the extension.
> 
> What do you think?

Anyway, the thing was not working in any form before. So I would implement it this time clean. It is also better to use bool values without negation, so true means true and not reversed.

sebastian-meyer:
Ah, OK. I wasn't aware this was broken. Then your fix is fine, of course, because we can assume nobody is using it anyway.

sebastian-meyer (approved):

PR #1502, opened 2025-02-28
[MAINTENANCE] Remove unused elements[$order] assignment in MetsDocument

fschoelzel:


sebastian-meyer (commented):
`$elements[$order]` is used for sorting the structural elements according to their `@ORDER` attribute (see line 1513). And the array is finally returned (line 1517).

Your proposed changes would lead to the methods always returning an empty array. Clearly that's not intended, is it?

fschoelzel:
I'm sorry. I immediately revoked this. Just realized, this was wrong. Might be a Friday...

PR #1503, opened 2025-03-03
[BUGFIX] Remove `excludes` from `FlexForms`

beatrycze-volk:


sebastian-meyer (approved):
From a technical point of view this is fine, but: why?

This is marked as a "bugfix", but I don't see the issue this fixes. What's wrong with the `exclude` setting?

PR #1505, opened 2025-03-04
[BUGFIX] Explicitly declare nullable function parameters

stweil:
PHP 7.1 added support for nullable types, but they could be declared implicitly. PHP 8.4 deprecates implicitly nullable types.

Using an explicit declaration avoids several warnings with PHP 8.4.

### DIFF
--- a/Classes/Controller/PageViewController.php
+++ b/Classes/Controller/PageViewController.php
@@ -312,7 +312,7 @@ class PageViewController extends AbstractController
      * @param int|null $docNumber
      * @return array
      */
-    protected function getMeasures(int $page, MetsDocument $specificDoc = null, $docNumber = null): array
+    protected function getMeasures(int $page, ?MetsDocument $specificDoc = null, $docNumber = null): array
     {
         if ($specificDoc) {
             $doc = $specificDoc;
@@ -380,7 +380,7 @@ class PageViewController extends AbstractController
      *
      * @return array URL and MIME type of fulltext file
      */
-    protected function getScore(int $page, MetsDocument $specificDoc = null)
+    protected function getScore(int $page, ?MetsDocument $specificDoc = null)
     {
         $score = [];
         $loc = '';
@@ -638,7 +638,7 @@ class PageViewController extends AbstractController
      *
      * @return array URL and MIME type of image file
      */
-    protected function getImage(int $page, MetsDocument $specificDoc = null): array
+    protected function getImage(int $page, ?MetsDocument $specificDoc = null): array
     {
         $image = [];
         // Get @USE value of METS fileGrp.
--- a/Classes/Controller/PageViewController.php
+++ b/Classes/Controller/PageViewController.php
@@ -312,7 +312,7 @@ class PageViewController extends AbstractController
      * @param int|null $docNumber
      * @return array
      */
-    protected function getMeasures(int $page, ?MetsDocument $specificDoc = null, $docNumber = null): array
+    protected function getMeasures(int $page, ?MetsDocument $specificDoc = null, ?int $docNumber = null): array
     {
         if ($specificDoc) {
             $doc = $specificDoc;


stweil:
I expect that there are more nullable types which should be declared explicitly, but until now only the 3 locations which are fixed here triggered warnings in my tests.

Classes/Controller/PageViewController.php
```diff
@@ -312,7 +312,7 @@ public function addDocumentAction(FormAddDocument $formAddDocument): ResponseInt
      * @param int|null $docNumber
      * @return array
      */
-    protected function getMeasures(int $page, MetsDocument $specificDoc = null, $docNumber = null): array
+    protected function getMeasures(int $page, ?MetsDocument $specificDoc = null, $docNumber = null): array
```
beatrycze-volk: Does function also with `int` or `string`?

If yes then, `?int $docNumber` could be added here.

stweil: > Does function also with `int` or `string`?

Yes, see for example https://www.php.net/manual/en/migration71.new-features.php.

> If yes then, ?int $docNumber could be added here.

That's right.

stweil:
```suggestion
    protected function getMeasures(int $page, ?MetsDocument $specificDoc = null, ?int $docNumber = null): array
```

stweil:
> I expect that there are more nullable types which should be declared explicitly, but until now only the 3 locations which are fixed here triggered warnings in my tests.

These code lines need similar fixes:
```
Classes/Common/DocumentAnnotation.php:    protected function getAudioPagesByFileId($fileId, $range = null)
Classes/Common/DocumentAnnotation.php:    protected function getMeasurePagesByFileId($fileId, $range = null)
Classes/Common/Solr/Solr.php:    public static function getInstance($core = null): Solr
Classes/Common/Solr/SolrSearch.php:    public function __construct(DocumentRepository $documentRepository, $collections, array $settings, array $searchParams, QueryResult $listedMetadata = null, QueryResult $indexedMetadata = null)
Classes/Controller/Backend/NewTenantController.php:    protected function htmlResponse(?string $html = null): ResponseInterface
Classes/Controller/NavigationController.php:    public function pageSelectAction(PageSelectForm $pageSelectForm = null): ResponseInterface
Classes/Controller/PageViewController.php:    public function convertMeasureOrPage($document, $measure = null, $page = null)
Classes/Domain/Repository/DocumentRepository.php:    public function findSolrByCollection(Collection $collection, $settings, $searchParams, $listedMetadata = null, $indexedMetadata = null)
Classes/Domain/Repository/DocumentRepository.php:    public function findSolrByCollections($collections, $settings, $searchParams, $listedMetadata = null, $indexedMetadata = null): SolrSearch
Classes/Domain/Repository/DocumentRepository.php:    public function findSolrWithoutCollection($settings, $searchParams, $listedMetadata = null, $indexedMetadata = null): SolrSearch
Classes/Domain/Repository/DocumentRepository.php:    private function findSolr($collections, $settings, $searchParams, $listedMetadata = null, $indexedMetadata = null): SolrSearch
Classes/Task/BaseAdditionalFieldProvider.php:    protected function getSolrField(int $solr, int $pid = null): array
Classes/Task/BaseAdditionalFieldProvider.php:    private function getSolrCores(int $pid = null): array
Classes/Validation/AbstractDlfValidationStack.php:    protected function addValidator(string $className, string $title, bool $breakOnError = true, array $configuration = null): void
```

sebastian-meyer (approved):

sebastian-meyer:
Since you identified more files which need adjustments: should I wait for you to fix those as well or do you want to open separate PRs for those?

stweil:
No, please merge. I currently struggle with the broken viewer which is much more important (see https://github.com/kitodo/kitodo-presentation/pull/1536#issuecomment-2732864469).

PR #1513, opened 2025-03-07
[BUGFIX] Ensure proper type casting and null handling in Helper and MetsDocument classes

fschoelzel:


Classes/Common/Helper.php
```diff
@@ -970,7 +970,7 @@ public static function filterFilesByMimeType($file, array $allowedCategories, nu
         // Retrieves MIME types from the TYPO3 Core MimeTypeCollection
         $mimeTypeCollection = GeneralUtility::makeInstance(MimeTypeCollection::class);
         $allowedMimeTypes = array_filter(
-            $mimeTypeCollection->getMimeTypes(),
+            (array) $mimeTypeCollection->getMimeTypes(),
```
sebastian-meyer: `TYPO3\CMS\Core\Resource\MimeTypeCollection::getMimeTypes()` always returns an array, so casting the return value shouldn't be necessary. Or am I missing something?

sebastian-meyer: https://api.typo3.org/12.4/classes/TYPO3-CMS-Core-Resource-MimeTypeCollection.html

fschoelzel: That's true. I got a intelliphense warning that said the PHP-Doc Block says `@return List<string>`.
But it's true, the return type is array anyway. I'll remove it, cause it's not necessary.

sebastian-meyer (approved):

PR #1523, opened 2025-03-12
[BUGFIX] Adjust reading and assigning `pageUid`

beatrycze-volk:


sebastian-meyer (approved):

Classes/Controller/AbstractController.php
```diff
@@ -121,7 +121,7 @@ protected function initialize(RequestInterface $request): void
     {
         /** @var Request $request */
         $this->requestData = $request->getQueryParams()['tx_dlf'] ?? [];
-        $this->pageUid = (int) ($this->requestData['id'] ?? 0);
+        $this->pageUid = !($this instanceof OaiPmhController) ? $this->request->getAttribute('routing')->getPageId() : 0;
```
stweil, line 124: This change breaks the TYPO3 backend which now throws a runtime exception as soon as "New Tenant" is called:
```
Error: Typed property TYPO3\CMS\Extbase\Mvc\Controller\ActionController::$request must not be accessed before initialization
```

stweil:
Commit 0f4914d746a4f9e7d4a306c6f49d5796642d29dd also breaks the viewer in my tests.

beatrycze-volk:
So in this case I would recommend just to go to the old working version:
`$this->pageUid = (int) GeneralUtility::_GET('id');`
It works for TYPO3 11.x, the was changed to reading of document id instead of page id in migration to TYPO3 12.x. I don't have running DDEV for 12.x so I have made adjustment according documentation 
https://docs.typo3.org/m/typo3/reference-coreapi/12.4/en-us/ApiOverview/RequestLifeCycle/Typo3Request.html
https://docs.typo3.org/m/typo3/reference-coreapi/12.4/en-us/ApiOverview/RequestLifeCycle/RequestAttributes/Routing.html#frontend

As we see it would demand to split our `AbstractController` into `FrontendAbstractController` and `BackendAbstractController`, so better for now to return into old way which should still work for TYPO3 12.x

PR #1526, opened 2025-03-12
[BUGFIX] Rewrite `CollectionController`, `ListViewController` and `SearchController`

beatrycze-volk:
`ListView` is now integrated into `Collection `plugin, so display of single collection should redirect to page on which `Collection` plugin for single collection is configured (plugin there needs to have empty `targetPid`!).

### DIFF
--- a/Classes/Controller/CollectionController.php
+++ b/Classes/Controller/CollectionController.php
@@ -140,7 +140,7 @@ class CollectionController extends AbstractController
 
         if (isset($searchRequestData['searchParameter']) && is_array($searchRequestData['searchParameter'])) {
             $this->searchParams = array_merge($this->searchParams ?: [], $searchRequestData['searchParameter']);
-            $GLOBALS['TSFE']->fe_user->setKey('ses', 'search', $this->searchParams);
+            $this->request->getAttribute('frontend.user')->setKey('ses', 'search', $this->searchParams);
         }
 
         // Get current page from request data because the parameter is shared between plugins
--- a/Classes/Controller/CollectionController.php
+++ b/Classes/Controller/CollectionController.php
@@ -14,11 +14,11 @@ namespace Kitodo\Dlf\Controller;
 use Kitodo\Dlf\Common\SolrPaginator;
 use Kitodo\Dlf\Common\Solr\Solr;
 use Kitodo\Dlf\Domain\Model\Collection;
+use Kitodo\Dlf\Domain\Repository\CollectionRepository;
+use Kitodo\Dlf\Domain\Repository\MetadataRepository;
 use TYPO3\CMS\Core\Pagination\SimplePagination;
 use TYPO3\CMS\Core\Utility\GeneralUtility;
 use TYPO3\CMS\Core\Utility\MathUtility;
-use Kitodo\Dlf\Domain\Repository\CollectionRepository;
-use Kitodo\Dlf\Domain\Repository\MetadataRepository;
 
 /**
  * Controller class for the plugin 'Collection'.
@@ -67,6 +67,12 @@ class CollectionController extends AbstractController
     }
 
     /**
+     * @access protected
+     * @var array The current search parameter
+     */
+    protected $searchParams = [];
+
+    /**
      * Show a list of collections
      *
      * @access public
@@ -75,10 +81,9 @@ class CollectionController extends AbstractController
      */
     public function listAction(): void
     {
-        $solr = Solr::getInstance($this->settings['solrcore']);
-
-        if (!$solr->ready) {
-            $this->logger->error('Apache Solr not available');
+        // Quit without doing anything if required variables are not set.
+        if (empty($this->settings['solrcore'])) {
+            $this->logger->warning('Incomplete plugin configuration for SOLR. Please check the plugin settings for UID of SOLR core.');
             return;
         }
 
@@ -103,7 +108,7 @@ class CollectionController extends AbstractController
             }
         }
 
-        $processedCollections = $this->processCollections($collections, $solr);
+        $processedCollections = $this->processCollections($collections);
 
         // Randomize sorting?
         if (!empty($this->settings['randomize'])) {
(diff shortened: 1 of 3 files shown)


beatrycze-volk:
Redirect exists also on SearchController so it needs to be also adjusted in this PR.

sebastian-meyer (dismissed):

sebastian-meyer:
> Redirect exists also on SearchController so it needs to be also adjusted in this PR.

Could you please elaborate on how the list view is supposed to work?

1. If list view for a single collection is now integrated into `CollectionController`, do I have to configure an instance of the `Collection` plugin for every single collection I want to show? Or is the list view just a `Collection` plugin with "empty" configuration (i. e. no collections selected and no targetPid set)?
2. How has the `SearchController` to be changed in order for this work? Does the `SearchController` also have to redirect to the `CollectionController`?
3. Why is there still a `ListViewController` if both controllers use a different list view now?

Sorry, I am confused... ;o)

beatrycze-volk:
> Could you please elaborate on how the list view is supposed to work?

Sure! It will be actually useful to write this into documentation (I have added this to my list).

> 1. If list view for a single collection is now integrated into `CollectionController`, do I have to configure an instance of the `Collection` plugin for every single collection I want to show? Or is the list view just a `Collection` plugin with "empty" configuration (i. e. no collections selected and no targetPid set)?

There are 3 possible states for CollectionController:
- display list of configured collections - `collections` setting contains one or more collections and `showSingle` setting is 0
- display single configured collection - `collections` setting contains one collection and `showSingle` setting is 1
- display single collection passed as request parameter - `collections` setting contains no collections, `showSingle` setting is 1 and collection id is passed in `searchParameter`

Code snippet:

```
        if ($this->settings['collections']) {
            foreach (GeneralUtility::intExplode(',', $this->settings['collections']) as $uid) {
                $collections[$uid] = $this->collectionRepository->findByUid($uid);
            }
        } else {
            $collections = $this->collectionRepository->findAll();
        }

        if ($this->settings['showSingle'] == 1) {
            if (count($collections) == 1 && is_array($collections)) {
                $this->forward('show', null, null, ['collection' => array_pop($collections)]);
            } else {
                $searchParams = $this->getParametersSafely('searchParameter');
                $collection = $this->collectionRepository->findByUid($searchParams['collection']);
                $this->forward('show', null, null, ['collection' => $collection]);
            }
        }
```

> 2. How has the `SearchController` to be changed in order for this work? Does the `SearchController` also have to redirect to the `CollectionController`?

On this part I still think. I try to collect all changes and reconstruct from those parts what was planned there. But I suspect it depends in which place the search action was called. If search happens in all collections then redirect should go to `ListViewController`, but if search was called on single collection then to `CollectionController`.

Actually lat week I have discover the possibility of search in `CalendarController`. It is full text search for single newspaper (all issues of given newspaper). Results of this one are redirected to `ListViewCotroller`.

> 3. Why is there still a `ListViewController` if both controllers use a different list view now?

Actually `CollectionController` uses templates from `ListViewController`:
https://github.com/kitodo/kitodo-presentation/blob/95a1ff54194e4cda5fb2dc8baa7cf16f44b89a8c/Resources/Private/Templates/Collection/Show.html#L36

The part which differs `CollectionController` from `ListViewController` is the inclusion of collection title, description and image: 
https://github.com/kitodo/kitodo-presentation/blob/95a1ff54194e4cda5fb2dc8baa7cf16f44b89a8c/Resources/Private/Templates/Collection/Show.html#L16

> Sorry, I am confused... ;o)

I'm not surprised that you are confused. I was too. I needed a bit of time to look through the code and understand the relation between those three controllers. In addition the work on `CollectionController` is not finished so I'm doing it now.

beatrycze-volk:
I would ignore error about usage of `GLOBALS['TSFE']` as method presented in https://docs.typo3.org/m/typo3/reference-coreapi/11.5/en-us/ApiOverview/TSFE/Index.html#tsfe_frontendUser does not work for me. Maybe I use it in wrong way, but I get error that variable `$request` does not exist.

beatrycze-volk:
I have used $this->request and it contains information about user and session. Unfortunately I can't test if it really works as contains parameter `dontSetCookie = 1`. But previous way was also not inserting those cookies to my session so I suspect that it can be related to my DDEV environment. Should be tested in live system after merge.

Now last error in Codacy is complexity. I could try to still decrease it but I'm a bit afraid if something will not break in process, so I would prefer to leave in the current state. In overall this function has now lower complexity then before I started to work on it.

sebastian-meyer:
Thanks for the explanations!

The initial idea with having a ListView controller was, that there can be several controllers which produce a list of documents (even if that list contains no or just one item): Collection controller, Search controller, Calendar controller, etc. Each of these lists would have to have the same features like sorting, filtering, pagination and so on. In order to not having to implement those features again and again, we used a generic ListView controller to handle those lists.
But as you know that came with a price: redirecting from one controller to the next while keeping the list persistent wasn't easy and sometimes faulty.

Since your pull requests are for the 5.0 and 5.1 branches: How do you plan to handle this for future releases? Will you open another pull request for `main`?

sebastian-meyer (approved):

beatrycze-volk:
> Thanks for the explanations!

You are welcome!

> The initial idea with having a ListView controller was, that there can be several controllers which produce a list of documents (even if that list contains no or just one item): Collection controller, Search controller, Calendar controller, etc. Each of these lists would have to have the same features like sorting, filtering, pagination and so on. In order to not having to implement those features again and again, we used a generic ListView controller to handle those lists. But as you know that came with a price: redirecting from one controller to the next while keeping the list persistent wasn't easy and sometimes faulty.
> 

It is also the same here. Collection Plugin uses the templates from ListView plugin to display the list of documents. On top of that it adds the displaying collection description and image.

> Since your pull requests are for the 5.0 and 5.1 branches: How do you plan to handle this for future releases? Will you open another pull request for `main`?

Yes, I will certainly do it, but not yet. I have already draft of those changes for main branch, but I have not yet setup ddev system for TYPO3 12 on which I could test it. After I will handle it and make sure that it works I will make PR for main branch.

PR #1531, opened 2025-03-14
[BUGFIX] Move fullscreen to own section

### LINKED ISSUE #1489: [BUG] Multi-View: General zoom, rotation and reset rotation buttons are not working
## Description

Clicking these buttons doesn't trigger any action. There's no JavaScript error either. It seems the function is not implemented for multi view.

## Reproduction

1. Navigate to:
https://example.com/viewer?tx_dlf%5Baction%5D=main&tx_dlf%5Bcontroller%5D=PageView&tx_dlf%5BdocPage%5D%5B1%5D=5&tx_dlf%5BdocPage%5D%5B2%5D=9&tx_dlf%5BdocPage%5D%5B3%5D=13&tx_dlf%5BdocPage%5D%5B4%5D=15&tx_dlf%5Bdouble%5D=0&tx_dlf%5Bid%5D=https%3A%2F%2Ftest.dfg-viewer.de%2Ffileadmin%2Fcomposition%2FMus.2389-N-9.xml&tx_dlf%5Bmultiview%5D=1&tx_dlf%5Bpage%5D=1&cHash=519b78c80be30d359759090f71db9599
2. Klick one of the zoom buttons + or -

## Expected Behavior

Hide general zoom, rotation and reset rotation buttons in multiview. 

## Screenshots and Examples

Image

## Environment

- TYPO3 version: [e.g. 11.5.41]

christopher-timm:
PR to fix #1489
There are also changes in slub_digitalcollection and dfg-viewer.

### DIFF
--- a/Classes/Controller/NavigationController.php
+++ b/Classes/Controller/NavigationController.php
@@ -154,6 +154,8 @@ class NavigationController extends AbstractController
             }
         }
 
+        $this->view->assign('multiview', $this->requestData['multiview'] ?? null);
+
         return $this->htmlResponse();
     }
 }
--- a/Resources/Private/Templates/Navigation/Main.html
+++ b/Resources/Private/Templates/Navigation/Main.html
@@ -209,19 +209,20 @@
 </f:section>
 
 <f:section name="render.zoom">
-    <div class="tx-dlf-navigation-zoom-in">
-        <a href="#" title="{f:translate(key: 'zoomIn')}" onclick="tx_dlf_viewer.map.zoomIn();">
-            <f:translate key="zoomIn"/>
-        </a>
-    </div>
-    <div class="tx-dlf-navigation-zoom-out">
-        <a href="#" title="{f:translate(key: 'zoomOut')}" onclick="tx_dlf_viewer.map.zoomOut();">
-            <f:translate key="zoomOut"/>
-        </a>
-    </div>
-</f:section>
-
-<f:section name="render.fullscreen">
+    <f:if condition="{multiview} != 1">
+        <f:then>
+            <div class="tx-dlf-navigation-zoom-in">
+                <a href="#" title="{f:translate(key: 'zoomIn')}" onclick="tx_dlf_viewer.map.zoomIn();">
+                    <f:translate key="zoomIn"/>
+                </a>
+            </div>
+            <div class="tx-dlf-navigation-zoom-out">
+                <a href="#" title="{f:translate(key: 'zoomOut')}" onclick="tx_dlf_viewer.map.zoomOut();">
+                    <f:translate key="zoomOut"/>
+                </a>
+            </div>
+        </f:then>
+    </f:if>
     <div class="tx-dlf-navigation-fullscreen">
         <a href="#" title="{f:translate(key: 'zoomFullscreen')}">
             <f:translate key="zoomFullscreen"/>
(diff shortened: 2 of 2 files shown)


sebastian-meyer (approved):

sebastian-meyer (changes requested):
I think simply moving the zoom and rotation buttons to their own section isn't sufficient. It still means that users of Kitodo.Presentation would have to add custom CSS in order to make sure, the zoom/rotation section is not visible in MultiView mode (which isn't even documented).

Instead, the template should only render the zoom/rotation section if the MultiView mode is not active. I suggest introducing a variable in the controller which can be used in the template to distinguish between MultiView and SingleDocument view and render the buttons accordingly.

sebastian-meyer:
Maybe we don't even need a new variable. Isn't there an URL parameter always set if MultiView is active?

christopher-timm:
Yes, there is the multiview parameter, which is also written to gp-multiview via Typoscript in the dfg-viewer extension:
https://github.com/slub/dfg-viewer/blob/471e1eec0e1fe67dc59197f8fa3ade9153243849/Configuration/TypoScript/Page/page.typoscript#L87

The parameter would then also have to be configured for Kitodo.presentation. Then it should also be available in the template, right?

markusweigelt:
I tested a version with your branches of all three extensions, and the buttons are still displayed.

What was the plan behind changing the fullscreen mode to a separate section without hiding the other sections?

> The parameter would then also have to be configured for Kitodo.presentation. Then it should also be available in the template, right?

Yes and no. Currently, the condition to show `gp-multiview` is based on a parameter in Kitodo.Presentation, whereas in the DFG-Viewer extension, the condition is handled through `docArray` (see [this line](https://github.com/slub/dfg-viewer/blob/471e1eec0e1fe67dc59197f8fa3ade9153243849/Resources/Private/Plugins/Kitodo/Templates/PageView/Main.html#L27)).

Perhaps we could unify these condition variants in the PageViewController ([see here](https://github.com/kitodo/kitodo-presentation/blob/3feed01d17de7f00d4596d0c8e94f7bc3c4142d9/Classes/Controller/PageViewController.php#L151)) and use a dedicated template variable from there. Maybe the existing `multiview`.

Is it actually correct to render the single view when docArray is empty? I think not. Instead of relying on a `docArray` check, we should use a dedicated template variable. If it’s empty, an empty grid should be displayed rather than defaulting to single view.

christopher-timm:
The MultiView parameter is now also available in the navigation plugin and is checked accordingly in the templates. The buttons should no longer be displayed in multi-view mode.

> Is it actually correct to render the single view when docArray is empty? I think not. Instead of relying on a `docArray` check, we should use a dedicated template variable. If it’s empty, an empty grid should be displayed rather than defaulting to single view.

Does this mean that there are documents that neither contain pages nor reference documents?

sebastian-meyer:
There are no documents without pages or referenced documents, but there should be a way to enable an empty multiview grid (and then add documents manually). I am not sure if that's what @markusweigelt is refering to.

markusweigelt (approved):
It's working on my end! I encountered an issue due to missing documents, but that has been resolved in this pull request: https://github.com/kitodo/kitodo-presentation/pull/1581 
So if you’d like to retest, please merge the PR. The issue was caused by missing documents, which throws an exception before you can test the current PR.

sebastian-meyer:
Thanks for testing and the additional bugfix!

sebastian-meyer (approved):

PR #1534, opened 2025-03-16
[MAINTENANCE] Remove bugfix versions of TYPO3 from dependencies

stweil:
~This fixes security issues reported by Dependabot.~


stweil:
[TYPO3 release 11.5.42](https://get.typo3.org/release-notes/11.5.42) and newer v11 releases are only available as ELTS releases. Therefore this patch will only work for customers who bought commercial TYPO3 support. It does not work in GitHub CI. How do we handle this case?

I'm afraid the only clean solution for master is dropping support for TYPO3 v11.

sebastian-meyer:
We should keep the requirements at the latest publicly available version. If we set it to a newer version all tests would fail, because dependencies are no longer satisfiable by Composer.

Also, version 11.5.41 (and earlier) are already marked as insecure on Packagist, so Composer should warn you anyway if you try to install of those versions.

In general, the requirements set in `composer.json` should represent the minimum versions (or range of versions) which are technically required to run the application. So, strictly speaking, the constraints should be `^11.5` and `^12.4`.

stweil:
> So, strictly speaking, the constraints should be `^11.5` and `^12.4`

That would also reduce the update frequency for these lines. Should we change composer.json to use these constraints? Would `^11|^12` work, too?

sebastian-meyer:
> > So, strictly speaking, the constraints should be `^11.5` and `^12.4`
> 
> That would also reduce the update frequency for these lines. Should we change composer.json to use these constraints? Would `^11|^12` work, too?

I'd need to be `^11.5|^12.4`, because those are the stable LTS releases.

composer.json
```diff
@@ -32,10 +32,10 @@
...
     "ext-openssl": "*",
     "ext-simplexml": "*",
-    "typo3/cms-core": "^11.5.41|^12.4.27",
-    "typo3/cms-extbase": "^11.5.41|^12.4.27",
-    "typo3/cms-scheduler": "^11.5.41|^12.4.27",
-    "typo3/cms-tstemplate": "^11.5.41|^12.4.27",
+    "typo3/cms-core": "^11.5.42|^12.4.27",
+    "typo3/cms-extbase": "^11.5.42|^12.4.27",
+    "typo3/cms-scheduler": "^11.5.42|^12.4.27",
+    "typo3/cms-tstemplate": "^11.5.42|^12.4.27",
```
sebastian-meyer:
```suggestion
    "typo3/cms-core": "^11.5|^12.4",
    "typo3/cms-extbase": "^11.5|^12.4",
    "typo3/cms-scheduler": "^11.5|^12.4",
    "typo3/cms-tstemplate": "^11.5|^12.4",
```

composer.json
```diff
@@ -46,10 +46,10 @@
...
     "phpunit/phpunit": "^9.6",
     "spatie/phpunit-watcher": "^1.23",
-    "typo3/cms-backend": "^11.5.41|^12.4.27",
-    "typo3/cms-fluid": "^11.5.41|^12.4.27",
-    "typo3/cms-fluid-styled-content": "^11.5.41|^12.4.27",
-    "typo3/cms-frontend": "^11.5.41|^12.4.27",
+    "typo3/cms-backend": "^11.5.42|^12.4.27",
+    "typo3/cms-fluid": "^11.5.42|^12.4.27",
+    "typo3/cms-fluid-styled-content": "^11.5.42|^12.4.27",
+    "typo3/cms-frontend": "^11.5.42|^12.4.27",
```
sebastian-meyer:
```suggestion
    "typo3/cms-backend": "^11.5|^12.4",
    "typo3/cms-fluid": "^11.5|^12.4",
    "typo3/cms-fluid-styled-content": "^11.5|^12.4",
    "typo3/cms-frontend": "^11.5|^12.4",
```

stweil:
@sebastian-meyer, I updated the description in the commit message. Hopefully the new description fits better.

sebastian-meyer (approved):

sebastian-meyer:
Thank you very much!

Issue #1515, opened 2025-03-07
[BUG] Indexing and view for cached documents is broken (regression)


stweil:
## Description

Indexing is currently broken. Report from `git bisect`:
```
27812ee974264d5c4d1e10ed693668383a4c3014 is the first bad commit
commit 27812ee974264d5c4d1e10ed693668383a4c3014
Author: Beatrycze Volk <beatrycze.volk@slub-dresden.de>
Date:   Tue Feb 4 13:00:00 2025 +0100

    [FEATURE] Implement `UseGroupsConfiguration` class (#1422)
    
    Co-authored-by: Sebastian Meyer <sebastian.meyer@opencultureconsulting.com>
```

## Reproduction

Steps to reproduce the behaviour:

1. Run `vendor/bin/typo3 kitodo:index --pid 3 --solr dlfCore0 --doc https://digital.ladenburg.de/data/Ladenburger_Wochenblatt/mets18690130.xml`
2. See error

```
Index single document into database and Solr.
=============================================


Uncaught TYPO3 Exception Typed property Kitodo\Dlf\Common\AbstractDocument::$useGroupsConfiguration must not be accessed before initialization
thrown in file /var/customers/webs/weil/viewer/vendor/kitodo/presentation/Classes/Common/MetsDocument.php
in line 1343
```

## Environment

 - OS version: Debian GNU Linux
 - TYPO3 version: 12.4

stweil:
Here is another protocol of indexing on a fresh installation (with the patches required for TYPO3 v12):
```
root@dfgviewer:/var/www/typo3_12/vendor/slub/dfgviewer# /var/www/typo3_12/vendor/bin/typo3 kitodo:index --pid 3 --solr dlfCore0 --doc https://digital.ladenburg.de/data/Ladenburger_Wochenblatt/mets18690130.xml -vv

Index single document into database and Solr.
=============================================

Indexing  ("https://digital.ladenburg.de/data/Ladenburger_Wochenblatt/mets18690130.xml") on PID 3.
--------------------------------------------------------------------------------------------------

Indexing 3 ("https://digital.ladenburg.de/data/Ladenburger_Wochenblatt/mets18690130.xml") on Solr core 1.
---------------------------------------------------------------------------------------------------------

                                                                                                                        
 [ERROR] ERROR: Document with UID "3" could not be indexed on Solr core 1. There are missing mandatory fields (at least 
         one of those: document_format,record_id) in this document.                                                     
[...]

PR #1536, opened 2025-03-17
[BUGFIX] Fix caching for classes IiifManifest and MetsDocument

stweil:
Fixes: 27812ee97426 ("Implement `UseGroupsConfiguration` class")

### DIFF
--- a/Classes/Common/AbstractDocument.php
+++ b/Classes/Common/AbstractDocument.php
@@ -1096,6 +1096,9 @@ abstract class AbstractDocument
      */
     protected function __construct(int $pid, string $location, $preloadedDocument, array $settings = [])
     {
+        // Note: Any change here might require an update in function __sleep
+        // of class MetsDocument and class IiifManifest, too.
+
         $this->pid = $pid;
         $this->useGroupsConfiguration = UseGroupsConfiguration::getInstance();
         $this->setPreloadedDocument($preloadedDocument);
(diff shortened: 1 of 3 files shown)


stweil:
Note that I still only work with METS documents which are unusable without this fix. But I assume that the same errors would also occur with IIIF manifests. They should be fixed now, too.

stweil:
This should fix issue #1515.

stweil:
> This should fix issue #1515.

It only fixes this issue partially. The viewer now no longer raises a runtime error ("oops") when it is called with a document id, but neither the calendar view nor the document view work:
- https://digital.ladenburg.world/viewer?tx_dlf[id]=2
- https://digital.ladenburg.world/viewer?tx_dlf[id]=5

@markusweigelt, @sebastian-meyer, the document view shows information which is related to musical notations although it is a normal text document with METS. Maybe the relevant information is not restored correctly from the cache?
Bildschirmfoto 2025-03-18 um 12 30 54

beatrycze-volk:
I am a bit confused. Why this property needs to be added to serialization? AbstractDocument has many properties and only few are included there. What is so special about this one that it causes this error?

stweil:
This property is accessed after the deserialisation, and this raised a fatal error because it was not set.

I think more properties must be serialized, because currently the viewer is broken for cached documents (which is a regression for a very basic functionality and makes Kitodo.Presentation unusable).

sebastian-meyer:
The problem is, that when deserializing an object, the constructor is not called. So all properties which are set in the constructor need to either be saved during serialization or re-initialized after deserialization. @stweil chose the former (which is fine).

Another way of doing this would be to have some sort of `initialize()` method, which sets those important properties and could be called from the constructor as well as from `__wakeup()`. This would avoid potential future problems when new properties are introduced and we would have to remember adding them to the constructor as well as to `__sleep()`.

stweil:
I don't have a personal preference. Any solution which fixes the current regression is fine for me.

sebastian-meyer (approved):

PR #1537, opened 2025-03-17
[MAINTENANCE] Use local OAI-PMH.xsd for CI tests

### LINKED ISSUE #1535: [BUG] CI unit tests fail (HTTP/1.1 403 Forbidden)
## Description

The CI unit tests currently fail:
```
1) Kitodo\Dlf\Tests\Unit\Validation\DOMDocumentValidationStackTest::testBreakOnError
DOMDocument::schemaValidateSource(https://www.openarchives.org/OAI/2.0/OAI-PMH.xsd): Failed to open stream: HTTP request failed! HTTP/1.1 403 Forbidden
```

I get the same error 403 when I try to get the URL with curl (it works fine in my browser):
``` sh
% curl -I https://www.openarchives.org/OAI/2.0/OAI-PMH.xsd
HTTP/2 403 
server: awselb/2.0
date: Sun, 16 Mar 2025 09:44:09 GMT
content-type: text/html
content-length: 118
```

So Kitodo should use a local copy of https://www.openarchives.org/OAI/2.0/OAI-PMH.xsd for the validation.

stweil:
This is required because the URL no longer works in scripts.

### DIFF
--- a/Tests/Unit/Validation/DOMDocumentValidationStackTest.php
+++ b/Tests/Unit/Validation/DOMDocumentValidationStackTest.php
@@ -69,7 +69,16 @@ class DOMDocumentValidationStackTest extends UnitTestCase
 
     public function testBreakOnError(): void
     {
-        $xmlSchemesValidatorConfig = ["className" => XmlSchemesValidator::class, "configuration" => [["namespace" => "http://www.openarchives.org/OAI/2.0/", "schemaLocation" => "https://www.openarchives.org/OAI/2.0/OAI-PMH.xsd"]]];
+        $xsdFile = __DIR__ . '/../../Fixtures/OaiPmh/OAI-PMH.xsd';
+        $xmlSchemesValidatorConfig = [
+            'className' => XmlSchemesValidator::class,
+            'configuration' => [
+                [
+                    'namespace' => 'http://www.openarchives.org/OAI/2.0/',
+                    'schemaLocation' => $xsdFile
+                ]
+            ]
+        ];
         $domDocumentValidationStack = new DOMDocumentValidationStack([$xmlSchemesValidatorConfig, $xmlSchemesValidatorConfig]);
         $result = $domDocumentValidationStack->validate(new DOMDocument());
         self::assertCount(1, $result->getErrors());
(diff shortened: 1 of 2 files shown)


stweil:
This fixes the "Unit and Functional Testing" which currently fails in most cases.

stweil:
Codacy does not like my new code and creates 4 issues for it. Is there a better solution, or should we just ignore these issues?

beatrycze-volk:
> Codacy does not like my new code and creates 4 issues for it. Is there a better solution, or should we just ignore these issues?

I have looked at the errors and there are all related to access to the file. 

Maybe we should place this file into `Test/Fixtures/` directory and then access its content similar like it is done in e.g. `AltoTest` class (`simplexml_load_file(__DIR__ . '/../../Fixtures/Format/alto.xml');`)?

I'm also ok with ignoring those errors, as they are in the test class.

Resources/Private/Data/OAI-PMH.xsd
sebastian-meyer, line 1: Please move this file to `Tests/Fixtures/OaiPmh/`.

Tests/Unit/Validation/DOMDocumentValidationStackTest.php
```diff
@@ -69,7 +69,8 @@ public function testValidatorDerivation(): void
 
     public function testBreakOnError(): void
     {
-        $xmlSchemesValidatorConfig = ["className" => XmlSchemesValidator::class, "configuration" => [["namespace" => "http://www.openarchives.org/OAI/2.0/", "schemaLocation" => "https://www.openarchives.org/OAI/2.0/OAI-PMH.xsd"]]];
+        $xsdFile = realpath($_SERVER['DOCUMENT_ROOT'] . 'Resources/Private/Data/OAI-PMH.xsd');
+        $xmlSchemesValidatorConfig = ["className" => XmlSchemesValidator::class, "configuration" => [["namespace" => "http://www.openarchives.org/OAI/2.0/", "schemaLocation" => $xsdFile]]];
```
sebastian-meyer:
```suggestion
        $xsdFile = __DIR__ . '/../../Fixtures/OaiPmh/OAI-PMH.xsd';
        $xmlSchemesValidatorConfig = [
            'className' => XmlSchemesValidator::class,
            'configuration' => [
                [
                    'namespace' => 'http://www.openarchives.org/OAI/2.0/',
                    'schemaLocation' => $xsdFile
                ]
            ]
        ];
```

sebastian-meyer (changes requested):
I agree with @beatrycze-volk.

Thank you for fixing this issue!

Resources/Private/Data/OAI-PMH.xsd
stweil: Done now.

sebastian-meyer (approved):
Thanks a lot!

PR #1539, opened 2025-03-18
[MAINTENANCE] Allow installation with PHP 8.4

stweil:


stweil:
See related changes https://github.com/slub/slub_digitalcollections/pull/128 and https://github.com/slub/dfg-viewer/pull/353.

With these minimal changes it is possible to install a working DFG Viewer with TYPO3 v12 on hosts with PHP 8.4. This is very helpful for my development even if is not an officially supported configuration.

stweil:
All changes are now in the main branch.

PR #1560, opened 2025-03-26
[FEATURE] Always show OAI-PMH request type

stweil:


stweil:
The old stylesheet did not show the verb in an error case. I also use an additional line in the existing table to show the request type because I think it looks nicer.

sebastian-meyer:
I think, originally the `verb` was omitted in error messages, because the error could be shown because of a missing `verb`. Did you check how the XSLT behaves when `verb` is missing?

stweil:
Yes, a missing verb simply shows an empty request type. See [example](https://digital.ladenburg.world/oai).

sebastian-meyer (approved):

PR #1566, opened 2025-04-03
[BUGFIX] Sync listview with facets and fix wrong hitcount

### LINKED ISSUE #1189: [BUG] Number of hits for facets incorrect
## Number of hits for facets incorrect

In the facets contained on the page, an incorrect number is displayed for the number of hits in almost all cases.

If the respective link is clicked, the number of hits corresponds to a separate Solr enquiry.

As far as could be determined by debugging, the numbers returned by the following lines are not correct:

```
// check for solr response
$solrRequest = $solr->service->createRequest($selectQuery);
$response = $solr->service->executeRequest($solrRequest);
```
contained in Classes/Controler/SearchControler.php 

## Reproduction

1. Call up any collection
2. In the facets contained on the page, an incorrect number is displayed for the number of hits in almost all cases.

## Expected Behavior

The number of hits displayed after a facet should correspond to the number of Solr queries.

### Example of facets
image

### Difference between number of hits for facets and results page when clicking on the facet link
image

## Environment

- OS version: Debian GNU/Linux 11 (bullseye)
 - RDBMS version: MySQL 5.5.5-10.5.23-MariaDB-0+deb11u1 
 - Apache Solr version: 8.11.2
 - TYPO3 version: 10.4.38
 - PHP version: 7.4.33

michaelkubina:
This fixes issues related to the facets in combination with the listview:
* removes fulltext pages from the facet count in metadata searches
* does actually a wildcard search when entering a collection (or performing an empty search), so that the facets and listview are in sync now, as the facets are already wildcard in that case.

Fixes: [BUG] Number of hits for facets incorrect #1189

### DIFF
--- a/Classes/Common/Solr/SolrSearch.php
+++ b/Classes/Common/Solr/SolrSearch.php
@@ -402,7 +402,7 @@ class SolrSearch implements \Countable, \Iterator, \ArrayAccess, QueryResultInte
             }
             $params['fulltext'] = true;
         } else {
-            $params['filterquery'][]['query'] = "-type:page";
+            $params['filterquery'][]['query'] = '-type:page';
             // Retain given search field if valid.
             if (!empty($this->searchParams['query'])) {
                 $query = Solr::escapeQueryKeepField(trim($this->searchParams['query']), $this->settings['storagePid']);
--- a/Classes/Common/Solr/SolrSearch.php
+++ b/Classes/Common/Solr/SolrSearch.php
@@ -402,9 +402,12 @@ class SolrSearch implements \Countable, \Iterator, \ArrayAccess, QueryResultInte
             }
             $params['fulltext'] = true;
         } else {
+            $params['filterquery'][]['query'] = "-type:page";
             // Retain given search field if valid.
             if (!empty($this->searchParams['query'])) {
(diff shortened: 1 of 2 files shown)


Classes/Common/Solr/SolrSearch.php
```diff
@@ -402,9 +402,12 @@ public function prepare()
             }
             $params['fulltext'] = true;
         } else {
+            $params['filterquery'][]['query'] = "-type:page";
```
sebastian-meyer:
```suggestion
            $params['filterquery'][]['query'] = '-type:page';
```

Classes/Controller/SearchController.php
```diff
@@ -285,6 +285,7 @@ private function makeFacetsMenuArray(array $facets): array
                 $search['query'] = $fields['fulltext'] . ':(' . Solr::escapeQuery(trim($searchParams['query'])) . ')';
             }
         } else {
+            $search['params']['filterquery'][]['query'] = "-type:page";
```
sebastian-meyer:
```suggestion
            $search['params']['filterquery'][]['query'] = '-type:page';
```

sebastian-meyer (dismissed):

sebastian-meyer (approved):

Issue #1567, opened 2025-04-03
[BUG] Not working metadata wrapper


beatrycze-volk:
## Description

Metadata wrapper for 'author' metadata does not work. There exists script for displaying the links for authors if the MODS contains URI attribute. Check for it happens in TypoScript, but this TypoScript contains not existing command.  It was working in previous versions because TYPO3 was from itself not generating links if the URI is empty. Since 11.x such cases are handled differently. Currently if the URI is empty, the current URL is used as `href` value.

### TypoScript

```
key.wrap = <dt class=\"tx-dlf-metadata-author\">|</dt>
value.required = 1
value.split.token.char = 31
value.split.cObjNum = 1
value.split.1.1 = CASE
value.split.1.1.key.data = register:SPLIT_COUNT
value.split.1.1.0 = LOAD_REGISTER
value.split.1.1.0.tx_dlf_metadata_author_name.current = 1
value.split.1.1.1 = LOAD_REGISTER
value.split.1.1.1.tx_dlf_metadata_author_uri.current = 1
value.postCObject = TEXT
value.postCObject.value = {register:tx_dlf_metadata_author_name}
value.postCObject.value.insertData = 1
value.postCObject.value.stdWrap.typolink.parameter = {register:tx_dlf_metadata_author_uri} _blank external
value.postCObject.value.stdWrap.typolink.parameter.insertData = 1
value.postCObject.value.stdWrap.typolink.title = {register:tx_dlf_metadata_author_name}
value.postCObject.value.stdWrap.typolink.ifNotEmpty = 1
value.wrap = <dd class=\"tx-dlf-metadata-author\">|</dd>
```

`typolink.ifNotEmpty` does not exists. I have checked out TYPO3 documentation for few last versions and none of them has such a command.

## Expected Behavior

Link tag should only be generated if URI is not empty.

## Environment

 - TYPO3 version: 11.x

## Additional Context

Metadata wrappers are relics from previous versions of TYPO3. Currently would be good to replace this parts with fluid templates. It would also decrease the complexity for Metadata plugin.

@sebastian-meyer what do you think?

sebastian-meyer:
You are right, `typolink` has no `ifNotEmpty` function, but `typolink.parameter` allows `stdWrap` and should therefore support `ifNotEmpty`. So maybe that would be an easy fix.

But in general I agree: Using Fluid templates is a way better solution!

PR #1568, opened 2025-04-03
[BUGFIX] Display author URI as an external link

beatrycze-volk:
Related to #1567

### DIFF
--- a/Classes/Format/Mods.php
+++ b/Classes/Format/Mods.php
@@ -173,11 +173,16 @@ class Mods implements MetadataInterface
                 $k++;
             }
             ksort($name);
+
             $this->metadata['author'][$i] = trim(implode(', ', $name));
         }
-        // Append "valueURI" to name using Unicode unit separator.
+
         if (isset($authors[$i]['valueURI'])) {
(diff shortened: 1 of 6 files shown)


Classes/Format/Mods.php
```diff
@@ -173,11 +173,16 @@ private function getAuthorFromXml(array $authors, int $i): void
...
         }
-        // Append "valueURI" to name using Unicode unit separator.
+
         if (isset($authors[$i]['valueURI'])) {
-            $this->metadata['author'][$i] .= pack('C', 31) . (string) $authors[$i]['valueURI'];
+            $displayName = $this->metadata['author'][$i];
+            $this->metadata['author'][$i] = [
+                'name' => $displayName,
+                'url' => (string) $authors[$i]['valueURI']
+            ];
```
sebastian-meyer, line 185: How does this affect the indexing of author names? You adjusted the templates for display, but I don't see any changes to the indexing routines. Are those not affected?

beatrycze-volk: Thanks for comment. I didn't consider this case. I will look into it.

sebastian-meyer (approved):
Thanks you!

PR #1573, opened 2025-04-05
[MAINTENANCE] Fix some issues reported by PHPStan

stweil:


Classes/Common/Helper.php
```diff
@@ -258,52 +256,71 @@ public static function getXmlFileAsString($content)
         if (\PHP_VERSION_ID < 80000) {
```
stweil: Should we remove the code for PHP_VERSION_ID < 80000?

beatrycze-volk: Sure, now it is not needed anymore.

stweil: Done now.

Classes/Domain/Repository/DocumentRepository.php
```diff
@@ -36,7 +36,7 @@
  *
  * @access public
  */
-class DocumentRepository extends Repository
+class DocumentRepository extends \TYPO3\CMS\Extbase\Persistence\Repository
```
sebastian-meyer: Which issue is addressed by this change? Since we declare `use TYPO3\CMS\Extbase\Persistence\Repository;` above we shouldn't need to add the namespace here, should we?

stweil:
I rebased this PR now and fixed some merge conflicts.

sebastian-meyer (approved):

PR #1578, opened 2025-04-14
[FEATURE] Validation form plugin

markusweigelt:
- Validation form plugin
- Adjusting middleware to support warnings and notice additionally to errors 
- Improvement of configuration from middleware 
  - It is now possible to create several validation configuration types and validate against them
- i18n

image


ext_localconf.php
```diff
@@ -382,6 +383,18 @@
     ]
 );
 
+\TYPO3\CMS\Extbase\Utility\ExtensionUtility::configurePlugin(
+    'Dlf',
+    'ValidationForm',
+    [
+        \Kitodo\Dlf\Controller\ValidationFormController::class => 'main',
```
sebastian-meyer, line 390: Shouldn't the validation form be non-cacheable?

markusweigelt: The plugin provides a container for requesting the validation middleware and displaying the validation response. This is primarily handled by JavaScript on the client side. The plugin's form itself can currently be cached, or rather I have not encountered any caching issues so far. If the plugin configuration is changed while being used as a content component, the page must be saved, which will automatically clear the page cache.

sebastian-meyer: Thanks for clarification!

sebastian-meyer (approved):

PR #1588, opened 2025-05-12
[BUGFIX] Use correct function for search features

beatrycze-volk:
Review from @fschoelzel needed


beatrycze-volk:
@sebastian-meyer : As @fschoelzel will be gone for next month so this change should be merged and if it will make some problems in DDB Zeitungsportal, then this JS should be adjusted separately in https://github.com/slub/ddb_kitodo_zeitungsportal

sebastian-meyer (approved):

PR #1623, opened 2025-05-23
[BUGFIX] Replace magic getter for `toplevelId` property

beatrycze-volk:
This magic function is actually not magic as it only works when it is called directly, otherwise the property is empty


sebastian-meyer (approved):
Technically this is fine, but out of curiosity: Why? What's the problem with the magic getter?

PR #1627, opened 2025-05-28
[OPTIMIZATION] Speed up pagegrid generation

michaelkubina:
The pagegrids initial loading time is very slow on very large documents and/or those with many filegroups. On a document with several thousand pages (e.g. Adressbücher, Realkataloge, Plenarprotokolle, Staatsstatistik) and seven filegroups the initial generation of the pagegrid took > 60 seconds resulting in script abortion. This is because the functions `getFileLocation()` xpath does not differentiate which filegroup to check, resulting in traversing through all filegroups, even though they are not of interest for the pagegrid. While we potentially only need to check one filegroup, the xpath actually checks all seven.

This PR introduces a new function `getFileLocationInUsegroup()`, where the xpath considers the `@USE` attribute, which reduces exection time for this function to roughly 1/number_of_filegroups (when compared to `getFileLocation()`), and applies it to the pagegrid controller. This will speed up execution time significantly for all documents and (except for maybe really extreme document sizes) prevent script timeouts.

### DIFF
--- a/Classes/Common/AbstractDocument.php
+++ b/Classes/Common/AbstractDocument.php
@@ -336,7 +336,7 @@ abstract class AbstractDocument
      * @abstract
      *
      * @param string $id The "@ID" attribute of the file node (METS) or the "@id" property of the IIIF resource
-     * 
+     *
      * @param string $useGroup The "@USE" attribute of the fileGrp node (METS)
      *
      * @return string The file's location as URL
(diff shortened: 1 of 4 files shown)


Classes/Common/AbstractDocument.php
```diff
@@ -328,6 +328,21 @@ abstract public function getFileInfo($id): ?array;
...
 
+    /**
+     * This gets the location of a file representing a physical page or track
+     *
+     * @access public
+     *
+     * @abstract
+     *
+     * @param string $id The "@ID" attribute of the file node (METS) or the "@id" property of the IIIF resource
+     *
```
```suggestion
     *
```

sebastian-meyer (approved):

PR #1628, opened 2025-06-02
[BUGFIX] Add option for inserting all cookies into request

beatrycze-volk:
During the update between version cookies logic for the request has changed.


sebastian-meyer (approved):

sebastian-meyer (changes requested):
Please check the Codacy issues.

beatrycze-volk:
Codacy complain ist incorrect, as the third paramater is optional: https://www.php.net/manual/en/function.filter-input.php

sebastian-meyer (approved):

Issue #1550, opened 2025-03-20
[BUG] Indexing reports missing required metadata field "record_id"


stweil:
## Description

All of our METS files report a missing required metadata field "record_id" although they contain a `mods:recordIdentifier` tag.
METS files from SLUB Dresden seem to raise the same error message:
```
vendor/bin/typo3 kitodo:index --pid 3 --solr dlfCore2 --doc https://digital.slub-dresden.de/data/kitodo/Abenaud_399196951-1823/Abenaud_399196951-1823_mets.xml -vvv
Cannot load Xdebug - it was already loaded

Index single document into database and Solr.
=============================================

Indexing 13 ("https://digital.slub-dresden.de/data/kitodo/Abenaud_399196951-1823/Abenaud_399196951-1823_mets.xml") on PID 3.
----------------------------------------------------------------------------------------------------------------------------

Thu, 20 Mar 2025 11:45:24 +0000 [ERROR] request="406fb3f8d96cd" component="Kitodo.Dlf.Validation.DocumentValidator": Missing required metadata field "record_id". 
Indexing 13 ("https://digital.slub-dresden.de/data/kitodo/Abenaud_399196951-1823/Abenaud_399196951-1823_mets.xml") on Solr core 1.
----------------------------------------------------------------------------------------------------------------------------------

Thu, 20 Mar 2025 11:45:24 +0000 [ERROR] request="406fb3f8d96cd" component="Kitodo.Dlf.Validation.DocumentValidator": Missing required metadata field "record_id". 

 [ERROR] ERROR: Document with UID "13" could not be indexed on Solr core 1. There are missing mandatory fields (at least
         one of those: document_format,record_id) in this document.
```

stweil:
@markusweigelt, do you have a suggestion where I should look to fix this issue? Which part of the code is relevant for setting `record_id`? As far as I know the `metadata` table has the required information, but I don't know where this information is used and why it does not set the value.

markusweigelt:
@stweil As I understood it, you need to add metadata with the name `record_id` under List -> Kitodo-Configuration -> Metadata, select `MODS` as the data format, and set the XPath to the identifier: `./mods:recordInfo/mods:recordIdentifier`.

Here is a screenshot of the input form:
![Image](https://github.com/user-attachments/assets/0d7ec34d-eec4-44af-a6cf-b90e57feebf8)

[...]

PR #1630, opened 2025-06-04
[BUGFIX] Display correct error message for indexing

beatrycze-volk:
The error message for field validation should be displayed if save to database was not successful as this validation. 

Fail on save to index can be causes by many reasons. Those reasons should be verified in error log, validation can be one of them. The other one can be fail to connect with SOLR Index. Actually it was the case in which I was getting this error message. In the TYPO3 log was other error: `Could not connect to Apache Solr server`

It could be related to #1550

### DIFF
--- a/Classes/Common/Indexer.php
+++ b/Classes/Common/Indexer.php
@@ -397,7 +397,8 @@ class Indexer
                     return false;
                 }
             } else {
-                Helper::log('There are missing mandatory fields (at least one of those: ' . $extConf['requiredMetadataFields'] . ') in this document', LOG_SEVERITY_ERROR);
+                Helper::error('There are missing mandatory fields (at least one of those: ' . $extConf['requiredMetadataFields'] . ') in this document');
+                Helper::notice('Tip: If "record_id" field is missing then there is possibility that METS file still contains it but with the wrong source type attribute in "recordIdentifier" element');
                 return false;
             }
         }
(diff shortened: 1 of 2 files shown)


sebastian-meyer (approved):

sebastian-meyer (changes requested):
Please check PHPStan issue.

sebastian-meyer (approved):

Classes/Common/Indexer.php
```diff
@@ -397,6 +397,7 @@ protected static function processLogical(Document $document, array $logicalUnit)
                     return false;
                 }
             } else {
+                Helper::error('There are missing mandatory fields (at least one of those: ' . $extConf['requiredMetadataFields'] . ') in this document');
                 Helper::notice('Tip: If "record_id" field is missing then there is possibility that METS file still contains it but with the wrong source type attribute in "recordIdentifier" element');
```
sebastian-meyer, line 401: Prior to your last change, you removed the notice in favor of the error message. Now you left it in and just added the error. Was that on purpose?

beatrycze-volk: Actually I didn't remove it in my original PR. This `Helper::notice` was inserted in one of PRs which were merged yesterday. I find this tip useful (thank you @stweil ), so now both log messages should be displayed.

sebastian-meyer (approved):

PR #1635, opened 2025-06-12
[MAINTENANCE] Insert metadata format in Kitodo storage

beatrycze-volk:


beatrycze-volk:
Codacy warning:

`All output should be run through an escaping function (see the Security sections in the WordPress Developer Handbooks), found '$e'.`

I suspect this error is false positive here.

sebastian-meyer (approved):

PR #1637, opened 2025-06-13
[MAINTENANCE] Adjust SQL queries

beatrycze-volk:


Classes/Common/Indexer.php
```diff
@@ -266,19 +266,19 @@ protected static function loadIndexConf(int $pid): void
...
-                    'tx_dlf_metadata.is_facet AS is_facet',
-                    'tx_dlf_metadata.is_listed AS is_listed',
-                    'tx_dlf_metadata.index_autocomplete AS index_autocomplete',
-                    'tx_dlf_metadata.index_boost AS index_boost'
+                    'index_indexed',
+                    'is_sortable',
+                    'is_facet',
+                    'is_listed',
+                    'index_autocomplete',
+                    'index_boost'
```
sebastian-meyer: Is there a reason why you adjusted all but the third line?

Classes/Domain/Repository/DocumentRepository.php
```diff
@@ -789,18 +785,18 @@ public function getFirstChild($uid)
...
 
         $child = $queryBuilder
             ->select(
-                'tx_dlf_documents.uid AS uid'
+                'uid'
             )
             ->from('tx_dlf_documents')
             ->where(
-                $queryBuilder->expr()->eq('tx_dlf_documents.partof', (int) $uid)
+                $queryBuilder->expr()->eq('partof', (int) $uid)
```
sebastian-meyer:
```suggestion
                $queryBuilder->expr()->eq('partof', $uid)
```
You can omit the type casting as you did in all the other methods. ;o)

Classes/Domain/Repository/DocumentRepository.php
```diff
@@ -789,18 +785,18 @@ public function getFirstChild($uid)
...
             )
             ->add('orderBy', 'volume_sorting desc')
             ->addOrderBy('tx_dlf_documents.volume_sorting')
```
sebastian-meyer:
```suggestion
            ->add('orderBy', 'volume_sorting desc')
            ->addOrderBy('volume_sorting')
```
As a side note: Aren't those two statements redundant? (Which applies to almost all methods of this class...)

sebastian-meyer (changes requested):

Classes/Domain/Repository/DocumentRepository.php
beatrycze-volk: I was also thinking about it. Just the first one contains `desc` an second one not. I will check it out.

Classes/Common/Indexer.php
beatrycze-volk: No reason, overlooked.

sebastian-meyer (approved):

PR #1640, opened 2025-06-16
[BUGFIX] Initialize properties in `AbstractController` depending on application context

### LINKED ISSUE #1553: [BUG] Getting `PageUID` in `AbstractController` is broken (regression)
## Description

"New Tenant" no longer works in the TYPO3 backend. It was broken in pull request ##1523.

beatrycze-volk:
Closes #1553

PHPStan fails as `$request` has function `getAttribute()` only in TYPO3 12.

For TYPO3 11 this function is missing

ActionController:
```
    /**
     * The current request.
     *
     * @var Request
     * @todo v12: Change @var to RequestInterface, when RequestInterface extends ServerRequestInterface
     */
    protected $request;
```

`ServerRequestInterface` provides the function `getAttribute()`.

I have checked my previous changes in this topic and there I was using `$this->request->getAttribute('routing')->getPageId()` instead of `RequestInterface $request` which is passed as parameter into the function. As the `Request` class implements `ServerRequestInterface`. 

@sebastian-meyer Should I use `$this->request` to keep compatibility with TYPO3 11?


sebastian-meyer:
> @sebastian-meyer Should I use `$this->request` to keep compatibility with TYPO3 11?

Yes, current main branch has to support TYPO3 v11 and v12.

beatrycze-volk:
This change causes failure in `OaiPmhTest`, it is not easy to explain as this test class doesn't use any controller class.

Error is thrown directly in the line:
`$oai = Endpoint::build($this->oaiUrl);`

After tracing the error origin I have got into `Phpoaipmh/Client` class:

```
try {
    $xml = @new \SimpleXMLElement($resp);
} catch (\Exception $e) {
    throw new MalformedResponseException(sprintf("Could not decode XML Response: %s", $e->getMessage()));
}
```

Actually I ask myself why do we test `Endpoint` class instead testing our own `OaiPmh` controller.

I can't find any logic explanation why this change causes this situation.

sebastian-meyer:
@chrizzor, @haogatyp, @oliver-stoehr Can you give some insight regarding the `OaiPmhTest`, please? What was your reasoning behind testing `Endpoint` instead of the controller? Do you think we could drop the test in favor of a test for `OaiPmhController` without losing overall test coverage?

sebastian-meyer:
I am not sure how to proceed with this. Shall I merge it, altough it contains inactive code for TYPO3 v11? What would be the consequences? Does the assignment in line 125 sufficiently work for TYPO3v11?

beatrycze-volk:
> I am not sure how to proceed with this. Shall I merge it, altough it contains inactive code for TYPO3 v11? What would be the consequences? Does the assignment in line 125 sufficiently work for TYPO3v11?

Yes, assignment from line 125 works correctly for TYPO3 11.

I have removed commented code.

sebastian-meyer:
Thank you very much!

sebastian-meyer (approved):

Classes/Controller/AbstractController.php
```diff
@@ -124,6 +125,12 @@ protected function initialize(RequestInterface $request): void
         $this->pageUid = (int) GeneralUtility::_GET('id');
         $this->requestData['page'] = $this->requestData['page'] ?? 1;
 
+        if (GeneralUtility::makeInstance(Typo3Version::class)->getMajorVersion() >= 12) {
+            if ($request->getAttribute('applicationType') === 1) {
```
stweil, line 129: @beatrycze-volk, @sebastian-meyer, it looks like this code is not in the main branch. Is it not needed there?

sebastian-meyer: See #1716

PR #1654, opened 2025-06-20
[BUGFIX] Force reload document in `BaseCommand`

beatrycze-volk:


sebastian-meyer (approved):
Technically this is fine, but why do we have to reload the parent document every time its ID is needed?

beatrycze-volk:
Generally we want to have all fresh documents, also the parent document for indexing. Indexer should not access any cached documents, as it cause problems during the reindexing of documents.

sebastian-meyer:
This will lead to re-parsing the parent document again and again when indexing multiple volumes of the same multi-volume work.
But you are right: obviously we always want to ingest the latest version of the documents.

PR #1658, opened 2025-06-23
Revert "[BUGFIX] Iterate over documents inside `SolrSearch`"

beatrycze-volk:
Reverts kitodo/kitodo-presentation#1598

It looks that this change is not triggering `offsetGet($offset)` function.


beatrycze-volk:
Tests should be restarted.

sebastian-meyer (approved):

PR #1660, opened 2025-07-04
[BUGFIX] Fixes navigation URL encoding and measure navigation

markusweigelt:
- Implement workaround cause ViewHelper f:link.action / UriBuilder does not properly encode specified entities in URL parameter
  - For more details, please see the following TYPO3 issue https://forge.typo3.org/issues/107026
- Fix measure navigation - cause behavior of jumping to the next measure page did not work

### DIFF
--- a/Classes/Controller/AbstractController.php
+++ b/Classes/Controller/AbstractController.php
@@ -140,6 +140,8 @@ abstract class AbstractController extends ActionController implements LoggerAwar
             'requestData' => $this->requestData
         ];
 
+        // Workaround cause ViewHelper f:link.action / UriBuilder does not properly encode specified entities in URL parameter
+        // For more details, please see the following TYPO3 issue https://forge.typo3.org/issues/107026
         if( isset($this->requestData['id']) ) {
            $this->viewData['partlyEncodedId'] = str_replace("%2F", "%252F", $this->requestData['id']);
         }
--- a/Classes/Controller/AbstractController.php
+++ b/Classes/Controller/AbstractController.php
@@ -140,7 +140,7 @@ abstract class AbstractController extends ActionController implements LoggerAwar
             'requestData' => $this->requestData
         ];
 
-        // Workaround cause ViewHelper f:link.action / UriBuilder does not properly encode specified entities in URL parameter
+        // TODO: ViewHelper f:link.action / UriBuilder does not properly encode specified entities in URL parameter
         // For more details, please see the following TYPO3 issue https://forge.typo3.org/issues/107026
         if ( isset($this->requestData['id']) ) {
             $this->viewData['partlyEncodedId'] = str_replace("%2F", "%252F", $this->requestData['id']);
--- a/Classes/Controller/AbstractController.php
+++ b/Classes/Controller/AbstractController.php
@@ -140,6 +140,10 @@ abstract class AbstractController extends ActionController implements LoggerAwar
             'requestData' => $this->requestData
         ];
 
+        if( isset($this->requestData['id']) ) {
+           $this->viewData['partlyEncodedId'] = str_replace("%2F", "%252F", $this->requestData['id']);
+        }
+
         try {
             $this->viewData['publicResourcePath'] = PathUtility::getPublicResourceWebPath('EXT:dlf/Resources/Public');
         } catch (InvalidFileException) {
--- a/Classes/Controller/AbstractController.php
+++ b/Classes/Controller/AbstractController.php
@@ -142,8 +142,8 @@ abstract class AbstractController extends ActionController implements LoggerAwar
 
         // Workaround cause ViewHelper f:link.action / UriBuilder does not properly encode specified entities in URL parameter
         // For more details, please see the following TYPO3 issue https://forge.typo3.org/issues/107026
-        if( isset($this->requestData['id']) ) {
-           $this->viewData['partlyEncodedId'] = str_replace("%2F", "%252F", $this->requestData['id']);
+        if ( isset($this->requestData['id']) ) {
+            $this->viewData['partlyEncodedId'] = str_replace("%2F", "%252F", $this->requestData['id']);
         }
 
         try {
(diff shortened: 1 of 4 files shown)


sebastian-meyer:
I am not quite sure if this is the best possible solution.

First of all, the workaround only works for the special case where an URL contains an already encoded `/`, but does not account for any other encoded character:

`$this->viewData['partlyEncodedId'] = str_replace("%2F", "%252F", $this->requestData['id']);`

I'd like to see a more general approach here which addresses all kinds of pre-encoded entities, not just `%2F`. Something like this:

`$this->viewData['partlyEncodedId'] = rawurlencode($this->requestData['id']);`

This would replace `%2F` with `%25F` as intended, but also encode everything else. But I am not sure what the `UriBuilder` does with it...

Classes/Controller/NavigationController.php
```diff
@@ -135,8 +135,11 @@ public function mainAction(): ResponseInterface
                 $musicalStructure = $this->document->getCurrentDocument()->musicalStructure;
                 $musicalStructureInfo = $this->document->getCurrentDocument()->musicalStructureInfo;
                 for ($i = 1; $i <= $this->document->getCurrentDocument()->numMeasures; $i++) {
-                    if (array_key_exists($i, $musicalStructure) && array_key_exists($i, $musicalStructureInfo)) {
-                        $measureOptions[$i] = '[' . $i . ']' . ($musicalStructureInfo[$musicalStructure[$i]['measureid']]['orderlabel'] ? ' - ' . htmlspecialchars($musicalStructureInfo[$musicalStructureInfo[$i]]['orderlabel']) : '');
+                    if (isset($musicalStructure[$i])
+                        && array_key_exists($i, $musicalStructure)
+                        && isset($musicalStructureInfo)
+                        && array_key_exists($musicalStructure[$i]['measureid'], $musicalStructureInfo)) {
+                        $measureOptions[$i] = '[' . $i . ']' . ($musicalStructureInfo[$musicalStructure[$i]['measureid']]['orderlabel'] ? ' - ' . htmlspecialchars($musicalStructureInfo[$musicalStructure[$i]['measureid']]['orderlabel']) : '');
```
sebastian-meyer: This is quite redundant. If `$musicalStructure[$i]` is set we do not need to check if the array key `$i` exists, because it obviously does.
Instead, we need to check for array keys `measureid` and `orderlabel` before accessing them.

```suggestion
                    if (isset($musicalStructure[$i]['measureid'])
                        && isset($musicalStructureInfo[$musicalStructure[$i]['measureid']]['orderlabel'])) {
                        $measureOptions[$i] = '[' . $i . ']' . (!empty($musicalStructureInfo[$musicalStructure[$i]['measureid']]['orderlabel']) ? ' - ' . htmlspecialchars($musicalStructureInfo[$musicalStructure[$i]['measureid']]['orderlabel']) : '');
```

Classes/Controller/PageViewController.php
```diff
@@ -523,10 +523,17 @@ protected function addViewerJS(): void
                 });';
         } else {
             $currentMeasureId = '';
-            $docPage = $this->requestData['page'] ?? 0;
+            $docPage = 0;
+
+            if (isset($this->requestData['page'])) {
+                $docPage = $this->requestData['page'];
+            }
```
sebastian-meyer, line 530: Why change this? It does exactly the same, but is more difficult to read...

Classes/Controller/PageViewController.php
```diff
@@ -523,10 +523,17 @@ protected function addViewerJS(): void
...
 
             $docMeasures = $this->getMeasures($docPage);
-            if ($this->requestData['measure'] ?? false) {
+            if (isset($docMeasures['measureCounterToMeasureId'])
+                && count($docMeasures['measureCounterToMeasureId']) > 0
+                && isset($this->requestData['measure'])
+                && array_key_exists($this->requestData['measure'], $docMeasures['measureCounterToMeasureId'])) {
```
sebastian-meyer: Why so many redundant checks? Wouldn't it be enough to just check for `isset($docMeasures['measureCounterToMeasureId'][$this->requestData['measure']])`?

At least we should omit the `count(...)`. Or am I missing something?

markusweigelt:
> I am not quite sure if this is the best possible solution.
> 
> First of all, the workaround only works for the special case where an URL contains an already encoded `/`, but does not account for any other encoded character:

Yes, that’s true—I’m not entirely satisfied with the solution either, but right now I have no idea how to fix it with a simpler approach.

> I'd like to see a more general approach here which addresses all kinds of pre-encoded entities, not just `%2F`. Something like this:
> 
> `$this->viewData['partlyEncodedId'] = rawurlencode($this->requestData['id']);`
> 
> This would replace `%2F` with `%25F` as intended, but also encode everything else. But I am not sure what the `UriBuilder` does with it...

The URIBuilder encodes this URL again, and then it can’t be found. Unfortunately, you can’t disable the encoding for the ViewHelper, and that won’t happen any time soon (see TYPO3 issue https://forge.typo3.org/issues/107026).

Classes/Controller/PageViewController.php
markusweigelt: Change this cause of PHPStan error https://github.com/kitodo/kitodo-presentation/actions/runs/16074314340. I thought that something had change with the configuration so i adjusted this.

Classes/Controller/NavigationController.php
markusweigelt: Thanks — I don't really know why I added so many checks here either. You're right, it's redundant.

Classes/Controller/PageViewController.php
markusweigelt: I will adjust these checks. Have no clue what got into me here 😄

sebastian-meyer (dismissed):

Classes/Controller/AbstractController.php
```diff
@@ -140,6 +140,12 @@ protected function initialize(RequestInterface $request): void
             'requestData' => $this->requestData
         ];
 
+        // Workaround cause ViewHelper f:link.action / UriBuilder does not properly encode specified entities in URL parameter
+        // For more details, please see the following TYPO3 issue https://forge.typo3.org/issues/107026
+        if ( isset($this->requestData['id']) ) {
```
```suggestion
        // TODO: ViewHelper f:link.action / UriBuilder does not properly encode specified entities in URL parameter
        // For more details, please see the following TYPO3 issue https://forge.typo3.org/issues/107026
        if ( isset($this->requestData['id']) ) {
```

sebastian-meyer (approved):

Classes/Controller/PageViewController.php
sebastian-meyer: I don't understand why PHPStan is complaining about this, but OK... ;o)

PR #1673, opened 2025-07-22
[BUGFIX] Parsing timecode in TableOfContent Chapter-Links for A/V Media Files

fschoelzel:
- accept new `uuid` format in RegEx for `fileIds`
- add possibility to parse Milliseconds in timecode like: `123.456` total seconds


Resources/Private/JavaScript/SlubMediaPlayer/SlubMediaPlayer.js
```diff
@@ -253,8 +253,10 @@ export default class SlubMediaPlayer extends DlfMediaPlayer {
       }
     }
 
-    // Attempt: Parse timecode hash in URL ("#timecode=120;fileId=FILE_0000_DEFAULT_MPD,FILE_0000_DEFAULT_HLS")
-    const timecodeMatch = link.hash.match(/timecode=(\d+(?:\.\d?)?)(?:;fileIds=([a-zA-Z0-9_,]+))?/);
+    // Attempt: Parse timecode hash in URL. Accepts URLs with hashes like:
+    // #timecode=120;fileIds=FILE_0000_DEFAULT_MPD,FILE_0000_DEFAULT_HLS
+    // #timecode=123.456;fileIds=uuid-0aabcd1e-12a3-xxx,uuid-0aabcd1e-12a3-yyy
+    const timecodeMatch = link.hash.match(/timecode=(\d+(?:\.\d+)?)(?:;fileIds=([\w,-]+))?/u);
```
sebastian-meyer: If those time codes and file IDs are supposed to be passed as URI fragments only, then we should include the hash mark (`#`) in the regular expression as well. This will prevent arbitrary matches in other URI parts.

sebastian-meyer (changes requested):

Resources/Private/JavaScript/SlubMediaPlayer/SlubMediaPlayer.js
sebastian-meyer:
```suggestion
    const timecodeMatch = link.hash.match(/#timecode=(\d+(?:\.\d+)?)(?:;fileIds=([\w,-]+))?/);
```

sebastian-meyer (approved):

PR #1679, opened 2025-08-07
[BUGFIX] Fix `storagePid` assignment for parent document

beatrycze-volk:


fschoelzel:
Thanks for the fix @beatrycze-volk.
With this Bugfix, the metadata of indexed documents is displayed again in the Kitodo Viewer.

sebastian-meyer (approved):

PR #1422, opened 2025-01-02
[FEATURE] Implement `UseGroupsConfiguration` class

beatrycze-volk:

PR #1680, opened 2025-08-08
[BUGFIX] Fix exception indexing documents if some pages miss fulltext

thomaslow:
While working on the Typo3 v13 update, I noticed problems in the current main branch with Typo3 v12 (probably also with Typo3 v11).

Related to #1422, #1683

When indexing documents via the Typo3 `kitodo:index` command, the following error message is reported:

```
Uncaught TYPO3 Exception Kitodo\Dlf\Common\MetsDocument::getFileLocation(): Argument #1 ($id) must be of type string, null given, called in /var/www/html/kitodo-presentation/Classes/Common/MetsDocument.php on line 1182
thrown in file /var/www/html/kitodo-presentation/Classes/Common/MetsDocument.php
in line 266
```

I traced the problem to the  `getFulltext()` method, which checks whether the document contains at least one page with fulltext by calling `$this->ensureHasFulltextIsSet();`. However, if another page does not have fulltext files available, extracting fulltext triggers the exception. The current implementation assumes that every page provides fulltext files.

The pull request now returns an empty string in case a page does not have fulltext available.


markusweigelt:
@thomaslow I encountered the same problem and solved it in a similar way. I didn’t submit a PR because after implementing my quick fix, I saw your draft.

Maybe you could change the log level after the fileLocations check to "warning" or "debug". Cause if there is no fulltext linked to current physical structured node than i think an error message is to heavy. 
For example following valid METS with fulltext files `https://digital.slub-dresden.de/data/kitodo/PflainNa_492784345_0002/PflainNa_492784345_0002_mets.xml`
contains a pointer to `FULLDOWNLOAD` file:

```
<mets:div TYPE="physSequence" ID="PHYS_0000">
    <mets:fptr FILEID="FULLDOWNLOAD"/>
    <mets:div ID="PHYS_0001" TYPE="page" ORDER="1" ORDERLABEL=" - ">
    ...
```
What do you mean?

thomaslow:
@markusweigelt Yes, that makes sense. Some physical divisions may not have fulltext, which would be fine and shouldn't trigger an error message. I lowered the log level to `debug`.

Classes/Common/MetsDocument.php
```diff
@@ -1178,10 +1178,17 @@ public function getFullText(string $id): string
...
                 while ($useGroup = array_shift($useGroups)) {
-                    $fileLocations[$useGroup] = $this->getFileLocation($physicalStructureNode['files'][$useGroup]);
+                    if (in_array($useGroup, array_keys($physicalStructureNode['files']))) {
+                        $fileLocations[$useGroup] = $this->getFileLocation($physicalStructureNode['files'][$useGroup]);
+                    }
                 }
             }
 
+            if (empty($fileLocations)) {
+                $this->logger->debug('No file locations for fulltext @ID "' . $id . '"');
```
```suggestion
                $this->logger->debug('No file locations for page @ID "' . $id . '"');
```

sebastian-meyer (approved):

PR #1281, opened 2024-07-20
[FEATURE] Support for musical sources, multi-level metadata and experimental annotations

sebastian-meyer:
This adds support for musical sources with MEI-encoded scores. They are visualized using [Verovio](https://www.verovio.org/index.xhtml). Furthermore, musical sources can be navigated page-based (like every other document) or alternatively based on measurement information.

In addition, this adds support for multi-level metadata (i. e. metadata grouping) for all documents.

This also adds experimental support for annotations.

PR #1682, opened 2025-08-08
[BUGFIX] Fix wrong viewer Javascript is triggered due to multiview setting

thomaslow:
While working on the Typo3 v13 update, I noticed problems in the current main branch with Typo3 v12 (probably also with Typo3 v11).

Related to #1281 

When opening a document with the Kitodo.Presentation demo theme, no images are shown on the detail page. The problem seems to be that the condition which decides whether the `multiview` mode is used only considers whether mutliple documents are available, but not whether the actual `multiview` setting is enabled.

In my case, the `documentArray` contains multiple documents, which triggers the `multiview`, but there is no support for it in the demo theme.


sebastian-meyer (approved):

fschoelzel:
It is necessary to define the `multiViewType` in the `tx_dlf_pageview` settings in the typoscript of the Portal or within Sitepackages. I added the Definition to the slub_digitalcollections. 
See: https://github.com/slub/slub_digitalcollections/pull/159

@sebastian-meyer @thomaslow 
It might be useful to check for the existing Array Key in the settings upfront. Otherwise we get an Exception. 
`PHP Warning: Undefined array key "multiViewType" in /var/www/html/packages/kitodo-presentation/Classes/Controller/PageViewController.php line 466`

fschoelzel:
Added this Bugfix for the Issue now: https://github.com/kitodo/kitodo-presentation/pull/1695

PR #1689, opened 2025-08-15
[BUGFIX] Document caching

### LINKED ISSUE #1555: [BUG] Excessive downloads for single document (regression)
## Description

With the latest code of Kitodo.Presentation and DFG Viewer, viewing of a single page of a document downloads the required METS file 34 times.

The official DFG Viewer which is based on older code downloads the same METS file only once. That's the desired behaviour.

## Reproduction

Use DFG Viewer with a METS URL on a webserver where the access log can be monitored and count the number of accesses (expected number is 1).

View the same METS URL with an installation based on the latest code and monitor the server accesses again. It will take significantly more time because it requires 34 accesses for the METS URL.

markusweigelt:
Unnecessary implementations for deleting the document cache have been changed. Currently, in the log of the server where the document is stored, there is only one retrieval recorded instead of for example 46 retrievals.

According to the current configuration, the cache is automatically invalidated after 24 hours.

At present, I have only been able to test the behavior partially, so side effects cannot be completely ruled out.

Fixes #1555

It would be good if the force reload were not so easily possible as a parameter, but instead as an explicit function call. Is it generally necessary to also invalidate the cache for the index? The force reloads are still in there at the moment. Documents might possibly be indexed faster if the cache is enabled.


sebastian-meyer (approved):
A forced reload of a document is only ever necessary while re-indexing a document. Then we don't want it to be loaded from database and/or index, but parsed from the (most likely changed) METS file instead.

My tests seem to confirm this fixes the issue, but my test environment is rather simple. So it would be nice to have this tested in a more production-like environment as well... So waiting for reviews from @beatrycze-volk and @stweil.

beatrycze-volk (approved):

sebastian-meyer:
Thank you!

PR #1695, opened 2025-08-20
[BUGFIX] Add check for Array Key `multiViewType` as it is not mandatory in typoscript config

fschoelzel:


sebastian-meyer (approved):

thomaslow:
I thought the variable `$this->settings['multiViewType']` contains a boolean depending on whether the option `multiViewType` in the `PageView`-Plugin is set?

thomaslow:
Nevermind, empty returns false if the value is false. I thought empty only checks whether a value is set.

PR #1706, opened 2025-08-26
[BUGFIX] check for imageUrls where empty nested arrays returned false positives

fschoelzel:
Added dlfUtils.hasContent to ensure initLayer only runs when at least one non-empty value exists, avoiding false positives from empty arrays inside arrays, as their length = 1 and return `TRUE`. For Example in non Image METS/MODS Documents.


sebastian-meyer (approved):

fschoelzel:
TODO:
Avoid init the dlfViewer for Images in Audio/Video Context

PR #1686, opened 2025-08-13
[MAINTENANCE] Typo3 v13 support

thomaslow:
The following pull request drops support for Typo3 v11 and updates Kitodo.Presentation for Typo3 v13.

Related Issues:
- Fixes #1545 
- Fixes #1518 
- Fixes #1442

Includes Pull Requests:
- #1685
- #1682
- #1681
- #1680

Blocked by:
- https://github.com/kitodo/kitodo-presentation/issues/1374 (see https://github.com/ubleipzig/php-iiif-prezi-reader/issues/1)

### Todo

Mandatory
- [x] Make tests run with Typo3 v12 and v13
- [x] Update CI for Typo3 v12 and v13
- [x] Fix deprecation warnings from Typo3 v12 extension scanner

Maybe
- ~~Test with [DFG viewer](https://github.com/slub/dfg-viewer/)~~
- [x] Check whether #966 works now due to Solr core initialization code fix, see #1703
- [x] Check whether #954 can be updated, see #1708

### Changes

- Update composer typo3 dependencies to `^12.4|^13.4`
  - ~~PHP 8.2 required due to `symfony/process>=7.0` required for Typo3 v13.4~~
- Delete file `ext_tables.php` (used in v11) which was replaced by `Configuration/Backend/Modules.php` in v12+
- Update TCA file field definitions according to [Feature #98479](https://docs.typo3.org/c/typo3/cms-core/main/en-us/Changelog/12.0/Feature-98479-NewTCATypeFile.html)
- Fix php 8.4 fatal typing errors
- Replace deprecated `GeneralUtility::_GET()` with `ServerRequestInterface`, see [Deprecation #100596](https://docs.typo3.org/c/typo3/cms-core/main/en-us/Changelog/12.4/Deprecation-100596-GeneralUtility_GET.html)
- Update use of `ModuleTemplate`, see [Feature #96730](https://docs.typo3.org/c/typo3/cms-core/main/en-us/Changelog/12.0/Feature-96730-SimplifiedExtbackendModuleTemplateAPI.html)
- Update Javascript import from RequireJS to ES6, see Typo3 v12 [migration instructions](https://docs.typo3.org/m/typo3/reference-coreapi/12.4/en-us/ApiOverview/Backend/JavaScript/ES6/Index.html#requirejs-migration)
[...]

PR #1711, opened 2025-08-28
[MAINTENANCE] Show more detailed phpunit messages

thomaslow:
This pull request adds phpunit options to show more detailed messages for unit and functional tests. This was quite useful during upgrading to Typo3 v13. However, if there are lots of problems, the output gets fairly long and needs to be scrolled heavily. Not sure what you prefer.

You can see an example when running functional tests with Typo3 v13, where lots of Typo3 v14 deprecation warnings are popping up.

This pull request includes #1686 because the phpunit config file schema changed.


sebastian-meyer:
This is very useful and comes in handy when reviewing pull requests. It also makes debugging a lot easier if some tests fail.

sebastian-meyer (approved):

PR #1719, opened 2025-09-04
[BUGFIX] Fix showing text in disabled fulltext view

### LINKED ISSUE #1408: [BUG] Viewer shows "Volltext wird geladen ..." when page is loaded with disabled fulltext view
## Description

A DFG Viewer test installation with TYPO3 11 and latest code from GitHub for dfgviewer, dlf and slub_digitalcollections shows "Volltext wird geladen ..." although the fulltext view is disabled.

## Reproduction

1. Run local DFG Viewer with recent code.
2. Modify URL to show a digitized newspaper issue which has FULLTEXT.
3. Disable the fulltext display if fulltext is shown.
4. Force reload of the page.
5. See "Volltext wird geladen ...".

## Expected Behavior

If the page is loaded while the fulltext display was disabled by the user, there should be no fulltext display.

## Environment

  - TYPO3 version: 11

markusweigelt:
- Now checks whether the control variable is initially active. If it is active, perform the necessary procedures to enable full-text view. Otherwise, disable the view.

Fixes #1408

### DIFF
--- a/Resources/Public/JavaScript/PageView/FulltextControl.js
+++ b/Resources/Public/JavaScript/PageView/FulltextControl.js
@@ -242,10 +242,8 @@ var dlfViewerFullTextControl = function(map) {
         this)
     };
 
-    if (this.isActive) {
-        this.activate();
-    } else {
-        this.deactivate();
+    if (!this.isActive) {
+      $(this.fullTextScrollElement).hide();
     }
 
     $(this.fullTextScrollElement).text(this.dic['fulltext-loading']);
--- a/Resources/Public/JavaScript/PageView/FulltextControl.js
+++ b/Resources/Public/JavaScript/PageView/FulltextControl.js
@@ -242,7 +242,13 @@ var dlfViewerFullTextControl = function(map) {
         this)
     };
 
-    $('html').find(this.fullTextScrollElement).text(this.dic['fulltext-loading']);
+    if (this.isActive) {
+        this.activate();
+    } else {
+        this.deactivate();
+    }
+
+    $(this.fullTextScrollElement).text(this.dic['fulltext-loading']);
 
     this.changeActiveBehaviour();
 };


stweil (approved):
I [tested](https://digital.ladenburg.world/viewer?tx_dlf[id]=2845) this fix and can confirm that it works fine.

Many thanks!

stweil:
@sebastian-meyer, @thomaslow, I am sorry but I also have just noticed a regression. When fulltext is enabled, the displayed text should follow the mouse cursor when it is moved above the image. This feature no longer works in my installation after the latest upgrade and with this pull request.

stweil (changes requested):
After running more tests, I think that the regression with the text which no longer follows the mouse is caused by this pull request. So it fixes a regression, but also introduces a new one.

markusweigelt:
@stweil Can you send me the document/URL you are testing with? In which environment are you conducting your tests?

In our testing environment, it works in both Chrome and Firefox.
https://3dtest.dfg-viewer.de/viewer?tx_dlf%5Bid%5D=https%3A%2F%2Fdigital.slub-dresden.de%2Fdata%2Fkitodo%2Fjohagovoq_358669413%2Fjohagovoq_358669413_mets.xml&tx_dlf%5Bpage%5D=7

stweil:
Your example does not work in my latest Firefox 142.0.1 on macOS. I load the page, see the image without fulltext, enable fulltext, see the first part of the fulltext, put the mouse pointer over the lower part of the image and see no change in the fulltext display.

stweil:
For [my test page](https://digital.ladenburg.world/viewer?tx_dlf%5Bid%5D=2845) the same procedure works fine: the fulltext display changes when the mouse pointer is moved to some different part of the image. This no longer works as soon as I apply your commit.

stweil:
@markusweigelt, I get the same behaviour with Google Chrome and Safari on macOS, so it does not look like a browser issue. All three browsers show the same regression with your test page.

Note that it's not about the highlighting of the text which is under the mouse pointer. This works fine. The problem is that the fulltext display does not scroll to the text under the mouse pointer if that is required.

markusweigelt:
@stweil You are right, I did not take the scrolling into account. I have provided an adjustment, and it should work now. Thank you for your testing.

sebastian-meyer (approved):

stweil (changes requested):
@markusweigelt, text scrolling still does not work reliably in your example. It does not work at all with your latest commit in my example.

stweil (approved):
It now works in my installation, too. I had to clear the TYPO3 caches more aggressively and deleted temporary files. Obviously this fixed it.

stweil:
Thank you, @markusweigelt, for this fix.

stweil:
@sebastian-meyer, the fix is also required in the main branch.

sebastian-meyer:
Yes, but I am currently on vacation and can't cherry-pick commits in the web client. I'll update the main branch when I am back home (in 3 weeks).
If that's too late, someone has to create a pull request.

markusweigelt:
@sebastian-meyer Created PR #1721 to migrate commits into main branch.

PR #1745, opened 2025-11-04
[MAINTENANCE] Drop support for PHP 8.1

stweil:
PHP 8.1 had active support until 2023-11-25.
Security support will end 2025-12-31.

Therefore new installations should not be based on PHP 8.1.

### DIFF
--- a/.github/workflows/phpstan.yml
+++ b/.github/workflows/phpstan.yml
@@ -12,7 +12,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        variants: [ {typo3: 12.4, php: 8.1}, {typo3: 13.4, php: 8.4}  ]
+        variants: [ {typo3: 12.4, php: 8.2}, {typo3: 13.4, php: 8.4}  ]
     steps:
       - name: Checkout code
         uses: actions/checkout@v4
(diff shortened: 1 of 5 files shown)


stweil:
See https://www.php.net/supported-versions.php.

sebastian-meyer:
While in principal it is the correct move to drop support for unsupported PHP versions, I am not sure if this is the right time to do so. There are still major distributions (like Ubuntu 24.04 or SLES 15) which have PHP 8.1 bundled and are still supported for a couple of years...

stweil:
PHP 8.1 is used in Ubuntu 22.04 LTS which was released in 2022 and is supported until 2027.

But here we are talking about the latest Kitodo code, not an old release. I see no need to support this new code on old LTS releases of Ubuntu. Ubuntu 24.04 LTS comes with PHP 8.3. This is a much better choice for new installations.

stweil:
PS. I was inspired to create this pull request because VuFind has just dropped support for PHP 8.1 in their new release, which is due to be published this month.

sebastian-meyer:
OK, you're right about Ubuntu (I got confused with the different LTS variants). But SLES16 was just release a couple of days ago, so SLES15 will most likely be around for at least a couple more months.

That being said: we are talking about the main branch and the upcoming major release 7.0, so I can see why PHP 8.1 won't be relevant anymore by then.

Thanks for the PR!

sebastian-meyer (approved):

Issue #1757, opened 2025-11-25
[BUG] Several new regressions (some of them only with DFG Viewer)


stweil:
## Description

The viewer no longer works with newspapers in an installation of latest Kitodo.Presentation with latest dfgviewer and slub-digitalcollections.

TYPO3 shows runtime exceptions. In addition, essential functionality is missing in the document viewer.

- [x] Runtime exception because of missing PHP dependency (fix: #1767)
- [x] Runtime exception because of missing type cast (possible fix: #1765)
- [x] Fulltext button and fulltext missing in viewer. Works again in test with code from 2025-12-08.
- [x] Missing dc:calc in dfgviewer, see [slub_digitalcollections](https://github.com/slub/slub_digitalcollections/pull/179). Fixed in https://github.com/slub/dfg-viewer/pull/383.
- [x] Runtime exception because of typed static property, see [slub_digitalcollections](https://github.com/slub/slub_digitalcollections/pull/186). Fixed in https://github.com/slub/slub_digitalcollections/pull/195.

## Reproduction

Install it and try to display a newspaper (years, months, individual issue).

## Expected Behavior

No regression.

## Environment

 - OS version: Debian GNU Linux stable
 - TYPO3 version: 12

## Additional Context

See my comments for the pull requests https://github.com/slub/slub_digitalcollections/pull/179 and https://github.com/slub/slub_digitalcollections/pull/186 which are responsible for two regressions.

One more regressions remains after reverting these two commits:
```
Tue, 25 Nov 2025 14:23:47 +0100 [ALERT] request="806b92604d22b" component="TYPO3.CMS.Frontend.ContentObject.Exception.ProductionExceptionHandler":
Oops, an error occurred! Code: 202511251423476245d421- Exception:
Kitodo\Dlf\ViewHelpers\LinkViewHelper::doubleEncode(): Argument #1 ($url) must be of type string, int given,
called in /viewer/vendor/kitodo/presentation/Classes/ViewHelpers/LinkViewHelper.php on line 75
[...]
```

stweil:
@markusweigelt, one regression is related to pull request #1749. I could fix it with PR #1765, but maybe you have a better fix.

stweil:
@markusweigelt, @michaelkubina, there is another regression which seems to exist since end of November according to the [CI tests](https://github.com/kitodo/kitodo-presentation/actions/workflows/tests.yml): 
```
[...]

PR #1765, opened 2025-12-01
[BUGFIX] Add missing type cast

stweil:
This fixes a regression (issue #1757).

Fixes: f6faaefbe451 ("Replace multiview source remove link (#1749)")


stweil:
@markusweigelt, please review whether this is the right fix for the regression. Other solutions (for example changing the type of the url argument for function doubleEncode) would also be possible.

sebastian-meyer (approved):

markusweigelt (approved):
Great thx!

PR #1767, opened 2025-12-01
[BUGFIX] Add required typo3/cms-reports

stweil:
This fixes lots of CI errors:

    Error: Interface "TYPO3\CMS\Reports\StatusProviderInterface" not found

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -35,6 +35,7 @@
     "typo3/cms-core": "^12.4|^13.4",
     "typo3/cms-extbase": "^12.4|^13.4",
     "typo3/cms-install": "^12.4|^13.4",
+    "typo3/cms-reports": "^12.4|^13.4",
     "typo3/cms-scheduler": "^12.4|^13.4",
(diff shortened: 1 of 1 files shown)


stweil:
Note that I don't know which commit added this new requirement, but it looks like it is not only required for the tests: I also got it in my installation.

Also not that I did not update composer.lock or composer.lock.v12. why was composer.lock.v12 added recently? It is nowhere mentioned or referred to as far as I see.

stweil:
I am not sure what caused the [new CI error](https://github.com/kitodo/kitodo-presentation/actions/runs/19824222096/job/56793463105?pr=1767), but it looks unrelated.

sebastian-meyer:
I think, TYPO3 changed their Composer dependencies and no longer requires `typo3/cms-reports`. That's why it was automatically installed up until now, but isn't anymore. We do depend on it, but haven't declared it an explicit dependency up until now, because it already was a dependency of one of the other `typo3/cms-*` packages.

Thanks for the fix!

sebastian-meyer (approved):

PR #1768, opened 2025-12-01
[OPTIMIZATION] Decrease pagegrid loading time and remove pagegrid cache

michaelkubina:
This PR decreases the initial loading time of the pagegrid in particular, removes the pagegrid cache and increases performance on retrieving file informations in general.

* moves extraction of `location` and `mimetype` for a file into the `magicgetFileGrps()` function. The overhead of extracting these informations from `FLocat` here as well is negligible, as we must already traversing all file nodes once - so we might simply peek down into `FLocat` for each file and quickly grab them right here.
* that makes the xpath expressions in `getFileLocation` and `getFileMimeType` obsolete, as we can get those informations from the `fileInfos` array now. This significantly reduces operations, as we dont need to parse the mets file over and over again in order to retrieve a specific file-information from the file sections. This is what made the pagegrid actually that slow...applying hundreds of xpath expressions on the mets file (one for each `@ID`) in order to extract file locations in a `for-loop`. With an increasing amount of filegroups this slowed down operations even more.
* we now no longer need the `getFileLocationByUsegroup` function, which was exclusivly used in the pagegrid controller, to reduce the lookup time.
* we now no longer need the `pagegrid` cache, as the performance is close to what we had with it in place. But with the benefit, that even the initial loading time is now as fast as if it had been retrieved from the cache.

I have made some performance measurements on a large and in depth structured "Hamburger Adressbuch" from 1960 - around 4600 pages, 4000 structure elements, 7 file groups:
https://mets.sub.uni-hamburg.de/kitodo/PPN773134158_1960

OLD pagegrid:
Initial loading time (without the cache in place) - around 11.000ms to 14.000ms
Browsing to next page (retrieving from cache) - around 600ms - 700ms

NEW pagegrid:
Initial loading time (no cache at all) - around 650ms - 750ms
Browsing to next page (no cache at all) - around 650ms - 750ms

Though browsing is slighlty slower on such a big document, we no longer need to maintain a seperate cache. Also each invocation of `getFileLocation/getFileMimeType` elsewhere benefits from reduces retrival time, as we do not need to xpath through the documents multiple times anymore as this has been done exactly once in `magicgetFileGrps()`.


Classes/Controller/PageGridController.php
```diff
@@ -87,28 +78,26 @@ public function mainAction(): ResponseInterface
...
      *
      * @return array The rendered entry ready for fluid
      */
-    protected function getEntry(int $number): array
+    protected function getEntry(int $number, array $fileGrpThumbs): array
     {
+        $phys = $this->document->getCurrentDocument()->physicalStructure[$number];
         $entry = [];
-
+
```
sebastian-meyer:
```suggestion

```

sebastian-meyer (changes requested):

sebastian-meyer (changes requested):
This is great! But in order to merge this, the tests need to be adjusted accordingly.

michaelkubina:
So, there where some issues, because prior to my changes the `fileInfos` were lazy-loaded...its now fixed by loading the `fileInfos` via `magicGetFileGrps` once if they have not been loaded already. Also i had a typo in the array key for the `mimeType`, and missing emptiness checks. Finally i made sure all variables are in scope.

The tests pass now and did not require a change, as the issue was elsewhere. But i dont understand the errors with the static analyzer...i dont think this is my fault..or is it?

sebastian-meyer:
> So, there where some issues, because prior to my changes the `fileInfos` were lazy-loaded...its now fixed by loading the `fileInfos` via `magicGetFileGrps` once if they have not been loaded already. Also i had a typo in the array key for the `mimeType`, and missing emptiness checks. Finally i made sure all variables are in scope.

Thank you very much!

> The tests pass now and did not require a change, as the issue was elsewhere. But i dont understand the errors with the static analyzer...i dont think this is my fault..or is it?

That's fine! It was a temporary problem with GitHub.

sebastian-meyer (approved):

PR #1774, opened 2025-12-11
[FEATURE] Introduce DocumentService for one-time loading of current document

marco-lengua:
This pull-request introduces a Document Service for document loading, so that the initialized document object persists across controllers. It adresses the current state issue, that every plugin controller (that implements AbstractController) loads the current document, initializes it and once the controller finishes it is deconstructed.
Every consecutive controller needs to load and initialize the document again.
This solution reuses the loaded document for the whole request and significantly optimizes execution time:
- 5000 Page Adressbuch loads one-time for about 0.5 Seconds in the first controller and then loading is nearly instantly.
- without this change each controller takes 0.5 seconds (chaining for example 6 controllers -e.g. detail view with PageGrid, PageView, ToC, Navigation, Toolbox, Metadata - leads to a loading time of about 3 seconds).
- in this case the change improves execution time by a factor of 6.


marco-lengua:
We doublecheck an issue with the dependency injection

PR #1775, opened 2025-12-11
[BUGFIX] Fix translation issues in new tenant controller

### LINKED ISSUE #1497: [BUG] Initialize new tenant with German and English Metadata and Structures
### Discussed in https://github.com/kitodo/kitodo-presentation/discussions/1471

<div type='discussions-op-text'>

<sup>Originally posted by **stweil** February  5, 2025</sup>
It's possible to create _Metadata_ and _Structures_ for a new tenant automatically in TYPO3. Ideally this should create 30 bilingual entries for _Metadata_ and 92 bilingual entries for _Structures_ like in this example.

Bildschirmfoto 2025-02-05 um 11 00 51

It seems to be difficult to set up the site correctly to get this result. Often there will only be entries which are all either in English or in German.

Instructions for the recommended site setup which produces the desired result would be helpful.</div>

michaelkubina:
This PR fixes an issue, where translations were not set properly when initializing new tenants metadata and structures.

In short: the prefixes of our language files are determined by the language code, whereas site languages have presets of languageCode and countryCode. This is the standard typo3 way for most of the languages/locales. The full discussion can be found here: #1471

Instead of relying on the sitelanguage for prefix determination, we use the locales languagecode now. This way sites that have used site language prefixes like `en_US` or `de_DE` will use the correct english or german translations. Site, that have provides just a language code as the sitelanguage (e.g. only `en` or `de`) will also get correct english and german translations, as the language code will be extracted the same way.

Custom country specific site-language like `de_CH` will still not be handled, but will at least use the standard `de` translation. This should be adressed when we encounter actual enduser issues regarding country specific translations for the same language - which i believe is not the case right now.

### DIFF
--- a/Classes/Controller/Backend/NewTenantController.php
+++ b/Classes/Controller/Backend/NewTenantController.php
@@ -244,7 +244,7 @@ class NewTenantController extends AbstractController
         $metadataDefaults = $this->getRecords('Metadata');
 
         // load language file in own array
-        $metadataLabels = $this->languageFactory->getParsedData('EXT:dlf/Resources/Private/Language/locallang_metadata.xlf', $this->siteLanguages[0]->getTypo3Language());
+        $metadataLabels = $this->languageFactory->getParsedData('EXT:dlf/Resources/Private/Language/locallang_metadata.xlf', $this->siteLanguages[0]->getLocale()->getLanguageCode());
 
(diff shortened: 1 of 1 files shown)


michaelkubina:
Fixes: #1497

sebastian-meyer (approved):

PR #1776, opened 2025-12-12
[FEATURE] Introduce DocumentService for one-time loading of current document

marco-lengua:
This pull-request introduces a Document Service for document loading, so that the initialized document object persists across controllers. It adresses the current state issue, that every plugin controller (that implements AbstractController) loads the current document, initializes it and once the controller finishes it is deconstructed.
Every consecutive controller needs to load and initialize the document again.
This solution reuses the loaded document for the whole request and significantly optimizes execution time:

- 5000 Page Adressbuch loads one-time for about 0.5 Seconds in the first controller and then loading is nearly instantly.
- without this change each controller takes 0.5 seconds (chaining for example 6 controllers -e.g. detail view with PageGrid,PageView, ToC, Navigation, Toolbox, Metadata - leads to a loading time of about 3 seconds).
- in this case the change improves execution time by a factor of 6.
 
Eventually it is required to flush all cashes in Admin->Maintainance so the dependency injection cache is cleared.


Classes/Controller/AbstractController.php
```diff
@@ -54,12 +56,15 @@ abstract class AbstractController extends ActionController implements LoggerAwar
...
+     * @access protected
+     * @var DocumentService
      */
+    protected DocumentService $documentService;
+ 
+    public function injectDocumentService(DocumentService $documentService): void
+    {
+        $this->documentService = $documentService;
+    }
     public function injectDocumentRepository(DocumentRepository $documentRepository): void
```
sebastian-meyer: Please restore the phpDoc for this method.

Classes/Service/DocumentService.php
```diff
@@ -0,0 +1,178 @@
...
+ */
+
+namespace Kitodo\Dlf\Service;
+
+use Kitodo\Dlf\Domain\Repository\DocumentRepository;
+use TYPO3\CMS\Core\Log\LogManager;
+use TYPO3\CMS\Core\Utility\GeneralUtility;
+use Kitodo\Dlf\Common\AbstractDocument;
+use Kitodo\Dlf\Domain\Model\Document;
+use TYPO3\CMS\Core\Utility\MathUtility;
```
sebastian-meyer: We typically sort imported namespaces alphabetically for better readability.
```suggestion
use Kitodo\Dlf\Common\AbstractDocument;
use Kitodo\Dlf\Domain\Model\Document;
use Kitodo\Dlf\Domain\Repository\DocumentRepository;
use TYPO3\CMS\Core\Log\LogManager;
use TYPO3\CMS\Core\Utility\GeneralUtility;
use TYPO3\CMS\Core\Utility\MathUtility;
```

Classes/Service/DocumentService.php
```diff
@@ -0,0 +1,178 @@
...
+     * @access protected
+     * @var Document|null This holds the current document
+     */
+    protected ?Document $document = null;
+    /**
+     * @access protected
+     * @var array
+     */
+    protected $settings;
+    /**
```
sebastian-meyer:
```suggestion

    /**
```

Classes/Service/DocumentService.php
```diff
@@ -0,0 +1,178 @@
...
+     * @access protected
+     * @var array
+     */
+    protected $logger;
+    protected DocumentRepository $documentRepository;
```
sebastian-meyer:
```suggestion

    /**
     * @access protected
     * @var DocumentRepository
     */
    protected DocumentRepository $documentRepository;
```

Classes/Service/DocumentService.php
```diff
@@ -0,0 +1,178 @@
...
+    }
+
+    protected function init()
+    {
+        return true;
+    }
+    protected function reset()
+    {
+    }
+    /**
```
sebastian-meyer:
```suggestion

    /**
```

Classes/Service/DocumentService.php
```diff
@@ -0,0 +1,178 @@
...
+     * @param int $recordId
+     * @param array $settings
+     */
+    private function serviceLoadDocument($recordId, $settings)
+    {
+        $this->settings = $settings;
+        // Get document ID from request data if not passed as parameter.
+        if (!empty($recordId)) {
+            $documentId = $recordId;
+        }
```
sebastian-meyer, line 97: This does not what the comment suggests. In fact, if the document ID is not passed as parameter, document loading will fail...
Please have a closer look at the original implementation!

Classes/Service/DocumentService.php
```diff
@@ -0,0 +1,178 @@
...
+     * @access protected
+     * @var array
+     */
+    protected $logger;
+    protected DocumentRepository $documentRepository;
+
+    public function __construct()
+    {
+        $this->documentRepository = GeneralUtility::makeInstance(DocumentRepository::class);
+        $this->logger = GeneralUtility::makeInstance(LogManager::class)->getLogger(__CLASS__);
```
sebastian-meyer: Please consider using [dependency injection](https://docs.typo3.org/m/typo3/reference-coreapi/12.4/en-us/ApiOverview/Logging/Logger/Index.html) (recommended) or the [LoggerAwareTrait](https://docs.typo3.org/m/typo3/reference-coreapi/10.4/en-us/ApiOverview/Logging/Logger/Index.html).

Classes/Service/DocumentService.php
```diff
@@ -0,0 +1,178 @@
...
+                $doc->cPid = max($this->settings['storagePid'], 0);
+            }
+
+            $this->document->setLocation($documentId);
+        } else {
+            $this->logger->error('Invalid location given "' . $documentId . '" for document loading');
+        }
+
+        return $doc;
+    }
```
sebastian-meyer, line 190: Those are duplicates of the same methods in `AbstractController`. Please try to avoid duplicate code.

Classes/Controller/AbstractController.php
```diff
@@ -244,55 +249,15 @@ protected function buildMultiViewDocuments(string $docUrl, AbstractDocument $doc
...
+     * Load the current Document to memory just once with DocumentService - it will then be available for all controllers.
      *
      * @access protected
-     *
-     * @param string $documentId The document's UID or URL (id), fallback: record ID (recordId)
-     *
      * @return void
      */
-    protected function loadDocument(string $documentId = ''): void
+    protected function loadDocument(): void
```
sebastian-meyer: There are still instances where `loadDocument()` is called with a parameter (e. g. in `BasketController`). Please make sure this doesn't break.

sebastian-meyer (changes requested):
Thank you for your contribution and welcome to Kitodo! ;o)

Please see my comments below.

In general, please add (or restore) phpDoc where necessary according to the [Coding Guidelines](https://github.com/kitodo/coding-guidelines).

Also, maybe I am missing something, but wouldn't the `DocumentService` have to be a singleton in order for every controller to query the same instance? And how does this work when instantiating multiple documents (like for the `BasketController` or the new MultiView feature?

Classes/Service/DocumentService.php
marco-lengua: done, thanks for the hint

Classes/Service/DocumentService.php
marco-lengua: that is actually my fault, fixed it

Classes/Service/DocumentService.php
marco-lengua: used loggeraware trait for logging and dependency injection for document repository, fixed

Classes/Service/DocumentService.php
marco-lengua: yes true - these two methods are needed here in the Document Service to load the proper documents. Removed one in the abstract Controller that is not needed anymore. The other one is called one time by a controller.

Classes/Controller/AbstractController.php
marco-lengua: thanks for the hint, fixed it

marco-lengua:
The Service Type is recommended since Typo3 Version 12 and Singletons are considered obsolete this is why we chose this pattern.
[Singletons Typo3 V12](https://docs.typo3.org/m/typo3/reference-coreapi/12.4/en-us/CodingGuidelines/CodingBestPractices/Singletons.html#singletons).
In practice for one request cycle they behave the same as singletons. One instance persists for one request lifecycle across all controllers. If we would like to further extend the loading of a document to be available over more than one request we recommend using the cache, but that would be another pull request. 
The Service does not change the controllers loading behaviour as it is now. The Basket Controller or even multiview still work the same.

Configuration/Services.yaml
```diff
@@ -53,3 +53,6 @@ services:
 
   Kitodo\Dlf\Validation\XmlSchemesValidator:
     autowire: false
+
+  Kitodo\Dlf\Service\DocumentService:
+    public: true
```
sebastian-meyer:
```suggestion
    public: true

```

Classes/Service/DocumentService.php
```diff
@@ -0,0 +1,177 @@
...
+     * @return ?Document
+     */
+    public function getDocument(string $documentId, int $recordId, array $settings)
+    {
+        if ($this->document === null) {
+            $this->serviceLoadDocument($documentId, $recordId, $settings);
+        }
+        return $this->document;
+    }
+    /**
```
sebastian-meyer, line 83:
```suggestion
    }

    /**
```

Classes/Service/DocumentService.php
```diff
@@ -0,0 +1,177 @@
...
+    }
+
+    protected function init()
+    {
+        return true;
+    }
+    protected function reset()
+    {
+    }
+    /**
```
sebastian-meyer, line 65:
```suggestion
    }

    /**
```

Classes/Service/DocumentService.php
```diff
@@ -0,0 +1,177 @@
...
+    public function injectDocumentRepository(DocumentRepository $documentRepository): void
+    {
+        $this->documentRepository = $documentRepository;
+    }
+
+    protected function init()
+    {
+        return true;
+    }
+    protected function reset()
```
sebastian-meyer, line 61:
```suggestion
    }

    protected function reset()
```

Classes/Service/DocumentService.php
```diff
@@ -0,0 +1,177 @@
...
+     * @access protected
+     * @var array
+     */
+    protected $settings;
+    /**
+     * @access protected
+     * @var DocumentRepository
+     */
+    protected DocumentRepository $documentRepository;
+    public function injectDocumentRepository(DocumentRepository $documentRepository): void
```
sebastian-meyer, line 51:
```suggestion
    protected DocumentRepository $documentRepository;

    public function injectDocumentRepository(DocumentRepository $documentRepository): void
```

Classes/Service/DocumentService.php
```diff
@@ -0,0 +1,177 @@
...
+     * @access protected
+     * @var Document|null This holds the current document
+     */
+    protected ?Document $document = null;
+    /**
+     * @access protected
+     * @var array
+     */
+    protected $settings;
+    /**
```
sebastian-meyer, line 45:
```suggestion
    protected $settings;

    /**
```

Classes/Service/DocumentService.php
```diff
@@ -0,0 +1,177 @@
...
+ */
+class DocumentService
+{
+    use LoggerAwareTrait;
+    /**
+     * @access protected
+     * @var Document|null This holds the current document
+     */
+    protected ?Document $document = null;
+    /**
```
sebastian-meyer, line 39:
```suggestion
    protected ?Document $document = null;

    /**
```

Classes/Service/DocumentService.php
```diff
@@ -0,0 +1,177 @@
...
+ *
+ * @package TYPO3
+ * @subpackage dlf
+ *
+ * @access public
+ */
+class DocumentService
+{
+    use LoggerAwareTrait;
+    /**
```
sebastian-meyer, line 33:
```suggestion
    use LoggerAwareTrait;

    /**
```

Classes/Service/DocumentService.php
```diff
@@ -0,0 +1,177 @@
...
+        }
+        return $this->document;
+    }
+    /**
+     * @access public
+     * @param string $documentId
+     * @param int $recordId
+     * @param array $settings
+     */
+    private function serviceLoadDocument(string $documentId, int $recordId, array $settings)
```
sebastian-meyer, line 91: The second parameter here is called `recordId`, but it's in fact the document ID from request data.

I suggest renaming the parameter in order to prevent confusion with the "real" record ID (which is used e. g. in line 103).

Classes/Controller/AbstractController.php
```diff
@@ -246,55 +251,16 @@ protected function buildMultiViewDocuments(string $docUrl, AbstractDocument $doc
...
 
     /**
-     * Loads the current document into $this->document
+     * Load the current Document to memory just once with DocumentService - it will then be available for all controllers.
      *
      * @access protected
-     *
-     * @param string $documentId The document's UID or URL (id), fallback: record ID (recordId)
-     *
+     * @param string $documentId
```
sebastian-meyer: Why did you remove the full documentation of this parameter?

Classes/Controller/AbstractController.php
```diff
@@ -56,12 +58,15 @@ abstract class AbstractController extends ActionController implements LoggerAwar
...
-     * @return void
+     * @access protected
+     * @var DocumentService
      */
+    protected DocumentService $documentService;
+ 
+    public function injectDocumentService(DocumentService $documentService): void
+    {
+        $this->documentService = $documentService;
+    }
```
sebastian-meyer, line 68:
```suggestion
    }

```

Classes/Service/DocumentService.php
```diff
@@ -0,0 +1,177 @@
...
+        return true;
+    }
+    protected function reset()
+    {
+    }
+    /**
+     * @access public
+     * @param int $recordId
+     * @param array $settings
+     * @return ?Document
```
sebastian-meyer, line 73: Please complete PHPDoc with short description of parameters and method.

Classes/Service/DocumentService.php
```diff
@@ -0,0 +1,177 @@
...
+        if ($this->document === null) {
+            $this->serviceLoadDocument($documentId, $recordId, $settings);
+        }
+        return $this->document;
+    }
+    /**
+     * @access public
+     * @param string $documentId
+     * @param int $recordId
+     * @param array $settings
```
sebastian-meyer: Please complete PHPDoc with short description of parameters and method.

Classes/Service/DocumentService.php
```diff
@@ -0,0 +1,177 @@
...
+                    $this->document->setCurrentDocument($doc);
+                } else {
+                    $this->logger->error('Failed to load document with record ID "' . $this->requestData['recordId'] . '"');
+                }
+            }
+        } else {
+            $this->logger->error('Invalid ID "' . $documentId . '" or PID "' . $this->settings['storagePid'] . '" for document loading');
+        }
+    }
+    /**
```
sebastian-meyer, line 131:
```suggestion
    }

    /**
```

sebastian-meyer (changes requested):

Classes/Service/DocumentService.php
```diff
@@ -0,0 +1,188 @@
...
+     * @access protected
+     * @var array
+     */
+    protected $settings;
+    /**
+     * @access protected
+     * @var DocumentRepository
+     */
+    protected DocumentRepository $documentRepository;
+
```
sebastian-meyer:
```suggestion

```

Classes/Service/DocumentService.php
```diff
@@ -0,0 +1,188 @@
...
+        return true;
+    }
+
+    protected function reset()
+    {
+    }
+
+    /**
+     * Method that is used by controllers. Check if Document is already loaded or not. If not, loads document
+     *
```
sebastian-meyer:
```suggestion
     *
```

Classes/Service/DocumentService.php
```diff
@@ -0,0 +1,188 @@
...
+    }
+
+    protected function reset()
+    {
+    }
+
+    /**
+     * Method that is used by controllers. Check if Document is already loaded or not. If not, loads document
+     * 
+     * @access public
```
sebastian-meyer:
```suggestion
     * @access public
```

Classes/Service/DocumentService.php
```diff
@@ -0,0 +1,188 @@
...
+
+    protected function reset()
+    {
+    }
+
+    /**
+     * Method that is used by controllers. Check if Document is already loaded or not. If not, loads document
+     * 
+     * @access public 
+     * @param string $documentId $documentId The document's UID or URL (id),
```
sebastian-meyer:
```suggestion
     * @param string $documentId $documentId The document's UID or URL (id),
```

Configuration/Services.yaml
```diff
@@ -53,3 +53,7 @@ services:
...
+
```
sebastian-meyer:
```suggestion

```

Classes/Controller/AbstractController.php
```diff
@@ -35,6 +36,7 @@
 use TYPO3\CMS\Extbase\Mvc\Request;
 use TYPO3\CMS\Extbase\Mvc\RequestInterface;
 
+
```
```suggestion
```

Classes/Service/DocumentService.php
```diff
@@ -0,0 +1,191 @@
...
+
+    protected function reset()
+    {
+    }
+
+    /**
+     * Method that is used by controllers. Check if Document is already loaded or not. If not, loads document
+     *
+     * @access public
+     * @param string $documentId $documentId The document's UID or URL (id),
```
sebastian-meyer:
```suggestion
     * @param string $documentId $documentId The document's UID or URL (id)
```

Classes/Service/DocumentService.php
```diff
@@ -0,0 +1,191 @@
...
+     * 
+     * @return ?Document
+     */
+    public function getDocument(string $documentId, int $recordId, array $settings)
+    {
+        if ($this->document === null) {
+            $this->serviceLoadDocument($documentId, $recordId, $settings);
+        }
+        return $this->document;
+    }
```
sebastian-meyer, line 81: Unfortunately, there are still some issues with this: In Kitodo.Presentation, documents can be referenced by ID (i.e. the UID in database) or by URL (i.e. a link to the METS file). The `recordId`, on the other hand, actually is an unique identifier for METS files. The difference between METS identifier and document identifiers is not obvious, but it is important here.
As you can see in the implementation below, we primarily use UID/URL for retrieving documents, which we either already have (`$documentId`) or fetch from request data (`$recordId`). Only if both are not available, we fallback to using the actual record ID if applicable (line 114f.).

Thus, the variable `$recordId` has a confusing name because it's not actually the record ID, but just the document UID/URL retrieved from request data (see `AbstractController` line 263). You should rename it.

Also, since both `$documentId` and `$recordId` can be either an UID or an URL, they need to be of type `string` (or `int|string`).

(This also applies to `serviceLoadDocument()` below.)

Classes/Service/DocumentService.php
sebastian-meyer: See new comment above.

Classes/Service/DocumentService.php
```diff
@@ -0,0 +1,191 @@
...
+                $doc->cPid = max($this->settings['storagePid'], 0);
+            }
+
+            $this->document->setLocation($documentId);
+        } else {
+            $this->logger->error('Invalid location given "' . $documentId . '" for document loading');
+        }
+
+        return $doc;
+    }
```
sebastian-meyer, line 190: There already is an implementation of `getDocumentByUrl()` in `AbstractController`. Also, both implementations differ slightly.
Please don't duplicate code and make sure you are using the correct implementation!

Classes/Service/DocumentService.php
```diff
@@ -0,0 +1,191 @@
...
+        $this->document = $this->documentRepository->findOneByIdAndSettings($documentId);
+
+        if ($this->document) {
+            $doc = AbstractDocument::getInstance($this->document->getLocation(), $this->settings, false);
+        } else {
+            $this->logger->error('Invalid UID "' . $documentId . '" or PID "' . $this->settings['storagePid'] . '" for document loading');
+        }
+
+        return $doc;
+    }
```
sebastian-meyer, line 152: The (removed) implementation of `getDocumentById()` from `AbstractController` differs slightly from this one. Please make sure to use the latest implementation here!

sebastian-meyer (changes requested):
Unfortunately there are still some issues. Please have a look at my comments below!

marco-lengua:
Hello Sebastian, sorry for any misunderstanding from my side. I know see why this approach won't work with multiview as it is part of the helper method (e.g. getdocument from UID / URL). I close this PR as it becomes obsolete with then new PR i have just created. The new method is much more straightforward and easier to implement. I guess that is what you meant with Singleton.

PR #1777, opened 2025-12-12
[BUGFIX] Fix missing viewdata in kitodo link in NavigationController

michaelkubina:
Fixes an exception from the `NavigationController` template due to a missing `viewData` attribute in the `kitodo:link` for the option `stepBack`.


sebastian-meyer (approved):
Nice catch!

PR #1779, opened 2025-12-16
[BUGFIX] Change return to empty string

beatrycze-volk:
This function can't return null, return needs to be changed to string or function signature needs to be adjusted


Classes/Common/MetsDocument.php
```diff
@@ -237,7 +237,7 @@ public function getDownloadLocation(string $id): string
             // TODO CVT is an optional IIP server capability; in theory, capabilities should be determined in the object request with '&obj=IIP-server'
             return $baseURL . '&CVT=jpeg';
         }
-        return $file['location'] ?? null;
+        return $file['location'] ?? "";
```
```suggestion
        return $file['location'] ?? '';
```

sebastian-meyer (approved):

PR #1780, opened 2025-12-17
[BUGFIX] Restore changes removed by musical sources

beatrycze-volk:


beatrycze-volk:
PHPStan is failing with this error:
`Error: Child process error (exit code 255): Fatal error: Unparenthesized `a ? b : c ? d : e` is not supported. Use either `(a ? b : c) ? d : e` or `a ? b : (c ? d : e)` in /app/vendor/ubl/php-iiif-prezi-reader/src/Iiif/Presentation/V2/Model/Resources/AbstractIiifResource2.php on line 495`

sebastian-meyer:
Yes, that's a common problem with 6.x branch. I haven't found a solution apart from patching `php-iiif-prezi-reader`. :o(

PR #1787, opened 2025-12-23
[IMPROVEMENT] Score rendering

markusweigelt:
- Switch Verovio to the `wasm` implementation → performance issue resolved (together with Daniel Röwenstrunk)
- Move the measure-building function call to after adding the score → timeout issues that occurred with large scores are now properly handled
- Improve height and width options to visualize the whole score in both portrait and landscape modes
- Adjust PDF rendering and take page orientation into account
- Fix a null pointer exception in file selection


sebastian-meyer (dismissed):
Thank you very much for this last minute present! And merry christmas to you!

markusweigelt:
Thx I wish you the same and all the best for the New Year, see you soon! 🐱

sebastian-meyer (approved):

PR #1796, opened 2026-01-06
[FEATURE] Provide structure paths/breadcrumbs to search results

### LINKED ISSUE #1851: [BUG] Searches fail unless the complete Solr index is rebuilt
https://github.com/kitodo/kitodo-presentation/pull/1796/changes/BASE..f539e6164165aa3929656718d50de97dca1fafc3#diff-fdf63d957c2e6980320971efb66ff8015351483bbd5021e474e97226689107c7R259

Searches now fail with an Oops unless the complete Solr index is rebuilt. I fixed it locally with `self::$fields['structure_path'] = $solrFields['structurePath'] ?? '';`.

_Originally posted by @stweil in https://github.com/kitodo/kitodo-presentation/pull/1796#discussion_r2842806730_

michaelkubina:
This PR adds a new feature, that (if activated in settings) displays the structure path to a search result (comparable to a breadcrumb) in the list view.

Current situation:
In the search result sets the documents have either...
* their distinct title, that is extracted from the `dmdsec`-Sections or the `@LABEL` from the ToC, and a page number => in the case for metadata search
* or only the page number => in the case for fulltext search

In both cases users cannot know, where within the logical structure of the document, the search result actually appears. Meaning, that a title like "Krankheiten" could both be in "Kapitel 1: Tiere" or "Kapitel 2: Pflanzen" and we would not know before hand. The same logic applies to fulltext results, where we know nothing of this at all and are completly lost.

Solution:
This PR adds a setting to the `ListView Plugin` that allows for displaying this information. For this to work:
* a new solr field `structure_path` was added, that holds a JSON encoded representation of the structure path/breadcrumb, which is required, so that the `type` (= structure without a custom `label`) can still be translated after retrieval to the correct language. The field is multivalued, because physical pages can be associated with multiple logical structures (= end of "Kapitel 1" and beginning of "Kapitel 2".
* it is required to update the solr `schema.xml`, so that the field is actually defined and used in your instance. Otherwise indexing will fail.

This PR provides the necessary changes to the schema.xml, translation files, listview flexform, the indexer and solrsearch/result as well as to the listview controller and corresponding templates.

Examples:
## Volltextsuche nach "Fischer"
grafik

## Metadatensuche nach "Optik"
grafik

## Volltextsuche nach "Schmied"
grafik

### DIFF
--- a/Classes/Common/Solr/Solr.php
+++ b/Classes/Common/Solr/Solr.php
@@ -256,6 +256,7 @@ class Solr implements LoggerAwareInterface
             self::$fields['type'] = $solrFields['type'];
             self::$fields['title'] = $solrFields['title'];
             self::$fields['volume'] = $solrFields['volume'];
+            self::$fields['structure_path'] = $solrFields['structure_path'];
             self::$fields['date'] = $solrFields['date'] ?? null;
             self::$fields['thumbnail'] = $solrFields['thumbnail'];
             self::$fields['default'] = $solrFields['default'];
--- a/Classes/Common/Solr/Solr.php
+++ b/Classes/Common/Solr/Solr.php
@@ -256,7 +256,7 @@ class Solr implements LoggerAwareInterface
             self::$fields['type'] = $solrFields['type'];
             self::$fields['title'] = $solrFields['title'];
             self::$fields['volume'] = $solrFields['volume'];
-            self::$fields['structure_path'] = $solrFields['structure_path'];
+            self::$fields['structure_path'] = $solrFields['structurePath'];
             self::$fields['date'] = $solrFields['date'] ?? null;
             self::$fields['thumbnail'] = $solrFields['thumbnail'];
             self::$fields['default'] = $solrFields['default'];
--- a/Classes/Common/Indexer.php
+++ b/Classes/Common/Indexer.php
@@ -83,9 +83,9 @@ class Indexer
     /**
      * @access protected
      * @static
-     * @var array List of already processed structure paths
+     * @var array List of already extracted structure nodes for structure path
      */
-    protected static array $processedStructurePaths = [];
+    protected static array $extractedStructurePathNodes = [];
 
     /**
      * @access protected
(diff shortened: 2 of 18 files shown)


sebastian-meyer:
Please fix the PHPStan issue and adjust the failing test. When all checks succeed I am happy to review this!

michaelkubina:
okay, i have found out why the test failed. now everything passed.

sebastian-meyer (changes requested):
Perfect! Thank you very much for this great feature!

Could you please extend documentation for this feature and also include all necessary migration steps when upgrading?

michaelkubina:
I have added the new plugin property to the documentation and i have also updated the administrators instructions for the upgrade path from version 5.1/6.0 to 7.0 ... kind of prematurely for the upcoming release.

To be honest i am not quite sure, if its required in the backend to use "maintenance -> flush cache" in order for the updated extension configuration to be loaded? and if there is a different path or if manual intervention is even required to register the new field in the configuration in a composer based installation (settings.php) or a legacy installation (LocalConfiguration.php). If there are any steps required, it would then luckily only require one additional sentence in the documentation.

Maybe you know the answer out of your pocket?

sebastian-meyer (approved):

sebastian-meyer:
Thank you very much!

> To be honest i am not quite sure, if its required in the backend to use "maintenance -> flush cache" in order for the updated extension configuration to be loaded? and if there is a different path or if manual intervention is even required to register the new field in the configuration in a composer based installation (settings.php) or a legacy installation (LocalConfiguration.php). If there are any steps required, it would then luckily only require one additional sentence in the documentation.

AFAIK this is perfectly fine!

fschoelzel:
@michaelkubina 
A "maintenance -> flush cache" was in my case enough. The field `'structurePath'` got automatically added to the `settings.php`. :+1:

Classes/Common/Solr/Solr.php
```diff
@@ -256,6 +256,7 @@ public static function getFields(): array
             self::$fields['type'] = $solrFields['type'];
             self::$fields['title'] = $solrFields['title'];
             self::$fields['volume'] = $solrFields['volume'];
+            self::$fields['structure_path'] = $solrFields['structurePath'];
```
stweil, line 259: Searches now fail with an Oops unless the complete Solr index is rebuilt. I fixed it locally with `self::$fields['structure_path'] = $solrFields['structurePath'] ?? '';`.

sebastian-meyer: Thanks for the report. I opened a new issue: #1851

michaelkubina: Dear Stefan,
do you mind checking if the field is actually registered in your LocalConfiguration.php or settings.php respectivly? I believe it isnt. Alternativly you can inspect the "Extension-Configuration" - it should be visible in the "solr"-Tab. Or you manually add `'structurePath' => 'structure_path',` to dlf.solr.fields.

grafik



When there is no such field registered, it should appear after "Maintenence" -> "Flush Cache", as i tried to describe in the migration path to 7.x . Florian did it this way and his configuration was updated. 

Searches fail, because in this case it tries to access a field, that has not been applied to your configuration yet. This would have happened to any other field as well, when its not in the configuration.

PR #1813, opened 2026-01-22
[BUGFIX] Adjust TEI HDR metadata for retrieving title

beatrycze-volk:
https://github.com/slub/dfg-viewer/issues/70
https://www.tei-c.org/release/doc/tei-p5-doc/en/html/ref-titleStmt.html
https://www.tei-c.org/release/doc/tei-p5-doc/en/html/ref-sourceDesc.html

Element `titleStmt` looks more correct here than `sourceDesc`.

### DIFF
--- a/Resources/Private/Data/MetadataDefaults.json
+++ b/Resources/Private/Data/MetadataDefaults.json
@@ -28,7 +28,7 @@
             },
             {
                 "format_root": "teiHeader",
-                "xpath": "./teihdr:fileDesc/teihdr:sourceDesc/teihdr:msDesc/teihdr:head/teihdr:note[@type=\"caption\"]",
+                "xpath": "./teihdr:fileDesc/teihdr:titleStmt/teihdr:title",
                 "xpath_sorting": ""
             },
             {
--- a/Resources/Private/Data/MetadataDefaults.json
+++ b/Resources/Private/Data/MetadataDefaults.json
@@ -28,7 +28,7 @@
             },
             {
                 "format_root": "teiHeader",
-                "xpath": "./teihdr:fileDesc/teihdr:titleStmt/teihdr:title",
+                "xpath": "./teihdr:fileDesc/teihdr:titleStmt/teihdr:title | ./teihdr:fileDesc/teihdr:sourceDesc/teihdr:msDesc/teihdr:head/teihdr:title",
                 "xpath_sorting": ""
             },
             {


Resources/Private/Data/MetadataDefaults.json
```diff
@@ -28,7 +28,7 @@
             },
             {
                 "format_root": "teiHeader",
-                "xpath": "./teihdr:fileDesc/teihdr:sourceDesc/teihdr:msDesc/teihdr:head/teihdr:note[@type=\"caption\"]",
+                "xpath": "./teihdr:fileDesc/teihdr:titleStmt/teihdr:title",
```
sebastian-meyer:
```suggestion
                "xpath": "./teihdr:fileDesc/teihdr:titleStmt/teihdr:title | ./teihdr:fileDesc/teihdr:sourceDesc/teihdr:msDesc/teihdr:head/teihdr:note[@type=\"caption\"]",
```

sebastian-meyer:
There actually is a difference between a bibliographic title and a source description. Printed books typically have "real" bibliographic titles (which would be encoded in `titleStmt`), while medieval manuscripts mostly don't have any title. Instead they are identified by a short description of their contents, i. e. a `sourceDesc`.

That means we actually need to respect both: for medieval manuscripts the current behaviour (using `sourceDesc`) is correct, while we are currently missing a mapping for bibliographic titles (using `titleStmt`). I suggest using an union operator in the xpath to select both variants (see my review comment - **untested!**).

sebastian-meyer (changes requested):

Resources/Private/Data/MetadataDefaults.json
beatrycze-volk: Could you provide some document so I could test it for `sourceDesc`?

beatrycze-volk: XPath is incorrect:
`Core: Error handler (FE): PHP Warning: DOMXPath::evaluate(): Invalid expression in /var/www/webroot/public/typo3conf/ext/dlf/Classes/Common/MetsDocument.php line 542`

beatrycze-volk: Ok, so I have tested it and found out that problem is with the last part, then I have compared it with the old value from DFG Viewer and there it was looking so: https://github.com/slub/dfg-viewer/issues/70#issuecomment-3783913674

So question is, should actually the value from Live System be taken? Then it will look so:
`./teihdr:fileDesc/teihdr:titleStmt/teihdr:title | ./teihdr:fileDesc/teihdr:sourceDesc/teihdr:msDesc/teihdr:head/teihdr:title`

sebastian-meyer: Yes, that should be correct!

sebastian-meyer (approved):

PR #1814, opened 2026-01-22
[FEATURE] Document loading helper

marco-lengua:
In this pull request we replace the "obsolete" approach by loading documents one time with a service and fall back to the more easier to implement and straight forward approach of using a static helper method with a static field (i think this is what you meant with "singleton" like pattern) for one time loading of a document within abstract controller. The speed gain is slightly lesse but not significant. This PR is compatible with multiview and all other plugins.
The speed gain is based on the fact, that only the first loaded controller (plugin) makes an initial load of the document and all following controllers (plugins) use this loaded document.

### DIFF
--- a/Classes/Common/Helper.php
+++ b/Classes/Common/Helper.php
@@ -102,7 +102,14 @@ class Helper
         'chemical/x-xyz' => 'xyz',
         'text/html' => 'html'
     ];
-
+    public static function getDocumentInstance($documentLocation, $settings)
+    {
+        static $doc;
+        if($doc === null){
+            $doc = AbstractDocument::getInstance($documentLocation, $settings);
+        }
+        return $doc;
+    }
     /**
      * Generates a flash message and adds it to a message queue.
      *
--- a/Classes/Common/Helper.php
+++ b/Classes/Common/Helper.php
@@ -202,13 +202,11 @@ class Helper
         libxml_use_internal_errors($libxmlErrors);
         return $xml;
     }
-    
     /**
      * @param string $documentLocation The URL of XML file or the IRI of the IIIF resource
      * @param array $settings
      * 
      * @return AbstractDocument
-     *  
      */
     public static function getDocumentInstance($documentLocation, $settings): AbstractDocument|null
     {
--- a/Classes/Common/Helper.php
+++ b/Classes/Common/Helper.php
@@ -205,7 +205,7 @@ class Helper
     /**
      * @param string $documentLocation The URL of XML file or the IRI of the IIIF resource
      * @param array $settings
-     * 
+     *
      * @return AbstractDocument
      */
     public static function getDocumentInstance($documentLocation, $settings): AbstractDocument|null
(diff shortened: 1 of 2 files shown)


Classes/Common/Helper.php
```diff
@@ -102,7 +102,6 @@ class Helper
         'chemical/x-xyz' => 'xyz',
         'text/html' => 'html'
     ];
-
     /**
```
sebastian-meyer, line 106: Why did you remove this empty line? Please re-add it. ;o)

Classes/Common/Helper.php
```diff
@@ -203,6 +202,20 @@ public static function getXmlFileAsString($content): \SimpleXMLElement|false
         libxml_use_internal_errors($libxmlErrors);
         return $xml;
     }
+    /**
```
sebastian-meyer, line 214:
```suggestion
    }

    /**
```

Classes/Common/Helper.php
```diff
@@ -203,6 +202,20 @@ public static function getXmlFileAsString($content): \SimpleXMLElement|false
...
+     * @return AbstractDocument
+     */
+    public static function getDocumentInstance($documentLocation, $settings): AbstractDocument|null
+    {
+        static $doc;
+        if ($doc === null) {
+            $doc = AbstractDocument::getInstance($documentLocation, $settings);
+        }
+        return $doc;
+    }
```
sebastian-meyer, line 229: How does that work with Multiview? Calling this method with different parameters will always return the initial document, even if different `$documentLocation` are given. IMHO it has to look something like this:

```
    private static array $docs = [];

    public static function getDocumentInstance($documentLocation, $settings): AbstractDocument|null
    {
        $hash = md5($documentLocation);
        if (!isset(static::$docs[$hash])) {
            static::$docs[$hash] = AbstractDocument::getInstance($documentLocation, $settings);
        }
        return static::$docs[$hash];
    }
```

sebastian-meyer (changes requested):

Classes/Common/Helper.php
marco-lengua: fixed

Classes/Common/Helper.php
marco-lengua: fixed

Classes/Common/Helper.php
marco-lengua: You are absolutely right. I missed that point. The problem here for me is, that i acutally can't test against the mutliview plugin, because we don't use it in our instance. I provided an update with the suggested implementation. Could you please test if the mutliview would still work. Thank you

sebastian-meyer (approved):
Thank you very much, this looks fine, now!

Would you consider checking [all the other places where `AbstractDocument::getInstance()` is called](https://github.com/search?q=repo%3Akitodo%2Fkitodo-presentation+AbstractDocument%3A%3AgetInstance%28&type=code)? Maybe the new helper method could be used there as well?

PR #1819, opened 2026-01-27
[BUGFIX] Fix link rendering by using current ViewHelper tag instance

### LINKED ISSUE #1815: [BUG] `title` attribute is missing in kitodo:link ViewHelper
## Description

The title Attribute is missing in the kitodo:link ViewHelper.
So the final <a> link misses the (translated) Title.

```
<a title="Längs des Schienenstranges und der Wasserstraßen" 
   href="/viewer/xxx">

## missing:
<a href="/viewer/xxx">
```

fschoelzel:
closes: #1815

Avoid creating a new static ViewHelper instance during render(). Render $this->tag to keep Fluid-initialized attributes from parent. 
This change renders the already-initialized `$this->tag` instance provided by Fluid, ensuring all registered attributes and arguments like `title` are preserved.

### DIFF
--- a/Classes/ViewHelpers/LinkViewHelper.php
+++ b/Classes/ViewHelpers/LinkViewHelper.php
@@ -98,12 +98,10 @@ final class LinkViewHelper extends AbstractTagBasedViewHelper
             $uri .= '#' . $this->arguments['section'];
         }
 
-        $tag = new static();
-        $tag->tag->setTagName('a');
-        $tag->tag->addAttribute('href', $uri);
-        $tag->tag->setContent($childContent);
+        $this->tag->addAttribute('href', $uri);
+        $this->tag->setContent($childContent);
 
-        return $tag->tag->render();
+        return $this->tag->render();
     }
 
     /**


fschoelzel:
This [LinkViewHelper.php](https://github.com/kitodo/kitodo-presentation/blob/main/Classes/ViewHelpers/LinkViewHelper.php) strongly needs refactoring. It silently ignores additional params and surpasses fluid ViewHelper logic. I know it was a quickly needed, hard to sort out and complex Bugfix. And @markusweigelt did a great Job, figuring out the deep routing stack in TYPO3.
But it needs optimization or replacement in the future.
And an idea, how we tackle this issue: https://forge.typo3.org/issues/107026

markusweigelt:
Thx for your analysis. There are definitely a few smaller details that still need adjustment. I actually missed the `title` attribute in my DFG-Viewer and Kitodo.Presentation test cases 🥲 

Regarding the current implementation:
Extending https://github.com/TYPO3/typo3/blob/main/typo3/sysext/fluid/Classes/ViewHelpers/Link/ActionViewHelper.php
is unfortunately not an option, since the ViewHelper is marked as final, which makes it impossible to properly access or override the internal logic. Various approaches in different direction have already been tried (see also https://forge.typo3.org/issues/107026)

Using a custom ViewHelper also had some positive side effects:
The frontend complexity has been significantly reduced https://github.com/slub/slub_digitalcollections/commit/24083c322c6eccc537dc9c001e63e956146b2300 and beside we were also able to get rid of addQueryString="untrusted", which is not recommended anyway (cause it inflates the URL with unnecessary parameters).

I’m looking forward to an improved solution and feel free to keep me posted 😉

sebastian-meyer (approved):

PR #1801, opened 2026-01-15
[BUGFIX] Fix moving metadata in backend

beatrycze-volk:

PR #1831, opened 2026-02-02
[BUGFIX] Prevent index_name duplication on proper localization of metadata

michaelkubina:
This PR fixes an error, where the `index_name`  of a new (proper) localization gets duplicated to `index_name0` in the metadata repository.

This fix will not prevent duplication, when translations are created through new records and by manual linking them afterwards (wrong usage!). But it will not duplicate anymore when page translations in the backend are used on the storage folder and new translations are created via "Localize To" - which is the proper way to do so.

grafik

### DIFF
--- a/Configuration/TCA/tx_dlf_metadata.php
+++ b/Configuration/TCA/tx_dlf_metadata.php
@@ -82,7 +82,7 @@ return [
         ],
         'index_name' => [
             'exclude' => 1,
-            'l10n_mode' => '',
+            'l10n_mode' => 'exclude',
             'label' => 'LLL:EXT:dlf/Resources/Private/Language/locallang_labels.xlf:tx_dlf_metadata.index_name',
             'config' => [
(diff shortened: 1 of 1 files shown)


sebastian-meyer (approved):
Thanks for the fix!

But I am a bit confused here. This actually was `'l10n_mode' => 'exclude'` for the last 9 years and was just recently changed to `'l10n_mode' => ''` in #1801. That means this should have worked correctly since 2017...

beatrycze-volk:
> Thanks for the fix!
> 
> But I am a bit confused here. This actually was `'l10n_mode' => 'exclude'` for the last 9 years and was just recently changed to `'l10n_mode' => ''` in #1801. That means this should have worked correctly since 2017...

I would suspect that problem was the way we create our translations, [here](https://github.com/kitodo/kitodo-presentation/discussions/1832) @michaelkubina have written more about the whole topic of translations.

PR #1838, opened 2026-02-04
[BUGFIX] Update Form.html

michaelkubina:
There is a missing "<" in the partial, resulting in an exception, when the template is being rendered


beatrycze-volk (approved):

beatrycze-volk:
Thanks! :)

sebastian-meyer (approved):

PR #1839, opened 2026-02-04
[MAINTENANCE] Translate Helper Refactoring

marco-lengua:
This PR reduces excessive SQL backend querying for single entry translations and lazy loading of single entries to the cache variable. Now we have just a single call to the backend per request and table and PID (retrieving as batch) and else O(1) lookup for translations.

### DIFF
--- a/Classes/Controller/TableOfContentsController.php
+++ b/Classes/Controller/TableOfContentsController.php
@@ -35,8 +35,8 @@ class TableOfContentsController extends AbstractController
      */
     protected array $activeEntries = [];
 
-    protected array $translations;
-    protected string $languageContentId;
+    //protected array $translations;
+    //protected string $languageContentId;
 
     /**
      * The main method of the plugin
@@ -47,8 +47,8 @@ class TableOfContentsController extends AbstractController
      */
     public function mainAction(): ResponseInterface
     {
-        $this->translations = Helper::translate_test('tx_dlf_structures', $this->settings['storagePid']);
-        $this->languageContentId = Helper::translate_contextId();
+        //$this->translations = Helper::translate_test('tx_dlf_structures', $this->settings['storagePid']);
+        //$this->languageContentId = Helper::translate_contextId();
         // Load current document.
         $this->loadDocument();
 
@@ -322,8 +322,8 @@ class TableOfContentsController extends AbstractController
      */
     private function getTranslatedType(string $type): string
     {
-        return Helper::translate_get($type, $this->translations, $this->languageContentId);
-        //return Helper::translate($type, 'tx_dlf_structures', $this->settings['storagePid']);
+        //return Helper::translate_get($type, $this->translations, $this->languageContentId);
+        return Helper::translate($type, 'tx_dlf_structures', $this->settings['storagePid']);
     }
 
     /**
--- a/Classes/Controller/TableOfContentsController.php
+++ b/Classes/Controller/TableOfContentsController.php
@@ -46,7 +46,6 @@ class TableOfContentsController extends AbstractController
     {
         // Load current document.
         $this->loadDocument();
-
         if ($this->isDocMissing()) {
             // Quit without doing anything if required variables are not set.
             return $this->htmlResponse();
@@ -57,7 +56,6 @@ class TableOfContentsController extends AbstractController
         }
 
         $this->view->assign('viewData', $this->viewData);
-    
         return $this->htmlResponse();
     }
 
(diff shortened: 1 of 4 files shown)


sebastian-meyer (dismissed):

Classes/Controller/TableOfContentsController.php
```diff
@@ -137,7 +137,7 @@ private function getMenuEntry(array $entry, bool $recursive = false): array
         }
         $entryArray['year'] = $entry['year'];
         $entryArray['orderlabel'] = $entry['orderlabel'];
-        $entryArray['type'] = $entry['type'];
+        $entryArray['type'] = $this->getTranslatedType($entry['type']);
```
beatrycze-volk: This change disables the option to display document type icons. This functions is used for e.g. in DFG Viewer `<span class="meta-type-icon meta-type-{child.type}">{child.type}</span>`. It is necessary that type is in English instead be translated here.

Resources/Private/Partials/TableOfContents/Title.html
```diff
@@ -18,7 +18,7 @@
             <f:format.crop maxCharacters="55" append="&nbsp;..."><f:format.htmlspecialchars doubleEncode="false">{child.title}</f:format.htmlspecialchars></f:format.crop>
         </f:then>
         <f:else>
-            <f:format.crop maxCharacters="55" append="&nbsp;..."><f:format.htmlspecialchars doubleEncode="false"><f:translate key="LLL:EXT:dlf/Resources/Private/Language/locallang_structure.xlf:structure.{child.type}" /></f:format.htmlspecialchars><f:if condition="{child.volume}"> &nbsp;<f:format.htmlspecialchars doubleEncode="false">{child.volume}</f:format.htmlspecialchars></f:if></f:format.crop>
```
beatrycze-volk, line 21: Translation should be kept here.

beatrycze-volk (changes requested):

Classes/Controller/TableOfContentsController.php
marco-lengua: thank you for your feedback, i was not aware of this. What i don't understand is why in the current version (branch 6) it is the same way. I think we made the change in one PR and now wanted to revert it. The purpose of this is a batch translation and a speed gain. It was like this before. Could you explain why it would not be working anymore now.

Resources/Private/Partials/TableOfContents/Title.html
marco-lengua: If we have the translated "entry.Type" we can use that translation in this fluid template, because it makes the page load much faster. That was the purpose of this PR. Do you know any case this is not feasible? Could you please provide me with more information on this topic. Please see also the GitHub Discussion for our usecases

Classes/Controller/TableOfContentsController.php
beatrycze-volk: The use case is as mentioned above, the template contains span with the `meta-type-{child.type}`, then there is css for this class e.g.

```
&.meta-type-article {
    background-position: -60px -30px;
}
```

If this is translated than css class is not working anymore.

Resources/Private/Partials/TableOfContents/Title.html
beatrycze-volk: image

Here is the example how this icons are used.
You can see next to the year the small newspaper icon.

Classes/Controller/TableOfContentsController.php
michaelkubina: Hi, just checking in...i believe here is a little misunderstanding @beatrycze-volk . What marco meant is that the lines you are pointing out are just undoing a PR i made for 7.x , that was meant to resolve performance issues: https://github.com/kitodo/kitodo-presentation/pull/1771
Marco solved the performance issue differently (at the root actually), so the cited lines are simply undoing my rather ugly fix that did not account dynamic translations from backend configuration.
But now you are pointing out CSS classes being affected by reverting to the state that is was prior to my "fix", which i as well was not aware of... This means that in Version 6.0.0 the CSS-classes must have already been broken (my PR was not applied there) as well.

The ToC controller seemingly does two things with types: it did translate them, so that the Title was using translated types. But it also used them for css-classes -> so translations are unwanted here, but thats how it is in 6.x! We cant resolve both with one kind of type...so i would suggest using {child.type} with the untranslated type (index_name) for css-classes and {child.translated_type} for the translated Label that is being used in the Title of ToC-Entries. Marco will add those to the commit

beatrycze-volk: > so i would suggest using {child.type} with the untranslated type (index_name) for css-classes and {child.translated_type} for the translated Label that is being used in the Title of ToC-Entries. Marco will add those to the commit

Sounds good for me!

sebastian-meyer (approved):

PR #1846, opened 2026-02-11
[MAINTENANCE] Adjust css class names in TOC Plugin

beatrycze-volk:


sebastian-meyer (commented):
Why the renaming? Everybody who uses the default templates will have to adjust their CSS files...

beatrycze-volk:
Because we use here mixed naming conventions:

```
<div class="tx-dlf-tableofcontents">
    <ul>
        <f:render partial="TableOfContents/Children" arguments="{children: toc, viewData: viewData}"/>
    </ul>
</div>
```

Ine the `Main.html` we have `tx-dlf-tableofcontents`, and then `tx-dlf-toc`. It would be good to unify it. It matches the standard approach where css classes are created 'tx-dlf-pluginname-xyz'.

Plus new names are more clear and descriptive.

sebastian-meyer (approved):
OK, I'll try to remember to include a hint in the release notes. ;o)

PR #1856, opened 2026-02-24
[BUGFIX] Fix date range sanitization in SearchController

### LINKED ISSUE #1852: [BUG] Date search fails if only "from" date is given

stweil:


stweil:
This should fix issue #1852.

sebastian-meyer (approved):

PR #1860, opened 2026-02-25
[DOCUMENTATION] Update plugins documentation

beatrycze-volk:


Documentation/Plugins/Index.rst
```diff
@@ -237,9 +235,13 @@ now.
 Collection
 ----------
 
-The collection plugin shows one collection, all collections or selected collections.
+The collection plugin shows one collection, all collections or selected collections. It can be used to display the list
+of collections or the overview of single collection. The plugin is configured in the TypoScript and in the FlexForm.
```
sebastian-meyer:
```suggestion
of collections or the contents of a single collection. The plugin is configured in the TypoScript and in the FlexForm.
```

Documentation/Plugins/Index.rst
```diff
@@ -584,7 +708,23 @@ Features `pagesBackward` and `pagesForward` offer adding multiple navigation but
    :Default:
       By default all features are activated. The selection is stored as comma separated list.
 
-       doublePage,pageFirst,pageBack,pageStepBack,pageSelect,pageForward,pageStepForward,pageLast,litView,zoom,rotation,measureForward,measureBack
+       doublePage,pageFirst,pageBack,pageStepBack,pageSelect,pageForward,pageStepForward,pageLast,listView,zoom,rotation,measureForward,measureBack
```
sebastian-meyer:
```suggestion
```
Is this duplication intentional?

Documentation/Plugins/Index.rst
```diff
@@ -1006,34 +1238,65 @@ Toolbox
...
+       * tx_dlf_fulltextdownloadtool
        * tx_dlf_fulltexttool
        * tx_dlf_imagedownloadtool
        * tx_dlf_imagemanipulationtool
+       * tx_dlf_modeldownloadtool
+       * tx_dlf_multiviewaddsourcetool
        * tx_dlf_pdfdownloadtool
-       * tx_dlf_fulltextdownloadtool
        * tx_dlf_searchindocumenttool
        * tx_dlf_scoretool
```
sebastian-meyer: Shouldn't we only propose one variant of each tool here? (I'd prefer the option without `tx_dlf_` prefix.)

sebastian-meyer (changes requested):

Documentation/Plugins/Index.rst
beatrycze-volk: We can do it so, I would also prefer only variant without the prefix.

Documentation/Plugins/Index.rst
beatrycze-volk: Duplication? It is correction from '_litView_' to '_listView_'.

sebastian-meyer: Sorry, that was misleading! I was refering to the following lines where all options are listed again. So, currently there are all options as comma-separated list and then immediately afterwards all options again as a pointed list. IMHO the pointed list under `Values:` alone would be sufficient.

Documentation/Plugins/Index.rst
sebastian-meyer:
```suggestion
       * annotationtool
       * audiovideotool
       * fulltextdownloadtool
       * fulltexttool
       * imagedownloadtool
       * imagemanipulationtool
       * modeldownloadtool
       * multiviewaddsourcetool
       * pdfdownloadtool
       * searchindocumenttool
       * scoretool
```

Documentation/Plugins/Index.rst
beatrycze-volk: Actually not all options are here, currently two more options are possible: 
* pagesBackward
* pagesForward

They replace single forward and backward navigations.

sebastian-meyer (approved):

Issue #1167, opened 2024-02-04
Replace JPlayer


stweil:
## Description

JPlayer is one of the libraries used by Kitodo.Presentation. It was added to the code base in commit e7a21a478d5eb6fae9b9ff22463c4191c5d9e39a to get an audio player GUI.

That library was last released ten years ago, a related website still uses Wordpress 3.9, the last commit was seven years ago, issues and pull requests are not answered. So there are a lot of indicators that jPlayer is unmaintained.

I now made a last try here: https://github.com/jplayer/jPlayer/issues/436

If jPlayer remains unmaintained, it should be replaced because unmaintained code is a security risk.
Meanwhile HTML 5 is standard, and using an extra library for an audio player might no longer be needed.

## Expected Benefits of this Development

This is a code maintenance task, necessary to eliminate potential risks by unmaintained code.

## Estimated Costs and Complexity

Hopefully the costs for this task are low (less than 5 working days).

sebastian-meyer:
As far as I am aware, @fschoelzel at SLUB Dresden is working on implementing [shaka-player](https://github.com/shaka-project/shaka-player) for video. Maybe we can use that for audio as well and get rid of jPlayer that way?

stweil:
Maybe. But perhaps it is sufficient to use simple HTML 5 code for audio. I updated the description above.

stweil:
Maybe. But perhaps it is sufficient to use simple HTML 5 code for audio. There even exist [code generators for audio players](https://www.sven-hoerig.de/html-code-generator/audio/). I updated the description above.

fschoelzel:
After the implementation of the dev-mediaplayer branch (PR: #818), the default audio player is the shaka-player integration, due to common desired features between audio and video files, and also to keep the UI the same. On my current development branch, shaka-player is running on the latest version and is well maintained, but I need to remove the old JPlayer functionality.

sebastian-meyer:
Votes: 0 (because this doesn't need more funding)

stweil:
@fschoelzel, it looks like JPlayer is still used?
```
git grep -li jplayer
.github/codeql.yml
Classes/Controller/AudioPlayerController.php
Configuration/TypoScript/setup.typoscript
Resources/Public/JavaScript/AudioPlayer/AudioPlayer.js
Resources/Public/JavaScript/jPlayer/blue.monday/css/jplayer.blue.monday.css
[...]

PR #1320, opened 2024-08-28
[FEATURE] Integrate media player from slub_web_sachsendigital

fschoelzel:
This PR integrates the video/audio player from slub_web_sachsendigital into a new MediaPlayer plugin. 
The PR is based on https://github.com/kitodo/kitodo-presentation/pull/796 and https://github.com/kitodo/kitodo-presentation/pull/818 from @dvoracek-slub 

> For testing this within the DFG Viewer, I have prepared a DDEV system at https://github.com/dvoracek-slub/ddev-dfgviewer-dist/tree/mediaplayer. It relies on https://github.com/dvoracek-slub/dfg-viewer/tree/dev-mediaplayer and includes a demo METS video document.
> 
> There also is a work-in-progress section in the documentation that describes some features and technical aspects of the player. `composer docs:serve` can be used for local preview.
> 
> Update (2022-07-06). Most notable changes:
> 
>     * Kitodo in general
>       
>       * Read production place and person names in `Mods` format class. I use this to test GND linking of some audio-related metadata.
>       * Add option to show full TOC.
> 
>     * Media player
>       
>       * It is now possible to set and export markers/segments in the media. The list is shown in audio mode (press F2).
>       * When the METS links preprocessed waveform data, this can be shown (and be used to edit markers). See the DDEV system for an example.
>       * First version of handling recordings that are split across multiple media files. Currently, this only works via page reloads.

PR #133, opened 2016-06-16
Audio support

jacmendt:
Adds basic functionality for playing audio files with kitodo-presentation. It is based on [jPlayer](http://jplayer.org/) and supports HTML5 and Flash playing. 

For testing the audio data was encoded like this:

``` xml
<mets:fileGrp USE="AUDIO">
    <mets:file ID="FILE_0001_MAX" MIMETYPE="audio/mpeg">
        <mets:FLocat xmlns:xlink="http://www.w3.org/1999/xlink" LOCTYPE="URL"
           xlink:href="http://localhost:8000/juicy.mp3"/>
    </mets:file>
</mets:fileGrp>
```

PR #1878, opened 2026-03-16
[MAINTENANCE] Remove jPlayer and dlf_audioplayer Plugin

fschoelzel:
- This is a **breaking change** for existing tx-dlf-audioplayer use
- Completly removed the tx-dlf-audioplayer plugin and the deprecated jPlayer lib
- Changes may need to be applied for thirdparty extensions depending on kitodo.presentation 
(eg. dfg-viewer - see: https://github.com/slub/dfg-viewer/blob/master/Resources/Private/Partials/PageView.html#L69)
- the `tx-dlf-audioplayer` plugin is replaced by the `tx-dlf-mediaplayer` plugin using the [shaka-player](https://github.com/shaka-project/shaka-player) library, which is now used for both audio and video mediatypes (first introduced with this PR: #1320)

This change should be tagged, or available at v7.0 first.

This close the Issue: https://github.com/kitodo/kitodo-presentation/issues/1167
And rollback the `tx-dlf-audioplayer` integration PR: https://github.com/kitodo/kitodo-presentation/pull/133


stweil:
@fschoelzel, could you please update the commit to resolve the new conflicts? Thanks!

beatrycze-volk (approved):

sebastian-meyer (approved):

PR #1890, opened 2026-03-24
[BUGFIX] Move fulltext container placed inside the map container

beatrycze-volk:


sebastian-meyer:
Are you sure this `div` is not needed by any of the fulltext features?

beatrycze-volk:
This div is inside the map div, what means that that fulltext is show on top of the page in the same container as image. The same div we have in slub_digitalcollections and dfg-viewer, but there they are outside of the image container.

sebastian-meyer:
> The same div we have in slub_digitalcollections and dfg-viewer, but there they are outside of the image container.

So the `div` is needed then. Therefore we should not remove it entirely, but just move it outside the image container.

beatrycze-volk:
This `div` should not be here, and even more it should not have the same `id`. My suspiscion is that someone has randomly copied this element from `slub_digitalcollections` or `dfg-viewer` without thinking that it is not allowed to have two `div`s with the same `id`.

beatrycze-volk:
In general I would strongly assumed that template for `PageView` plugin should not contain `div` element which is a container for tool from `Toolbox` plugin. Is there any explanation why this annotations and fulltext elements were inserted there?

beatrycze-volk:
Eventually we can move it below the map div, but then we need the full container for fulltext and not only div for selection of the text, something like that:

```
<div class="fulltext-container">
    <div id="tx-dlf-fulltextselection"></div>
</div>
```

But it means that we have problem if we want to include also fullltext search as this is made e.g. in this way:

```
<div class="fulltext-container">
    <div class="tx-dlf-toolsFulltextsearch">
        <div class="fulltext-search-toggle"></div>
        <f:cObject typoscriptObjectPath="lib.kitodo.fulltext.search" />
    </div>
    <div id="tx-dlf-fulltextselection"></div>
</div>
```

Those snippets are taken from `dfg-viewer` (https://github.com/slub/dfg-viewer/blob/40895d4ea943426e06a7052e578c9c11f33ebf86/Resources/Private/Partials/PageView.html#L76) and `slub_digitalcollection `(https://github.com/slub/slub_digitalcollections/blob/6173b82ce9edc0b8f2e5c31c815f83ac42593915/Resources/Private/Partials/KitodoNewspaper.html#L66).

sebastian-meyer (approved):
Thank you!

PR #1897, opened 2026-03-27
Calculation corrected for pages > 1; error for pages > 9

BFallert:
calculation is incorrect for pages > 9, so "jump to first page" is disabled.


BFallert:
My fix works only in Typo3 Version 12, not in Version 13.

BFallert:
My Changes work localy in my typo3 version 12, but tests are not ok. So i close my pr.

PR #1931, opened 2026-04-23
[DOCUMENTATION] Improve and standardize JSDoc comments

beatrycze-volk:
Standardizes type casing (e.g., number instead of Number), fixes various tags like @augments and @returns, and improves the formatting of JSDoc annotations across the codebase for better consistency and tool compatibility.

### DIFF
--- a/Resources/Private/JavaScript/DlfMediaPlayer/3rd-party/VideoFrame.js
+++ b/Resources/Private/JavaScript/DlfMediaPlayer/3rd-party/VideoFrame.js
@@ -55,15 +55,15 @@ var VideoFrame = function(options) {
  * FrameRates - Industry standard frame rates
  *
  * @namespace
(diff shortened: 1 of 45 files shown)


sebastian-meyer (commented):
I suggest not changing any third party libraries as long as we don't need to patch them for functional or security reasons. All changes would be overwritten anyways when we update those dependencies...

beatrycze-volk:
Should third party libraries not be minimizied and excluded from the Codacy checks?

sebastian-meyer:
> Should third party libraries not be minimizied and excluded from the Codacy checks?

Yes, they should be. But I can only exclude directories from checks, so third-party libraries must not be placed in the same directories as our own code.

sebastian-meyer (approved):

PR #1933, opened 2026-04-24
[MAINTENANCE] Add type hints to `OaiPmhController`

beatrycze-volk:
Modernize the controller by adding PHP type hints to class properties, method parameters, and return types.


sebastian-meyer (approved):
Awesome! I very much appreciate your refactoring, clean-up, tests, etc.! Great work!

PR #1937, opened 2026-04-27
[MAINTENANCE] Remove redundant `formatsLoaded` flag

beatrycze-volk:
The loading state of the available metadata formats can be directly inferred from the emptiness of the `$formats` array. This change removes the dedicated `$formatsLoaded` property and its associated logic, simplifying the format loading process.


sebastian-meyer:
I believe this would break Kitodo.Presentation. You removed the hard-coded formats (OAI, METS and XLINK) which are not part of the default formats either (see https://github.com/kitodo/kitodo-presentation/blob/main/Resources/Private/Data/FormatDefaults.json).

We could add them to the defaults, but that would allow deleting or changing those formats in the backend, which again would break Kitodo.Presentation...

METS, OAI and XLINK are hard-coded for a reason, because they are basically minimum requirements.

I'd like to dismiss this pull request. Are you fine with that?

beatrycze-volk:
Yes, if you think that it breaks the Kitodo.Presentation, then you can close it.

PR #1941, opened 2026-04-28
[MAINTENANCE] Introduce `AbstractRepository` for unified storage PID management

beatrycze-volk:
Currently functions like `findAll()` and custom functions which use `createQuery()` do not work, becuase pid in the generated SQL is always 0. It is necessary to set up `storagePid`.

See:
https://docs.typo3.org/permalink/t3coreapi:extbase-repository-query-setting
https://docs.typo3.org/permalink/t3coreapi:typo3-cms-extbase-persistence-repository-setdefaultquerysettings

### DIFF
--- a/Tests/Functional/Controller/CollectionControllerTest.php
+++ b/Tests/Functional/Controller/CollectionControllerTest.php
@@ -84,7 +84,7 @@ class CollectionControllerTest extends AbstractControllerTestCase
             'storagePid' => self::$storagePid,
             'solrcore' => self::$solrCoreId,
             'collections' => '1',
-            'showSingle' => 'some_value',
+            'showSingle' => '0',
             'randomize' => ''
         ];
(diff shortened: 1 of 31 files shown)


Tests/Functional/Controller/CollectionControllerTest.php
```diff
@@ -79,11 +81,11 @@ public function canListActionForwardToShow()
     public function canShowAction()
     {
         $settings = [
+            'storagePid' => self::$storagePid,
             'solrcore' => self::$solrCoreId,
             'collections' => '1',
-            'showSingle' => '0',
-            'randomize' => '',
-            'storagePid' => self::$storagePid
+            'showSingle' => 'some_value',
```
sebastian-meyer:
```suggestion
            'showSingle' => '0',
```

sebastian-meyer (approved):

PR #1951, opened 2026-04-29
[MAINTENANCE] Add query methods for record visibility and standardize default settings

beatrycze-volk:
Depends on #1946

### DIFF
--- a/Classes/Domain/Repository/AbstractRepository.php
+++ b/Classes/Domain/Repository/AbstractRepository.php
@@ -48,7 +48,6 @@ class AbstractRepository extends Repository
      *
      * @return void
      */
-
     public function activateDebugMode(): void
     {
         $this->debug = true;
@@ -60,7 +59,7 @@ class AbstractRepository extends Repository
      * @access public
      *
      * @return void
-        */
+     */
     public function ignoreDeleted(): void
     {
         $querySettings = $this->getDefaultQuerySettings();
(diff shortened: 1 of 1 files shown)


Classes/Domain/Repository/AbstractRepository.php
```diff
@@ -47,11 +48,40 @@ class AbstractRepository extends Repository
...
+     * @access public
+     *
+     * @return void
+     */
+    public function ignoreHidden(): void
+    {
+        $querySettings = $this->getDefaultQuerySettings();
+        $querySettings->setIgnoreEnableFields(false);
+        $this->setDefaultQuerySettings($querySettings);
+    }
```
sebastian-meyer, line 82: Are you sure this really does what it says? The method is called `ignoreHidden`, but it sets `ignoreEnableFields` to `false`. That's confusing...

Classes/Domain/Repository/AbstractRepository.php
```diff
@@ -118,4 +148,46 @@ protected function debugQueryBuilder(QueryBuilder $queryBuilder): void
...
+     * @access public
+     *
+     * @return void
+     */
+    public function showHidden(): void
+    {
+        $querySettings = $this->getDefaultQuerySettings();
+        $querySettings->setIgnoreEnableFields(true);
+        $this->setDefaultQuerySettings($querySettings);
+    }
```
sebastian-meyer: Same as above: I think this should be `setIgnoreEnableFields(false)`, right?

sebastian-meyer: Also, the phpDoc is wrong ("Sets hidden records to be *not* returned..."). ;)

Classes/Domain/Repository/AbstractRepository.php
beatrycze-volk: Yes, this method does eactly this what it says. If `setIgnoreEnableFields(false)`, then hidden objects are not returned, if it is set to `true`, then hidden objects are returned. TYPO3 often makes stuff confusing.

```
    /**
     * An array of column names in the enable columns array (array keys in $GLOBALS['TCA'][$table]['ctrl']['enablecolumns']),
     * to be ignored while building the query statement. Adding a column name here effectively switches off filtering
     * by this column. This setting is only taken into account if $this->ignoreEnableFields = TRUE.
     *
     * @see setIgnoreEnableFields()
     */
    public function setEnableFieldsToBeIgnored(array $enableFieldsToBeIgnored): self
    {
        $this->enableFieldsToBeIgnored = $enableFieldsToBeIgnored;
        return $this;
    }
```

So, if it is set to` true`, `hidden` field is ignored and also hidden records are shown. I have looked a bit about it in the internet, as I was also not sure what does this function do.

See:
https://code711.de/tutorials/typo3/apply-defaultquerysettings-to-magic-methods-in-a-repository
https://www.in2code.de/en/recent/pass-hidden-objects-to-extbase-actions/
https://talk.typo3.org/t/show-hidden-records-in-the-frontend/5817

Or do I understand it wrong?

sebastian-meyer: Yes, you are right. I was confused! Sorry!

sebastian-meyer (approved):

PR #1955, opened 2026-04-30
[FEATURE] Introduce global debug mode for Fluid templates

beatrycze-volk:
This change adds an `activateDebugMode()` method to `AbstractController` which assigns a `debugActive` flag to the view. Numerous Fluid templates are updated to conditionally render `f:debug title="all">{_all}` when `debugActive` is true.

This provides a centralized way for developers to easily inspect all variables passed to a template during development by activating debug mode in the relevant controller.


sebastian-meyer (changes requested):
That's very convenient! Could you please add a hint to the documentation?

sebastian-meyer (approved):
Thank you very much!

PR #1959, opened 2026-05-06
[MAINTENANCE] Enhance type hints and docblocks in `Helper` class

beatrycze-volk:
This commit refines various method signatures and their docblock annotations in the `Helper` class. Changes include adding `mixed` type hints for parameters, specifying nullable return types, providing more precise array return type hints, and adding `@throws` annotations where exceptions can occur. Additionally, a minor typo in a comment is corrected, a comment describing a lookup table is clarified, and an unused `use` statement is removed.


Classes/Common/Helper.php
```diff
@@ -126,6 +127,8 @@ class Helper
      * @param string $queue The queue's unique identifier
      *
      * @return FlashMessageQueue The queue the message was added to
+     *
+     * @throws \TYPO3\CMS\Core\Exception
```
sebastian-meyer: I don't see any `throw` statement in the method. The phpDoc should only document exceptions thrown by this method. Exceptions thrown by calls to other methods should be documented at the respective methods.
If it's necessary to explicitly handle this exception further down the stack, we should catch and re-throw the exception here for clarity. Then the phpDoc would be correct.

Classes/Common/Helper.php
```diff
@@ -377,6 +380,8 @@ public static function getHookObjects(string $scriptRelPath): array
      * @param int $pid Get the "index_name" from this page
      *
      * @return string "index_name" for the given UID
+     *
+     * @throws \Doctrine\DBAL\Exception
```
sebastian-meyer: See above.

Classes/Common/Helper.php
```diff
@@ -482,7 +487,9 @@ public static function getLanguageName(string $code): string
      *
      * @param int $pid Get the "index_name" from this page only
      *
-     * @return mixed[]
+     * @return array<int|string, string> Array mapping record UID to its index_name
+     *
+     * @throws \Doctrine\DBAL\Exception
```
sebastian-meyer: See above.

Classes/Common/Helper.php
```diff
@@ -542,7 +550,9 @@ public static function isValidHttpUrl(string $url): bool
      * @param bool $reverseOrder Should the data map be reversed?
      * @param bool $cmdFirst Should the command map be processed first?
      *
-     * @return mixed[] Array of substituted "NEW..." identifiers and their actual UIDs.
+     * @return array<string,int> Array of substituted "NEW..." identifiers and their actual UIDs.
+     *
+     * @throws AspectNotFoundException
```
sebastian-meyer: See above.

Classes/Common/Helper.php
```diff
@@ -648,6 +658,8 @@ public static function timeCodeToSeconds(string $timeCode): float
      * @param string $pid Get the translation from this page
      *
      * @return string Localized label for $indexName
+     *
+     * @throws AspectNotFoundException
```
sebastian-meyer: See above.

sebastian-meyer (changes requested):

sebastian-meyer (approved):

PR #1961, opened 2026-05-06
[DOCUMENTATION] Improve and standardize PHPDoc array return types

beatrycze-volk:


Classes/Controller/CollectionController.php
```diff
@@ -222,7 +223,9 @@ public function showSortedAction(): ResponseInterface
      * @param array<Collection> $collections to be processed
      * @param Solr $solr for query
      *
-     * @return mixed[]
+     * @return array<int,array{collection:Collection,info:array<string,mixed>}> Processed collections keyed by priority-based integer
+     *
+     * @throws RandomException
```
sebastian-meyer: phpDoc should only contain exceptions thrown by this method. There is no `throw` statement in this method.

sebastian-meyer (changes requested):

sebastian-meyer (approved):

PR #1970, opened 2026-05-11
[BUGFIX] Fix Verovio scaling

sebastian-meyer:
This fixes a scaling problem with PDF files generated by Verovio where some lines were missing when using landscape format.
It also justifies the lines in the Verovio SVG rendering for both landscape and portrait format.


markusweigelt (approved):
Currently, I can only review it through the code, and it looks good. I’m not able to perform a functional test at the moment.

PR #2000, opened 2026-06-16
[BUGFIX] Initialize frontend TypoScript for expression language functions

beatrycze-volk:
Ensures that the TypoScript context is properly initialized with the current page ID when document expression language functions are evaluated. This allows expressions to correctly resolve TypoScript settings that depend on the frontend page.


ByteParty (approved):

stweil:
s/[Bugfix/[Bugfix]/ (closing "]" was missing in the subject text).

sebastian-meyer (approved):

PR #2006, opened 2026-06-18
[MAINTENANCE] composer: Ignore security advisories for TYPO3 12.4 packages

stweil:
TYPO3 12.4 LTS reached end-of-life but is still used in Kitodo installations. Add policy.advisories.ignore for the TYPO3 12.4 security advisories so that composer update works with ^12.4|^13.4 constraints.

Assisted-by: minimax-m2.7 (MiniMax)

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -135,6 +135,58 @@
     "allow-plugins": {
       "typo3/class-alias-loader": true,
       "typo3/cms-composer-installers": true
+    },
+    "policy": {
+      "advisories": {
+        "ignore": [
+          "PKSA-2yr3-by9d-r1gh",
+          "PKSA-q3vc-jr63-rrrj",
+          "PKSA-hqhs-7j5f-td2d",
+          "PKSA-gj1k-954p-nhmk",
+          "PKSA-gr4f-6g49-cg8v",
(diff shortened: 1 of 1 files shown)


stweil:
The CI tests which require an installation of T12 were broken since June 11 because of new security advisories. This is fixed now.

stweil:
Link: https://packagist.org/api/security-advisories/?packages[]=typo3/cms-backend

beatrycze-volk:
Please take a look at this change: https://github.com/slub/slub_digitalcollections/pull/273/changes

stweil:
> Please take a look at this change: https://github.com/slub/slub_digitalcollections/pull/273/changes

Yes, that's an alternative which would also work. But I think that ignoring all security advisories for all TYPO3 release would not be a good idea.

Unrelated: `git config --global --add safe.directory /app` from this change should also be added to the Kitodo.Presentation CI.

stweil:
There are currently 10 pending pull requests with broken CI tests. Therefore I think that the PR here (or any other fix for the CI) is urgent.

sebastian-meyer (approved):

PR #2028, opened 2026-06-23
[MAINTENANCE] Enhance Solr core ID retrieval in BaseCommand

beatrycze-volk:
The `getSolrCoreUid` method is made more robust by explicitly checking for string keys within the `$solrCores` array. This prevents potential `Undefined index` errors if `$inputSolrId` is a string that does not correspond to an existing core.

Additionally, the method now clarifies its return behavior by explicitly returning `null` if a valid Solr core UID cannot be determined from the provided input.

### DIFF
--- a/Classes/Command/BaseCommand.php
+++ b/Classes/Command/BaseCommand.php
@@ -136,11 +136,11 @@ class BaseCommand extends Command
     protected function getSolrCoreUid(array $solrCores, bool|string|null $inputSolrId): ?int
     {
         if (MathUtility::canBeInterpretedAsInteger($inputSolrId)) {
-            $solrCoreUid = MathUtility::forceIntegerInRange((int) $inputSolrId, 0);
-        } else {
-            $solrCoreUid = $solrCores[$inputSolrId];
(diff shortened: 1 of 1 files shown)


Classes/Command/BaseCommand.php
```diff
@@ -136,11 +136,11 @@ protected function initializeRepositories(int $storagePid): void
     protected function getSolrCoreUid(array $solrCores, bool|string|null $inputSolrId): ?int
     {
         if (MathUtility::canBeInterpretedAsInteger($inputSolrId)) {
-            $solrCoreUid = MathUtility::forceIntegerInRange((int) $inputSolrId, 0);
-        } else {
-            $solrCoreUid = $solrCores[$inputSolrId];
+            return MathUtility::forceIntegerInRange((int) $inputSolrId, 0);
+        } else if (is_string($inputSolrId) && array_key_exists($inputSolrId, $solrCores)) {
```
sebastian-meyer:
```suggestion
        } elseif (is_string($inputSolrId) && array_key_exists($inputSolrId, $solrCores)) {
```

sebastian-meyer (approved):

PR #2054, opened 2026-07-28
[MAINTENANCE] Enhance PHPStan type inference and compliance for repositories

beatrycze-volk:
The `DocumentRepository` now includes a more precise `@method` PHPDoc type hint for `findAll()`, allowing PHPStan to correctly infer its return type as `array|QueryResultInterface`. This enables the removal of an `@phpstan-ignore-line` annotation in `ReindexCommand`.

For `NewTenantController`, a PHPStan error regarding the `getFirst()` method on `QueryResultInterface` has been moved from an inline ignore to a configuration-based ignore in `phpstan_12.4.neon`. This centralizes the ignore for a specific type inference limitation.


beatrycze-volk:
PHPStan needs to be restarted:
`Error: Could not authenticate against github.com`

sebastian-meyer (approved):

PR #2056, opened 2026-07-28
[MAINTENANCE] Expose MetsDocument::$mets property publicly

beatrycze-volk:
This allows direct access to the `SimpleXMLElement` object representing the METS document part, facilitating external processing or modification.

Error comes from slub_digitalcollections: 
`Error: Access to protected property Kitodo\Dlf\Common\MetsDocument::$mets.`

See: 
https://github.com/slub/slub_digitalcollections/actions/runs/30346470277/job/90233742182
https://github.com/slub/slub_digitalcollections/blob/602891882d97545dd35b5ef1f08c858e93cf53a6/Classes/ViewHelpers/XpathViewHelper.php#L129

### DIFF
--- a/Classes/Common/MetsDocument.php
+++ b/Classes/Common/MetsDocument.php
@@ -326,7 +326,7 @@ final class MetsDocument extends AbstractDocument
     }
 
     /**
-     * Return clone of METS for external uage.
+     * Return clone of METS for external usage.
      *
      * @access public
      *
--- a/Classes/Common/MetsDocument.php
+++ b/Classes/Common/MetsDocument.php
@@ -177,10 +177,10 @@ final class MetsDocument extends AbstractDocument
     protected bool $fileGrpsLoaded = false;
 
     /**
-     * @access public
+     * @access protected
      * @var SimpleXMLElement|null This holds the XML file's METS part as SimpleXMLElement object
      */
-    public ?SimpleXMLElement $mets = null;
+    protected ?SimpleXMLElement $mets = null;
 
     /**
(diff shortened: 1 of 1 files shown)


sebastian-meyer (commented):
I am not sure publicly exposing the SimpleXML object is a good idea. How do we make sure the object stays valid and immutable? This would be especially important with the upcoming features for rights-protected documents.
I was thinking of using a getter method instead to make the SimpleXML object accessible. That would make the property read-only, but you could still change the SimpleXML...

The safest option would be to have a getter method which returns a **copy** of the SimpleXML object. What do you think?

Classes/Common/MetsDocument.php
```diff
@@ -325,6 +325,18 @@ public function getFileLocation(string $id): string
         }
     }
 
+    /**
+     * Return clone of METS for external uage.
```
```suggestion
     * Return clone of METS for external usage.
```

Classes/Common/MetsDocument.php
```diff
@@ -325,6 +325,18 @@ public function getFileLocation(string $id): string
...
+     *
+     * @access public
+     *
+     * @return SimpleXMLElement|null
+     */
+    public function getMets(): ?SimpleXMLElement
+    {
+        return clone $this->mets;
```
sebastian-meyer: Since `$this->mets` can be `null` and `clone` cannot be called on a non-object, there has to be a check first.

sebastian-meyer (changes requested):

sebastian-meyer (approved):

PR #2095, opened 2026-08-04
[BUGFIX] Adjust Extbase persistence storage PID in CLI commands

beatrycze-volk:
Inject `ConfigurationManager` into `BaseCommand` to explicitly set the Extbase framework's `persistence.storagePid`. This ensures that all Extbase-managed repositories and queries within a CLI command consistently use the command's defined `storagePid`, preventing potential inconsistencies in data save. While individual repositories are already configured, this change aligns the global Extbase configuration for robustness in the CLI context.

There was too much removed in https://github.com/kitodo/kitodo-presentation/pull/1952/changes

image

This bugfix restores necessary part. There is no need for `$request` and `$persitenceManager` initialization inside `BaseCommand` class, but injecting `storagePid` into `$frameworkConfiguration `is still needed.

### DIFF
--- a/Classes/Command/BaseCommand.php
+++ b/Classes/Command/BaseCommand.php
@@ -29,6 +29,7 @@ use Symfony\Component\Console\Style\SymfonyStyle;
 use TYPO3\CMS\Core\Configuration\ExtensionConfiguration;
 use TYPO3\CMS\Core\Utility\GeneralUtility;
 use TYPO3\CMS\Core\Utility\MathUtility;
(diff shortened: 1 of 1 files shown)


sebastian-meyer (approved):

fschoelzel:
Thank you @beatrycze-volk. This solved my indexing problems. :partying_face:

beatrycze-volk:
> Thank you @beatrycze-volk. This solved my indexing problems. 🥳

Glad to read it, I have actually looked into this problem today again and found even better solution: https://github.com/kitodo/kitodo-presentation/pull/2099

PR #2117, opened 2026-08-14
[FEATURE] Add structure type for journal

beatrycze-volk:
See https://github.com/slub/dfg-viewer/issues/175


sebastian-meyer (approved):
Technically this is fine, but please see my comment in the related DFG-Viewer issue.
I am not convinced we really need another structure type for journals...

beatrycze-volk:
It doesn't add any additional logic on the application side, but allows to display correctly METS files with this type. I'm also not convinced that we need `periodical`s and `newspaper`s which both have as label 'Zeitschrift', but apparently there are people who see things differently. I think it would be beneficial to handle properly the documents' types which are containers for years/issues/volumes and so on.

PR #2134, opened 2026-08-28
[FEATURE] Mark text in image on hover in fulltext box

BFallert:
Hovering over a textline in the fulltext list now highlights the corresponding textline in the page image; the highlight is removed when the mouse leaves.

Until now, only the reverse direction was implemented: hovering over text in the page image highlights the corresponding textline in the fulltext list.


sebastian-meyer (approved):
Thank you!
Please update your branch to latest, so I can merge it into `main`.

(Hint: When opening a new pull request you can actually check an option to allow updating your branch by maintainers.)

PR #2148, opened 2026-09-17
[BUGFIX] OAI-PMH GetRecord always answered idDoesNotExist

stweil:
The OAI-PMH GetRecord verb failed for every valid identifier with the error "idDoesNotExist" because the record lookup query no longer selected the document's uid. The controller checks empty($document['uid']) to detect a missing record, so a found record without a uid column was treated as not existing.

Restore the uid to the SELECT list of getOaiRecord() and getOaiDocumentList(). Add functional tests for the GetRecord verb, including the success case and the unknown-identifier and unknown-metadataPrefix error cases, which previously had no coverage.

Assisted-by: OpenCode / qwen3.8-27b-thinking (Alibaba Cloud)
Fixes: 05c90acde207 ("[MAINTENANCE] Refactor OAI-PMH queries to...")

PR #2149, opened 2026-09-17
[BUGFIX] Remember failed document loads to stop repeated HTTP fetches

stweil:
AbstractDocument::getInstance() only cached documents that loaded successfully. A document URL that does not resolve to a readable METS or IIIF document (e.g. an OAI-PMH GetRecord URL for an unknown identifier) returned null without being remembered, so it was fetched over the network again on every document load.

In a single page view the document is loaded multiple times (e.g. by the getDocumentType TypoScript condition and by the plugin controller), so one request for such a document hammered the upstream OAI endpoint with 35 of identical fetches, once per evaluation.

DocumentCacheManager is now a SingletonInterface so the request-scoped instance (and a test mock) is shared. A unit test asserts that a cached failure short-circuits getInstance() without fetching.


stweil:
The first commit implements the fix using a new cache for failed OAI-PMH GetRecord requests.

I then thought that it would also be possible to use the already existing cache for failed requests, too. This simplifies the code a little bit and was implemented in the second commit.

So there is a choice between two options: a single extended cache which also caches failed requests or a new additional cache for failed requests. I can update this PR to use a single commit as soon as this decision is clear.

stweil:
This PR was motivated by high load issues on our digi.bib.uni-mannheim.de which were partially caused by the augmented traffic for bad identifiers, so it fixes a denial of service weakness in all current installations which provide OAI-PMH. See my related comment in PR #2148.

sebastian-meyer:
Personally, I prefer the second commit re-using the existing cache also for failed loading attempts. The small benefit of having both caches logically separated IMHO doesn't outweight the added complexity.

stweil:
> Personally, I prefer the second commit re-using the existing cache also for failed loading attempts.

So do I. Both commits are now squashed into a single commit, and I rebased the PR on top of the latest main branch.

sebastian-meyer (changes requested):
This is fine apart from the missing German localization (in `Resources/Private/Language/de.locallang_labels.xlf`). Could you please add that?

stweil:
The German translation was now added.

sebastian-meyer (approved):

PR #944, opened 2023-04-11
[MAINTENANCE] Move proxy URL configuration to AbstractController

beatrycze-volk:

PR #2150, opened 2026-09-18
[BUGFIX] Apply proxy URL in configureProxyUrl()

stweil:
configureProxyUrl() takes the URL by reference (`string &$url`) and is expected to rewrite it in place, but it built the proxy URL with ->build() and discarded the result, so `$url` was left unchanged.

The result was that enabling `useInternalProxy` had no effect: the viewer kept using the direct (cross-origin) image/fulltext URL instead of the same-origin PageViewProxy URL. This regressed in #944, when the previously inline `$uri = ...->build(); $image['url'] = $uri;` was moved into the new method.

Assign the built URL back to `$url` so the pass-by-reference contract holds again. The `setArguments()` calls still read the original URL for `url`/`uHash`, as the assignment only completes after the RHS has been fully evaluated.


stweil:
@sebastian-meyer, meanwhile I also have a local test case.

sebastian-meyer (approved):

PR #2152, opened 2026-09-18
[MAINTENANCE] Serve functional test METS documents from local fixtures

stweil:
The functional tests fetch METS documents from digital.slub-dresden.de, which is now protected by an Anubis anti-bot challenge. When the challenge page is returned instead of XML, parsing fails and a large number of tests break (e.g. "StartTag: invalid element name"), making the CI depend on external availability and behaviour.

Snapshot the used documents under Tests/Fixtures/Remote mirroring the origin paths and point the affected data sets and tests at them via the local test web container (http://web:8001), following the existing pattern of documents_local.csv. Add a README explaining the origin of the files and how to refresh them.

Assisted-by: OpenCode / Qwen3.8-27b-Thinking (Alibaba Cloud)


stweil:
@sebastian-meyer, all recent pull requests failed because SLUB recently added Anubis protection. This PR fixes it for the CI. But the DFG viewer is also affected. The demo pages for SLUB now produce a fatal error:
```
Gateway Timeout
The gateway did not receive a timely response from the upstream server or application.
```

Showing a more user friendly error message is another issue which must be fixed in Kitodo.Presentation ...

stweil:
It is now also possible to run the unit and functional tests without any online connection (as long as the Docker images and required packages already exist).

sebastian-meyer (approved):

PR #2153, opened 2026-09-18
[FEATURE] Show error message when a page image cannot be loaded

stweil:
When the image server is unavailable (e.g. returns HTTP 503) the page grid previously showed only broken image icons and the PageView full page image silently failed to load: the OpenLayers image source never settled, the map was never created and the viewer hung with no feedback.

The page grid thumbnails now fall back to the placeholder image and reveal a localized message via an  handler. The PageView fixes the unhandled rejection in fetchImageData()/initLayer() so the image load failure settles the deferred, and renders a localized message (including the image source URL) inside the map container.

Adds translatable strings pageGrid.thumbnailError and pageview.imageError (EN/DE).

Add functional tests for the image load error rendering.


stweil:
This PR was motivates by the current downtime of digital.slub-dresden.de. A thumbnail image for a failing server now looks like this:

Bildschirmfoto 2026-09-18 um 13 39 09

The failing page image shows the same error message plus the failing URL.

sebastian-meyer (approved):

PR #2154, opened 2026-09-19
[BUGFIX] Allow URL document ids in the page select form

stweil:
The page navigation form submits to NavigationController::pageSelectAction(), which reads the document reference from PageSelectForm::getId() to rebuild the redirect. That reference (tx_dlf[id]) is either a numeric uid or a URL. But PageSelectForm::$id was typed int, so when a document opened by URL was selected, ExtBase property mapping could not convert the URL into the field and left it unset; the redirect was then not built and the page never switched.

The int typing dates from 8facf4fa9 ("Adaptation of the page navigation form"), which created the form and its action with getId(): int / setId(int). (3c6342372 later tightened the property to a native int type, but the URL ids already failed before that.)

Type $id as string (property plus getId()/setId()), matching the sibling recordId property, which already carries the same uid-or-URL semantics.

Add a regression test that property-maps the submitted form fields into a PageSelectForm with a URL id; it fails against the int typing and passes now.


stweil:
Pull request #2156 fixes the unrelated CI failure.

sebastian-meyer (approved):

PR #2155, opened 2026-09-19
[TASK] Add unit tests for configureProxyUrl()

stweil:
configureProxyUrl() is expected to rewrite the URL by reference, but this contract was not covered by any test, which is how the regression from #944 (build() result discarded) went unnoticed.

Add unit tests asserting that
- the URL is rewritten in place to the result of UriBuilder->build(),
- the url/uHash arguments are computed from the original URL.

The first test fails against the pre-fix code.


stweil:
The test was missing in pull request #2150 which implemented the fix.

stweil:
Pull request #2156 fixes the unrelated CI failure.

Tests/Unit/Controller/AbstractControllerTest.php
```diff
@@ -0,0 +1,116 @@
...
+
+        $url = self::ORIGINAL_URL;
+        $controller->callConfigureProxyUrl($url);
+    }
+
+    /**
+     * Creates a PageViewController with a mocked UriBuilder that exposes the
+     * protected configureProxyUrl() method for testing.
+     */
+    private function createControllerWithUriBuilder(UriBuilder $uriBuilder): ConfigureProxyUrlCaller
```
sebastian-meyer:
```suggestion
    private function createControllerWithUriBuilder(UriBuilder $uriBuilder): ConfigureProxyUrlCallerInterface
```

sebastian-meyer (approved):

